PR #2498, opened 2022-10-05
Use derived store for `confirmedEventIndexesLoaded`

hpeebles:


julianjelfs (changes requested):
Overall potential problem is that `chatStateStore.getProp(chatId, "confirmedEventIndexesLoaded")` gets the state _specifically_ for `chatId` whereas `get(confirmedEventIndexesLoaded)` gets the data for the _selected_ chat. I am worried that that might reintroduce some subtle problems.  Not quite sure how to deal with that.

hpeebles:
> Overall potential problem is that `chatStateStore.getProp(chatId, "confirmedEventIndexesLoaded")` gets the state _specifically_ for `chatId` whereas `get(confirmedEventIndexesLoaded)` gets the data for the _selected_ chat. I am worried that that might reintroduce some subtle problems. Not quite sure how to deal with that.

Sorted. We now check that the chatId we want the range for is the currently selected chatId.

PR #2511, opened 2022-10-06
Only clear the `users` table in IndexedDb

hpeebles:
I've tested this locally and can see it successfully clearing the users table. So this is now ready for review.


julianjelfs (commented):
That seems a bit dodgy to me. Better to just go the version I reckon.

julianjelfs (commented):
I don't the point of this. If you bump the version the user store will get deleted along with everything else anyway?

hpeebles:
> I don't the point of this. If you bump the version the user store will get deleted along with everything else anyway?

But we don't want to clear everything. Only the user data is incorrect.

julianjelfs (approved):

PR #2518, opened 2022-10-08
a few changes to do with how users are loaded and cached

julianjelfs:
There are a few changes here:
 
1) staggering the loading of group details and previous events
2) splitting users into their own indexeddb since they rarely need to be ditched
3) vastly improve the efficiency of adding users to the user store (this makes by far the most difference)

Experimented with local storage which _is_ a good deal faster but a) it's sync and b) has limited capacity so isn't an ideal alternative.

### DIFF
--- a/frontend/src/components/home/CurrentChatMessages.svelte
+++ b/frontend/src/components/home/CurrentChatMessages.svelte
@@ -572,9 +572,12 @@
             initialised = false;
 
             if ($focusMessageIndex !== undefined) {
-                loadEventWindow(api, user, serverChat, chat, $focusMessageIndex)
-                    .then(onMessageWindowLoaded)
-                    .then(() => loadDetails(api, user, chat, events));
+                loadEventWindow(api, user, serverChat, chat, $focusMessageIndex).then(
+                    (messageIndex: number | undefined) => {
+                        loadDetails(api, user, chat, events);
+                        onMessageWindowLoaded(messageIndex);
+                    }
+                );
             } else {
                 loadPreviousMessages(api, user, serverChat, chat).then(() => {
                     loadDetails(api, user, chat, events);
--- a/frontend/src/components/home/CurrentChatMessages.svelte
+++ b/frontend/src/components/home/CurrentChatMessages.svelte
@@ -572,13 +572,15 @@
             initialised = false;
 
             if ($focusMessageIndex !== undefined) {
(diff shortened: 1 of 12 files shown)


frontend/src/components/home/CurrentChatMessages.svelte
```diff
@@ -572,13 +572,15 @@
...
-                    onMessageWindowLoaded
-                );
+                loadEventWindow(api, user, serverChat, chat, $focusMessageIndex)
+                    .then(onMessageWindowLoaded)
+                    .then(() => loadDetails(api, user, chat, events));
             } else {
-                loadPreviousMessages(api, user, serverChat, chat).then(onLoadedPreviousMessages);
+                loadPreviousMessages(api, user, serverChat, chat).then(() => {
+                    loadDetails(api, user, chat, events);
+                    onLoadedPreviousMessages();
```
hpeebles, line 581: Is there a reason the details are loaded after onMessageWindowLoaded but before onLoadedPreviousMessages?

julianjelfs: made them the same.

frontend/src/services/userIndex/userIndex.client.ts
```diff
@@ -50,9 +51,9 @@ export class UserIndexClient extends CandidService implements IUserIndexClient {
         );
     }
 
-    static create(identity: Identity, db?: Database): IUserIndexClient {
-        return db && process.env.CLIENT_CACHING && !cachingLocallyDisabled()
-            ? new CachingUserIndexClient(db, new UserIndexClient(identity))
+    static create(identity: Identity, db?: Database, userdb?: UserDatabase): IUserIndexClient {
```
hpeebles, line 54: You don't need to pass in db here

frontend/src/stores/user.ts
```diff
@@ -43,13 +43,11 @@ const { subscribe, update, set } = immutableStore<UserLookup>({
 });
 
 export function overwriteUser(lookup: UserLookup, user: PartialUserSummary): UserLookup {
-    return {
-        ...lookup,
-        [user.userId]: {
-            ...user,
-            username: user.username ?? lookup[user.userId]?.username,
-        },
+    lookup[user.userId] = {
```
hpeebles, line 46: So are we ok with this mutating the input?

hpeebles (approved):

PR #2530, opened 2022-10-11
remove our notifications service worker

julianjelfs:
This functionality is not rolled into the root level service worker


frontend/src/components/App.svelte
```diff
@@ -153,10 +155,11 @@
     }
 
     function startOnlinePoller() {
-        api?.markAsOnline();
         markOnlinePoller = new Poller(
             () => api?.markAsOnline() ?? Promise.resolve(),
-            MARK_ONLINE_INTERVAL
+            MARK_ONLINE_INTERVAL,
+            MARK_ONLINE_INTERVAL,
```
hpeebles, line 161: This will mark users as online when they are idle. I think we should leave this as it was where it would only mark a user as online while they have OC focused. Passing in `undefined` as the 2nd interval will do that.

hpeebles (approved):

PR #2575, opened 2022-10-21
Fix condition under which we fetch root key

hpeebles:


julianjelfs (commented):
Needs the same test in `data.client.ts`

julianjelfs (approved):

PR #2586, opened 2022-10-24
Apply local poll votes to reply contexts

hpeebles:


julianjelfs (commented):
Do _not_ like the idea of void functions mutating input at all. I get that the input has already been cloned but still seems a bit yucky.

julianjelfs:
I don't really like _anything_ about this pr! But maybe I don't understand it.

julianjelfs (approved):

PR #2603, opened 2022-10-31
Force new textbox instance to be created after sending each message

hpeebles:


frontend/app/src/components/home/MessageEntry.svelte
```diff
@@ -76,6 +76,9 @@
     $: isGroup = chat.kind === "group_chat";
     $: messageIsEmpty = (textContent?.trim() ?? "").length === 0 && fileToAttach === undefined;
 
+    // Update this to force a new textbox instance to be created
+    $: textboxId = Symbol();
```
julianjelfs, line 80: this doesn't need to be reactive since it is not computed. Seems like `let textBoxId = Symbol()` would be fine.

hpeebles: Done

PR #2614, opened 2022-11-02
Move local chat summary updates into their own store

hpeebles:


frontend/openchat-client/src/stores/chat.ts
```diff
@@ -116,6 +117,14 @@ export const chatSummariesStore: Readable<Record<string, ChatSummary>> = derived
     }
 );
 
+export function getChat(chatId: string): ChatSummary | undefined {
+    return get(chatSummariesStore)[chatId];
```
julianjelfs, line 121: We are already subscribing to these stores in the LiveState class so if we put this logic in the open chat client we can avoid the extra subscriptions. Not sure if it used anywhere else though.

hpeebles: Ah that `LiveState` class is cool.
Fixed.

PR #2623, opened 2022-11-03
Implement `freeze_group`

hpeebles:


backend/canisters/group/impl/src/updates/c2c_freeze_group.rs
```diff
@@ -0,0 +1,34 @@
...
+    run_regular_jobs();
+
+    mutate_state(|state| c2c_freeze_group_impl(args, state))
+}
+
+fn c2c_freeze_group_impl(args: Args, runtime_state: &mut RuntimeState) -> Response {
+    if runtime_state.data.frozen.is_none() {
+        let now = runtime_state.env.now();
+
+        runtime_state.data.frozen = Timestamped::new(
```
megrogan, line 20: Do we need to use Timestamped if the value already has the timestamp?

backend/canisters/user_index/api/src/queries/mod.rs
```diff
@@ -1,6 +1,6 @@
-pub mod c2c_is_super_admin;
 pub mod c2c_lookup_principal;
 pub mod c2c_lookup_user_id;
+pub mod c2c_lookup_user_id_v2;
```
megrogan, line 3: c2c_lookup_user ?

frontend/app/src/components/home/MessageEntry.svelte
```diff
@@ -455,6 +460,11 @@
         </div>
     {:else if preview}
         <div class="preview">
+            {#if isSuperAdmin}
+                <Button secondary={true} small={true} on:click={freezeGroup}>
```
megrogan, line 464: You might also want to join the group to see more details and then should also be able to freeze from the group menu

frontend/app/src/components/home/Home.svelte
```diff
@@ -690,6 +690,10 @@
         });
     }
 
+    function freezeGroup(ev: CustomEvent<string>) {
+        client.freezeGroup(ev.detail);
```
julianjelfs, line 694: We might be swallowing any error here. Typical pattern is for the _client_ to simply resolve to a `Promise<boolean>` based on overall success / failure. Component code will then toast an error message if it fails. Internally the client lib will log the actual api response and the exact failure mode. Should be able to find plenty of other similar examples I think.

frontend/openchat-client/src/openchat.ts
```diff
@@ -2597,6 +2602,10 @@ export class OpenChat extends EventTarget {
         }
     }
 
+    freezeGroup(chatId: string): Promise<FreezeGroupResponse> {
```
julianjelfs, line 2605: As mentioned I think this should probably just return a `Promise<boolean>` and deal with logging the specific error internally to the client library.

backend/canisters/user_index/api/src/queries/mod.rs
hpeebles: Done. We can consolidate the usages of the other 2 methods as a follow on commit.

megrogan:
Probably worth adding some integration tests now they're working again

frontend/app/src/components/home/MessageEntry.svelte
```diff
@@ -78,6 +78,8 @@
     $: userStore = client.userStore;
     $: isGroup = chat.kind === "group_chat";
     $: messageIsEmpty = (textContent?.trim() ?? "").length === 0 && fileToAttach === undefined;
+    $: isSuperAdmin = client.isSuperAdmin();
```
julianjelfs, line 81: Seems like this one will never react to anything and can just be a `let`

frontend/app/src/i18n/en.json
```diff
@@ -614,5 +614,10 @@
     "alsoCanisterId": "also your canister id",
     "version": "version",
     "websiteVersion": "of the website",
-    "advanced": "Advanced"
+    "advanced": "Advanced",
```
julianjelfs, line 617: need to generate the other languages

hpeebles:
> Doesn't this have to be split into backend and front end so the backend can go out first?

Done.

megrogan (approved):

PR #2628, opened 2022-11-06
Support direct messages to external bots

hpeebles:


julianjelfs (commented):
Maybe this is just a naming issue, but a `readonlyChat` property on a _user_ makes no sense to me.

hpeebles:
> Maybe this is just a naming issue, but a `readonlyChat` property on a _user_ makes no sense to me.

Yeah I didn't really like that either. Maybe something like `acceptsDirectMessages` is better?

hpeebles:
I felt `readonly` on its own wasn't very clear, and on the UI we say 'This chat is read-only' which is why I went with 'readonlyChat'.

julianjelfs:
> > Maybe this is just a naming issue, but a `readonlyChat` property on a _user_ makes no sense to me.
> 
> 
> 
> Yeah I didn't really like that either. Maybe something like `acceptsDirectMessages` or `canMessage` is better?

It's not _just_ the name either. The typing seems a bit loose. It seems like a User should be a union type and this property only exists in the BotUser branch.

julianjelfs (approved):

PR #2631, opened 2022-11-07
Fix reactivity

hpeebles:


julianjelfs (approved):
It's a bit horrible but I'm not sure I have any better ideas at the moment

PR #2644, opened 2022-11-09
Show user avatar at start of chat

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -2806,6 +2806,10 @@ export class OpenChat extends EventTarget {
         }
     }
 
+    get OPENCHAT_BOT_USER_ID(): string {
```
julianjelfs, line 2809: wouldn't it be nicer to have an `isBot(userId)` method on the client rather than expose the bot user id?

hpeebles: Done.

julianjelfs (approved):

PR #2646, opened 2022-11-09
Show avatar for empty chats

hpeebles:


frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -82,6 +82,8 @@
     $: chatStateStore = client.chatStateStore;
     $: userStore = client.userStore;
     $: isBot = chat.kind === "direct_chat" && $userStore[chat.them]?.kind === "bot";
+    // Show the avatar if the earliest loaded event is the earliest available event
```
julianjelfs, line 85: this seems to describe exactly the condition we already have i.e. that there are no more previous messages to load. Don't get why it needs to change.

frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -507,7 +509,9 @@
     $: {
         if (chat.chatId !== currentChatId) {
             currentChatId = chat.chatId;
-            initialised = false;
+
+            // If chat.latestEventIndex < 0, then no events need to be loaded so we set initialized to true
+            initialised = chat.latestEventIndex < 0;
```
julianjelfs, line 514: This is not what `initialised` is intended to mean. It's supposed to mean that we have loaded the first chunk of events for the chat and is used to control when we start listening for scroll events.

hpeebles: For an empty direct chat its fine to start listening to scroll events immediately because you can't scroll until there are some events.

frontend/app/src/components/home/CurrentChatMessages.svelte
hpeebles: The issue is that the current check was based purely on confirmed events, so the avatar would hide, then reappear when the messages confirm.

frontend/app/src/components/home/CurrentChatMessages.svelte
julianjelfs: Yes it _seems_ fine but I don't think it was added to short circuit user initiated scroll events. I still think it's seems confusing, risky and unnecessary to redefine "initialised" to mean something else.

julianjelfs (approved):

frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -496,18 +497,18 @@
         return firstKey;
     }
 
-    function chatUpdated(): void {
```
hpeebles, line 499: On chatUpdated we now check for earlier and later events, because for a new chat you can be within the loading distance of the top and the bottom.

PR #2665, opened 2022-11-16
Disallow usernames which look like the SNS bot

hpeebles:


backend/canisters/user_index/impl/src/updates/set_username.rs
```diff
@@ -76,8 +76,16 @@ pub fn validate_username(username: &str) -> UsernameValidationResult {
         return UsernameValidationResult::Invalid;
     }
 
-    if username.replace('_', "").to_uppercase() == "OPENCHATBOT" {
-        return UsernameValidationResult::Invalid;
+    let normalised = username.replace('_', "").to_uppercase();
```
megrogan, line 79: Nit - I would put this in a separate function something like isUsernameReserved

megrogan (dismissed):

backend/canisters/user_index/impl/src/updates/set_username.rs
hpeebles: Done

megrogan (approved):

PR #2669, opened 2022-11-16
Close notifications for msgs that have been read

megrogan:
This depends on the service worker change in this PR
https://github.com/open-ic/oc-landing/pull/19

### DIFF
--- a/frontend/openchat-client/src/events.ts
+++ b/frontend/openchat-client/src/events.ts
@@ -48,7 +48,7 @@ export class ChatUpdated extends Event {
     }
 }
 
-export class ChatsLoaded extends Event {
+export class ChatsUpdated extends Event {
     constructor() {
         super("openchat_event");
     }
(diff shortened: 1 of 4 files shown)


frontend/openchat-client/src/events.ts
```diff
@@ -48,6 +48,12 @@ export class ChatUpdated extends Event {
     }
 }
 
+export class ChatsLoaded extends Event {
```
julianjelfs, line 51: I slightly wonder about the name `ChatsLoaded` just because it might not quite convey what it means. Specifically it misses that this is something that is happening periodically and also that it may not have loaded any chats. 

Maybe `ChatsSynchronised`?

hpeebles (approved):

frontend/openchat-client/src/events.ts
megrogan: ChatsUpdated?

hpeebles (approved):

julianjelfs (approved):

PR #2702, opened 2022-11-22
1 message(s) -> 1 message

megrogan:


frontend/app/src/i18n/en.json
```diff
@@ -616,5 +616,6 @@
     "websiteVersion": "of the website",
     "advanced": "Advanced",
     "nMessagesDeleted": "{number} message(s) deleted",
```
hpeebles, line 618: You don't need the (s) in brackets anymore (same with the translations)

megrogan: More haste less speed!

hpeebles (approved):

PR #2707, opened 2022-11-23
FreezeChat FE changes

hpeebles:


frontend/app/src/components/home/Home.svelte
```diff
@@ -697,6 +697,14 @@
         });
     }
 
+    function freezeGroup(ev: CustomEvent<string>) {
```
julianjelfs, line 700: There's no need to bubble this stuff all the way up to Home now that we have global access to the client. We should just handle it directly in the MessageEntry component.

frontend/openchat-client/src/openchat.ts
```diff
@@ -2645,6 +2656,14 @@ export class OpenChat extends EventTarget {
         }
     }
 
+    freezeGroup(chatId: string, reason: string | undefined): Promise<FreezeGroupResponse> {
```
julianjelfs, line 2659: We have no error handling on this. I think we should follow the pattern used elsewhere where we parse the response inside the client and log the specific error if anything goes wrong. Then we actually just return `Promise<boolean>` from these functions indicating binary success or failure. Then the UI will toast an error message if it gets a `false` response.

frontend/openchat-client/src/openchat.ts
```diff
@@ -2678,6 +2690,54 @@ export class OpenChat extends EventTarget {
         }
     }
 
+    freezeGroup(chatId: string, reason: string | undefined): Promise<boolean> {
+        return this.api.freezeGroup(chatId, reason)
```
julianjelfs, line 2694: We should have catch handlers on these (and log and return false from them).

frontend/openchat-client/src/openchat.ts
```diff
@@ -2678,6 +2690,54 @@ export class OpenChat extends EventTarget {
...
+                }
+
+                this._logger.error("Unable to freeze group", resp);
+                return false;
+            });
+    }
+
+    unfreezeGroup(chatId: string): Promise<boolean> {
+        return this.api.unfreezeGroup(chatId)
+            .then((resp) => {
```
julianjelfs, line 2708: as above

julianjelfs (approved):

PR #2711, opened 2022-11-23
Undelete message on frontend

megrogan:


megrogan:
We can't deploy this until user and group canisters have been deployed so don't merge yet.

frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -119,6 +120,9 @@
     $: msgUrl = `/#/${chatId}/${msg.messageIndex}?open=true`;
     $: isProposal = msg.content.kind === "proposal_content";
     $: inert = deleted || collapsed;
+    $: undeletingMessagesStore = client.undeletingMessagesStore;
+    $: undeleting = $undeletingMessagesStore.has(msg.messageId) ?? false;
```
julianjelfs, line 124: Surely `$undeletingMessageStore.has(msg.messageId)` will _always_ return a boolean and therefore the `??` is redundant.

megrogan: If I don't do that then the compiler thinks undeleting is `boolean | undefined`

frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -602,7 +611,7 @@
                                         <div slot="text">{$_("forward")}</div>
                                     </MenuItem>
                                 {/if}
-                                {#if confirmed && groupChat && !me && !inThread && !isProposal}
+                                {#if confirmed && groupChat && !me && !inThread && !isProposal && !inert}
```
julianjelfs, line 614: this is all getting a bit horrible isn't it? How much of this information is available to the client library? Probably all of it which would allow us to write `client.canReplyToMessage(msg)` - but then maybe that causes reactivity problems.

frontend/app/src/components/home/DeletedContent.svelte
```diff
@@ -16,9 +17,14 @@
...
 <div class="deleted">
-    {$_("messageDeleted", {
-        values: { username, timestamp: timestampStr },
-    })}
+    <p>
+        {$_("messageDeleted", {
+            values: { username, timestamp: timestampStr },
+        })}    
+    </p>
+    {#if undeleting}
```
julianjelfs, line 25: isn't this going to cause an unpleasant jiggle (since it adds height). I had in mind that we would just _replace_ the original message with an "undeleting message ..." placeholder.

megrogan: Yes I did start doing that and that was also a bit weird.

frontend/openchat-client/src/openchat.ts
```diff
@@ -1056,16 +1057,44 @@ export class OpenChat extends EventTarget {
...
+        return this.api
+            .undeleteMessage(chat.kind, chatId, messageId, threadRootMessageIndex)
+            .then((resp) => { 
+                const success = resp.kind === "success";
+                if (success) {
+                    localMessageUpdates.markUndeleted(messageId.toString(), resp.message.content);
+                }
+                return success;
+            })
+            .catch((_) => false)
```
julianjelfs, line 1092: I think in other cases we would typically log the exact failure response here - otherwise we won't really know exactly what went wrong.

megrogan: Yes ok - I used deleteMessage as a template...

frontend/app/src/components/home/ChatMessage.svelte
julianjelfs: Hmmm - not when _I_ do it. `$: undeleting = $undeletingMessagesStore.has(msg.messageId);` seems to work fine.

frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -412,6 +415,19 @@
...
+        doUndeleteMessage(ev.detail);
+    }
+
+    function doUndeleteMessage(message: Message) {
+        if (message.content.kind !== "deleted_content" || 
+            message.content.deletedBy !== user.userId) {
+            return
+        };
+
+        client.undeleteMessage(chat.chatId, undefined, message.messageId);
```
julianjelfs, line 428: We aren't checking whether this worked or not. Generally we would toast a failure (though I can see that deleteMessage does not). Something like `blockUser` in the same file looks slightly better to me.

megrogan: I suppose the reasoning for delete message is that the user can see for themseves that the delete fails because the message reappears. Likewise they can see if the undelete fails because the message returns to the deleted state.

megrogan: I'll add toast for undelete failing and leave delete as is.

julianjelfs (approved):

PR #2713, opened 2022-11-24
Support searching chat messages by sender

megrogan:


backend/libraries/search/src/query.rs
```diff
@@ -17,6 +21,7 @@ impl Query {
                     value_lower: word.to_lowercase(),
                 })
                 .collect(),
+            users: HashSet::new(),
```
hpeebles, line 24: Maybe rename this to `new`, and pass in the users here.

hpeebles (dismissed):

hpeebles (approved):

PR #2720, opened 2022-11-28
Search by sender frontend

megrogan:
Don't merge until groups are released


frontend/app/src/components/home/CurrentChatSearchHeader.svelte
```diff
@@ -5,26 +5,40 @@
...
     import HoverIcon from "../HoverIcon.svelte";
     import { iconSize } from "../../stores/iconSize";
     import type { OpenChat } from "openchat-client";
-
-    const dispatch = createEventDispatcher();
+    import MentionPicker from "./MentionPicker.svelte";
 
     export let chat: ChatSummary;
     export let searchTerm = "";
+    export let members: Member[];
```
julianjelfs, line 20: we are already referencing the client in this component so there is no need to pass `members` and `blockedUsers` in is there?

julianjelfs (approved):

PR #2736, opened 2022-12-01
Add `initial_state_v2` and `updates_v2`

hpeebles:


backend/canisters/group_index/api/src/queries/filter_groups.rs
```diff
@@ -0,0 +1,22 @@
+use candid::CandidType;
+use serde::{Deserialize, Serialize};
+use types::{ChatId, DeletedGroupInfo, TimestampMillis};
+
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub struct Args {
+    pub chat_ids: Vec<ChatId>,
+    pub active_since: Option<TimestampMillis>,
```
megrogan, line 8: Maybe we should leave the name the same 
active_in_last

megrogan (approved):

megrogan (approved):

PR #2753, opened 2022-12-05
reported messages

julianjelfs:
This is not quite complete yet - need to wait for the backend to be available and then change the mapping. 

image


frontend/openchat-agent/src/services/userIndex/candid/types.d.ts
```diff
@@ -1,7 +1,7 @@
 import type { Principal } from '@dfinity/principal';
 import type { ActorMethod } from '@dfinity/agent';
 
-export type AccountIdentifier = Uint8Array | number[];
```
megrogan, line 4: You need to pull/build/use the latest didc

megrogan (approved):

PR #2762, opened 2022-12-06
Optionally suspend members when freezing a group

hpeebles:


backend/canisters/user_index/impl/src/updates/c2c_suspend_users.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    if let Some(ts) = suspended_until {
+        runtime_state
+            .data
+            .set_user_suspended_queue
+            .schedule(users_to_suspend.iter().map(|u| SetUserSuspended::Unsuspend(*u)).collect(), ts);
+    }
+
+    Success
+}
```
megrogan, line 45: Presumably we also want to queue up a UserEvent for each suspended user like we do in suspend_user:

runtime_state.data.user_event_sync_queue.push(
        user_id,
        UserEvent::UserSuspended(UserSuspended {
            timestamp: now,
            duration: duration.map_or(SuspensionDuration::Indefinitely, SuspensionDuration::Duration),
            reason,
        }),
    );

hpeebles: Here we're just queuing them all up to be handled later.
They then get processed during heartbeat and we simply call into `suspend_user`.
See here - https://github.com/open-ic/open-chat/blob/d60744087a8077115abc321677d8da7f5798befa/backend/canisters/user_index/impl/src/lifecycle/heartbeat.rs#L425

hpeebles: Doh! I've removed that duplication now.

megrogan (approved):

PR #2776, opened 2022-12-08
Add `notifications_index` canister for multi subnet notifications

hpeebles:


backend/canisters/local_group_index/c2c_client/src/lib.rs
```diff
@@ -2,6 +2,7 @@ use canister_client_macros::*;
 use local_group_index_canister::*;
 
 // Updates
+generate_c2c_call!(c2c_can_push_notifications);
```
megrogan, line 5: Nit - the comment is "Updates" but this is a query

backend/canisters/local_group_index/impl/src/lifecycle/init.rs
```diff
@@ -23,7 +23,7 @@ fn init(args: Args) {
         args.group_canister_wasm,
         args.user_index_canister_id,
         args.group_index_canister_id,
-        args.notifications_canister_ids,
+        vec![args.notifications_canister_id],
```
megrogan, line 26: Why not just have hold single notifications_canister_id?

backend/canisters/local_group_index/c2c_client/src/lib.rs
hpeebles: Fixed

megrogan (dismissed):

megrogan (approved):

PR #2789, opened 2022-12-12
Output stack trace for failing query call

megrogan:


julianjelfs (commented):
Not convinced that this is necessary when the underlying problem is fixed (which is that `message` and `stack` properties of `Error` are not serialised by default. I have already addressed that in my current branch which includes all the landing page stuff. 

We will probably end up seeing the stack _twice_ with this change as well.

megrogan:
> Not convinced that this is necessary when the underlying problem is fixed (which is that `message` and `stack` properties of `Error` are not serialised by default. I have already addressed that in my current branch which includes all the landing page stuff.
> 
> We will probably end up seeing the stack _twice_ with this change as well.

Yes I did that before I say your message on the subject

PR #2805, opened 2022-12-14
Move `updates` orchestration to the frontend

hpeebles:


frontend/openchat-agent/src/services/user/mappers.ts
```diff
@@ -867,40 +951,18 @@ function chatSummary(candid: ApiChatSummary): ChatSummary {
...
-        imageMessages: Number(candid.image_messages),
-        replies: Number(candid.replies),
-        videoMessages: Number(candid.video_messages),
-        polls: Number(candid.polls),
-        reactions: Number(candid.reactions),
-    };
-}
-
-function groupChatSummary(candid: ApiGroupChatSummary): GroupChatSummary {
+function groupChatSummary(candid: ApiGroupChatSummary, limitReadByMeUpTo = true): GroupChatSummary {
```
hpeebles, line 954: If `limitReadByMeUpTo` is true, then `readByMeUpTo` is capped to the latest message index.
This is usually what we want, but in the case where we load cached groups from the user canister, the cached latest message can be old and so can be a lower index than `readUpToByMe`.

frontend/openchat-agent/src/services/user/user.client.ts
```diff
@@ -123,6 +127,28 @@ export class UserClient extends CandidService implements IUserClient {
         );
     }
 
+    @profile("userClient")
+    getInitialStateV2(): Promise<InitialStateV2Response> {
+        const args = {
+            disable_cache: apiOptional(identity, false),
```
julianjelfs, line 133: this just always returns false

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -851,6 +869,129 @@ export class OpenChatAgent extends EventTarget {
...
+        const cached = await getCachedChatsV2(this.db, this.principal);
+        if (cached !== undefined) {
+            return await this.getUpdatesV2(cached);
+        }
+
+        let state: ChatStateFull;
+        const userResponse = await this.userClient.getInitialStateV2();
+        if (userResponse.cacheTimestamp === undefined) {
+            const groupPromises = userResponse.groupChatsAdded.map((g) => this.getGroupClient(g.chatId).summary());
+            const groupChats = (await Promise.all(groupPromises)).filter(isSuccessfulGroupSummaryResponse);
```
julianjelfs, line 882: There are potential problems here with all uses of `Promise.all`. If any of the promises reject, the whole `Promise.all` will reject. Because you are using async/await you will get an exception. We probably need to do _something_ about this. When `Promise.all` rejects (or throws in the async/await case) the error will itself have an `errors` propery which will be an array of errors for the rejected promises. We need to at least handle this otherwise it could blow up. Unless we decide that in that case it _should_ blow up.

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -851,6 +869,129 @@ export class OpenChatAgent extends EventTarget {
...
+                avatarId: userResponse.avatarId,
+                blockedUsers: userResponse.blockedUsers,
+                pinnedChats: userResponse.pinnedChats,
+                affectedEvents: {}
+            }
+        } else {
+            const groupPromises = userResponse.groupChatsAdded.map((g) => this.getGroupClient(g.chatId).summary());
+            const groupUpdatePromises = userResponse.cachedGroupChatSummaries.map((g) => this.getGroupClient(g.chatId).summaryUpdates(g.lastUpdated));
+
+            const groups = (await Promise.all(groupPromises)).filter(isSuccessfulGroupSummaryResponse);
```
julianjelfs, line 897: Same problems here.

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -851,6 +869,129 @@ export class OpenChatAgent extends EventTarget {
...
+        const groupIndexResponse = await this._groupIndexClient.filterGroups(groupChatIds, current.timestamp);
+
+        const activeGroups = new Set(groupIndexResponse.activeGroups);
+
+        const groupPromises = userResponse.groupChatsAdded.map((g) => this.getGroupClient(g.chatId).summary());
+        const groupUpdatePromises = current.groupChats
+            .filter((g) => activeGroups.has(g.chatId))
+            .map((g) => this.getGroupClient(g.chatId).summaryUpdates(g.lastUpdated));
+
+        const groups = (await Promise.all(groupPromises)).filter(isSuccessfulGroupSummaryResponse);
```
julianjelfs, line 931: same error handling question here.

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -851,6 +869,129 @@ export class OpenChatAgent extends EventTarget {
...
+        await setCachedChatsV2(this.db, this.principal, state);
+
+        return this.hydrateChatState(state);
+    }
+
+    async hydrateChatState(state: ChatStateFull): Promise<ChatStateFull> {
+        const directChatPromises = state.directChats.map((c) => this.hydrateChatSummary(c));
+        const groupChatPromises = state.groupChats.map((c) => this.hydrateChatSummary(c));
+
+        const directChats = await Promise.all(directChatPromises);
```
julianjelfs, line 962: and here

frontend/openchat-agent/src/utils/chat.ts
```diff
@@ -256,6 +256,172 @@ function mergeParticipants(_: Member | undefined, updated: Member) {
...
+
+    return Object.values(threadsRecord).map((t) => {
+        const readUpToUpdate = readUpToUpdates[t.threadRootMessageIndex];
+        return readUpToUpdate !== undefined
+            ? { ...t, readUpTo: readUpToUpdate }
+            : t;
+    });
+}
+
+export function isSuccessfulGroupSummaryResponse(response: GroupCanisterSummaryResponse): response is GroupCanisterGroupChatSummary {
```
julianjelfs, line 408: this return type is very cool! Did not know that was a thing.

frontend/openchat-client/src/openchat.ts
```diff
@@ -2990,6 +3003,89 @@ export class OpenChat extends EventTarget {
...
 
+    private async initialStateV2(): Promise<MergedUpdatesResponse> {
+        const response = await this.api.getInitialStateV2();
+
+        const chatSummaries = (response.directChats as ChatSummary[]).concat(response.groupChats);
+
+        this.updateReadUpToStore(chatSummaries);
+
+        return {
+            wasUpdated: true,
```
julianjelfs, line 3014: this hard-coding of `wasUpdated` to true looks a bit fishy. We don't want that to be true unless it really should be true.

frontend/openchat-client/src/openchat.ts
```diff
@@ -2990,6 +3003,89 @@ export class OpenChat extends EventTarget {
...
+        this.updateReadUpToStore(chatSummaries);
+
+        const avatarIdUpdate = response.avatarId === avatarId
+            ? undefined
+            : response.avatarId !== undefined
+                ? { value: response.avatarId }
+                : "set_to_none";
+
+        return {
+            wasUpdated: true,
```
julianjelfs, line 3061: as above

julianjelfs (commented):
Doesn't seem too horrendous overall - I just have a few worries about error handling.

frontend/openchat-agent/src/services/user/user.client.ts
hpeebles: Yeah previously it was driven off a local storage setting but that isn't accessible in the agent so I've just hardcoded `false` for now.

frontend/openchat-client/src/openchat.ts
hpeebles: Oh yeah good point I'll fix that.

frontend/openchat-agent/src/utils/promise.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function waitAll<T, E>(promises: Promise<T>[]): Promise<WaitAllResult<T, E>> {
+    const results = await Promise.allSettled(promises);
+
+    const success: T[] = [];
+    const errors: E[] = [];
+    for (const result of results) {
+        if (result.status === "fulfilled") {
+            success.push(result.value);
+        } else {
+            errors.push(result.reason as E);
```
julianjelfs, line 15: Seems like we are plucking `E` out of thin air here. Do we need to assume that we know the type of `E` rather than just leaving it as `unknown`?

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -851,6 +871,164 @@ export class OpenChatAgent extends EventTarget {
...
+        await setCachedChatsV2(this.db, this.principal, state);
+
+        return await this.hydrateChatState(state).then((s) => ({
+            state: s,
+            affectedEvents,
+            anyUpdates,
+            anyErrors,
+        }));
+    }
+
```
julianjelfs, line 994: seems like we might want to use `waitAll` here as well

hpeebles: I only used `waitAll` for groups because their updates are processed individually, so if a group fails, it can simply be tried again on the next iteration.
Whereas if hydrating a direct chat fails, we can't return all of the other direct chats and skip the one that failed, because there is only a single timestamp for all direct chats which means we could miss updates for the one that failed because the timestamp will have been incremented.
I think we can improve on this in the future though to make it more resilient.

julianjelfs (approved):
🤞

PR #2816, opened 2022-12-17
Improve how we keep users up to date

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -2581,7 +2581,14 @@ export class OpenChat extends EventTarget {
     }
 
     getUsers(users: UsersArgs, allowStale = false): Promise<UsersResponse> {
-        return this.api.getUsers(users, allowStale).then((resp) => {
+        const userGroups = users.userGroups.filter((g) => g.users.length > 0);
```
hpeebles, line 2584: Short circuiting prevents the logs getting filled up due to empty 'getUsers' requests, so we can detect how many genuine requests are being made.

megrogan (approved):

PR #2829, opened 2022-12-20
Update group controller

megrogan:


backend/libraries/utils/src/canister/update_settings.rs
```diff
@@ -0,0 +1,48 @@
+use crate::canister;
+use candid::{CandidType, Principal};
+use ic_cdk::api;
+use serde::{Deserialize, Serialize};
+use tracing::error;
+use types::CanisterId;
+
+pub async fn update_settings(canister_id: CanisterId, controllers: Vec<Principal>) -> Result<(), canister::Error> {
```
hpeebles, line 8: I think this should either take the full `CanisterSettings` or should be renamed to `set_controllers`.

backend/canisters/group_index/impl/src/model/canisters_requiring_controller_swap.rs
```diff
@@ -0,0 +1,49 @@
+use candid::CandidType;
+use serde::{Deserialize, Serialize};
+use std::collections::{HashSet, VecDeque};
+use tracing::info;
+use types::CanisterId;
+
+#[derive(CandidType, Serialize, Deserialize, Default)]
+pub struct CanistersRequiringControllerSwap {
```
hpeebles, line 8: It would be good to expose the count pending / failed in the metrics so we can see the progress.

backend/libraries/utils/src/canister/update_settings.rs
megrogan: Oh yes indeed I had meant to name it set_controllers

hpeebles (approved):

PR #2834, opened 2022-12-20
Use cycles_dispenser_client for local_group_index

megrogan:


backend/canisters/local_group_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -34,6 +40,8 @@ fn post_upgrade(args: Args) {
         utils::cycles::check_cycles_balance(user_index_canister_id)
     });
 
+    cycles_dispenser_client::init_from_bytes(&cycles_dispenser_client_state);
```
hpeebles, line 43: We need to kick off the timer job in `init` too. I think that's missing from all canisters though so we can add it to all of them as a follow up PR.

hpeebles (approved):

backend/canisters/local_group_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -34,6 +40,8 @@ fn post_upgrade(args: Args) {
         utils::cycles::check_cycles_balance(user_index_canister_id)
```
hpeebles, line 40: This needs to be changed... we no longer use `check_cycles_balance` here and also it was using the user_index... not sure how we both missed that!

hpeebles (approved):

PR #2849, opened 2022-12-22
Get last online dates from `online_users` canister

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -2996,6 +3015,27 @@ export class OpenChat extends EventTarget {
         }
     }
 
+    private async getLastOnlineDatesBatched(userIds: string[]): Promise<Record<string, number>> {
```
hpeebles, line 3018: I batched this because we load many avatars at a time and each one will request the last online date, so we want to batch them into a single request to the canister.

PR #2855, opened 2022-12-23
Use canister timers to retry sending failed messages

hpeebles:


backend/canisters/user/impl/src/updates/mod.rs
```diff
@@ -11,7 +11,6 @@ pub mod c2c_mark_read_v2;
 pub mod c2c_notify_group_deleted;
 pub mod c2c_notify_user_events;
 pub mod c2c_remove_from_group;
-pub mod c2c_retry_sending_failed_messages;
```
hpeebles, line 14: I opted to remove this so that if any do fail we'll see it in the user_index logs rather than it being invisible.

megrogan (dismissed):

megrogan (approved):

PR #2870, opened 2022-12-29
Backend to show whether pinned messages are read

megrogan:


hpeebles:
AFAICT this can't handle a few cases which aren't all that uncommon.
1 - if a user pins a message, then unpins it, everyone will see that there is a new pinned message when there isn't.
2 - if a user reads a pinned message, but then that message is unpinned and pinned again it'll appear as unread again when it isn't.

I think we should be handling these cases.

hpeebles (dismissed):

hpeebles (approved):

hpeebles (dismissed):

backend/canisters/user/impl/src/updates/mark_read_v2.rs
```diff
@@ -29,6 +30,11 @@ fn mark_read_impl(args: Args, runtime_state: &mut RuntimeState) -> Response {
                     .threads_read
                     .insert(thread.root_message_index, thread.read_up_to, now);
             }
+
+            group_chat.date_read_pinned = Timestamped::new(
```
hpeebles, line 34: This is always bumping the timestamp, we don't want that.

backend/canisters/user/impl/src/updates/mark_read_v2.rs
```diff
@@ -29,6 +30,13 @@ fn mark_read_impl(args: Args, runtime_state: &mut RuntimeState) -> Response {
                     .threads_read
                     .insert(thread.root_message_index, thread.read_up_to, now);
             }
+
+            if chat_messages_read.date_read_pinned.is_some() {
```
hpeebles, line 34: Maybe this?

`if chat_messages_read.date_read_pinned > group_chat.date_Read_pinned.value {
...
}`

megrogan: Yes I was just reviewing this

PR #2873, opened 2022-12-30
Frontend pinned read

megrogan:


megrogan:
Show an accent dot next to the pinned messages icon in if there are unread pinned messages. When the pinned messages panel is open the pinned messages are marked as read.

Screenshot 2022-12-30 at 11 25 29

frontend/app/src/components/home/Home.svelte
```diff
@@ -652,6 +652,10 @@
                     kind: "show_pinned",
                 },
             ]);
+            let chat = $selectedChatStore!;
```
julianjelfs, line 655: feels like this should probably be done inside the PinnedMessages.svelte component where the pinned messages are _actually_ loaded.

frontend/app/src/components/HoverIcon.svelte
```diff
@@ -27,4 +35,19 @@
             background-color: var(--icon-hv);
         }
     }
+
+    .unread {
```
julianjelfs, line 39: not too sure about this. Seems a bit of a stretch to make it part of the HoverIcon component rather than something that is just part of the slot injected in the pinned messages case. 

e.g.

```
            <HoverIcon title={$_("showPinned")}>
                <div class="pin">
                    <Pin
                        size={$iconSize}
                        color={pinnedSelected ? "var(--icon-selected)" : "var(--icon-txt)"} />
                </div>
            </HoverIcon>
```

```
    .pin {
        position: relative;
        display: grid;
        align-content: center;

        &::after {
            content: "";
            width: 10px;
            height: 10px;
            background-color: red;
            border-radius: 50%;
            position: absolute;
            bottom: -4px;
            right: -4px;
        }
    }
```

Also the combination of the icon-selected colour and the accent colour doesn't really work in light mode when the pinned panel is actually showing (arguably a problem with the icon-selected colour in light mode more than anything else).

megrogan: Yes probably is a bit of a stretch to make it part of HoverIcon.

Btw we don't show the unread dot when the pinned panel is open.

megrogan:
Just noticed you actually suggested some code for the dot - another commit coming...

julianjelfs (dismissed):

megrogan:
I won't merge yet as waiting for @hpeebles to review the backend PR which this is built on

julianjelfs (approved):

PR #2874, opened 2023-01-03
Update cn.json

SensyCheung:
change Chinese translation


julianjelfs (approved):

julianjelfs:
Replacing with [this duplicate](https://github.com/open-ic/open-chat/pull/2875) as there seems to be some problem with the build on this PR - the content is the same.

PR #2914, opened 2023-01-06
Fe view deleted

megrogan:


megrogan:
Not ready for review

backend/canisters/group/impl/src/queries/deleted_message.rs
```diff
@@ -0,0 +1,42 @@
...
+        if let Some((chat_events, min_visible_event_index)) = runtime_state
+            .data
+            .events
+            .get_with_min_visible_event_index(args.thread_root_message_index, min_visible_event_index)
+        {
+            if let Some(event_index) = chat_events.event_index_by_message_id(args.message_id) {
+                if event_index < min_visible_event_index {
+                    return NotAuthorized;
+                } else if let Some(message) = chat_events.message_internal_by_event_index(event_index) {
+                    return Success(SuccessResult {
```
hpeebles, line 31: We should probably check if the message is deleted (if we care, based on the response types it looks like we do)

backend/canisters/group/impl/src/queries/deleted_message.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+fn deleted_message_impl(args: Args, runtime_state: &RuntimeState) -> Response {
+    let caller = runtime_state.env.caller();
+    let participant = match runtime_state.data.participants.get(caller) {
+        None => return CallerNotInGroup,
+        Some(p) => p,
+    };
+
+    if !participant.role.can_delete_messages(&runtime_state.data.permissions) {
```
hpeebles, line 17: What about viewing your own message?
You may want to view it in case you deleted a few and can't remember which one was which.

frontend/CHANGELOG.md
```diff
@@ -9,3 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Added `max_messages` to `events` and `events_window` ([#2947](https://github.com/open-ic/open-chat/pull/2947))
+
+### Added
```
hpeebles, line 13: Nit: Added should come first

frontend/openchat-agent/src/services/user/mappers.ts
```diff
@@ -1014,3 +1016,28 @@ function formatIcrc1Account(candid: ApiIcrc1Account): string {
 
     return subaccount !== undefined ? `${owner}:${subaccount}` : owner;
 }
+
+export function deletedMessageResponse(candid: ApiDeletedDirectMessageResponse): DeletedDirectMessageResponse {
+    if ("Success" in candid) {
+        return { 
+            kind: "success",
+            content: messageContent(candid.Success.content, "unknown"),
```
hpeebles, line 1024: Sender is "unknown"? Isn't the sender on the message response? We need the userId not the username.

hpeebles:
LGTM other than the tiny comments I added

frontend/openchat-agent/src/services/user/mappers.ts
megrogan: It's only needed for the crypto content type which cannot be deleted. The response returns the MessageContent not the full Message.

frontend/CHANGELOG.md
megrogan: Oh I thought I saw it the other way round when you were looking at examples

frontend/openchat-agent/src/services/user/mappers.ts
megrogan: Changed

hpeebles (dismissed):

julianjelfs (approved):

hpeebles (approved):

PR #2918, opened 2023-01-06
Initialize each new local user index with all of the existing users

hpeebles:


backend/libraries/utils/src/canister_event_sync_queue.rs
```diff
@@ -55,20 +65,33 @@ impl<T> CanisterEventSyncQueue<T> {
...
-            } else {
-                self.queue.drain(..MAX_CANISTERS_PER_BATCH).collect()
-            };
-
-            let mut results = Vec::new();
-            for canister_id in canisters {
-                if let Some(events) = self.events.remove(&canister_id).filter(|events| !events.is_empty()) {
-                    results.push((canister_id, events));
+            let mut batch = Vec::new();
+            let mut canisters_to_readd = Vec::new();
```
megrogan, line 69: typo

PR #2922, opened 2023-01-09
Allow admins and senders to see deleted message content

megrogan:


backend/canisters/group/impl/src/queries/deleted_message.rs
```diff
@@ -0,0 +1,49 @@
...
+            .get_with_min_visible_event_index(args.thread_root_message_index, min_visible_event_index)
+        {
+            if let Some(event_index) = chat_events.event_index_by_message_id(args.message_id) {
+                if event_index < min_visible_event_index {
+                    return NotAuthorized;
+                } else if let Some(message) = chat_events.message_internal_by_event_index(event_index) {
+                    if message.deleted_by.is_none() {
+                        return MessageNotDeleted;
+                    } else if matches!(message.content, MessageContentInternal::Deleted(_)) {
+                        return MessageHardDeleted;
```
hpeebles, line 31: This is a new term we'd be introducing.
Within `undelete_messages` we say `ContentRemoved`, but this isn't exposed to the outside world.
But maybe we should say `MessageContentRemoved`?

backend/canisters/group/impl/src/queries/deleted_message.rs
```diff
@@ -0,0 +1,49 @@
...
+                    } else if participant.user_id != message.sender
+                        && !participant.role.can_delete_messages(&runtime_state.data.permissions)
```
hpeebles, line 33: This allows an admin to see a message which was deleted by the sender, do we definitely want that?

megrogan: As discussewd we'll use the "hard deleted" in general

hpeebles (approved):

PR #2924, opened 2023-01-09
make sure we don't get duplicate history entries for right hand panel

### LINKED ISSUE #2916: Make sure that the right hand panel history does not contain duplicate entries

julianjelfs:
fixes #2916


frontend/app/src/components/home/Home.svelte
```diff
@@ -544,7 +544,10 @@
     function addMembers() {
         if ($selectedChatId !== undefined) {
             rightPanelHistory.update((history) => {
-                return [...history, { kind: "add_members" }];
+                return [
+                    ...history.filter((p) => p.kind !== "add_members"),
```
hpeebles, line 548: Do we really want to clear all of the matching historical entries?
I thought the issue was that it was adding lots of the same entries in a row so we could just skip adding if the latest entry is of type 'add_members'?

hpeebles (dismissed):

hpeebles (approved):

PR #2926, opened 2023-01-09
fix group ownedBy translation bug

SensyCheung:


hpeebles (approved):

hpeebles:
Thanks for fixing this!

PR #2928, opened 2023-01-09
fix lastonline problem

julianjelfs:
I _think_ this should fix the problem.


hpeebles:
You only check for last online being 0 when getting the status, I think we'll need another check when getting the display date of when they were last online.

hpeebles (approved):

PR #2939, opened 2023-01-10
Avoid ICQC's when getting recommended groups

hpeebles:


frontend/openchat-agent/src/services/groupIndex/groupIndex.client.ts
```diff
@@ -48,6 +50,18 @@ export class GroupIndexClient extends CandidService implements IGroupIndexClient
         );
     }
 
+    recommendedGroups(exclusions: string[]): Promise<GroupChatSummary[]> {
+        const args = {
+            count: 20,
```
megrogan, line 55: I think we should increase this number

megrogan (approved):

PR #2941, opened 2023-01-10
Simplify how we handle each newly joined group

hpeebles:


megrogan (dismissed):

frontend/openchat-client/src/openchat.ts
```diff
@@ -1430,15 +1433,15 @@ export class OpenChat extends EventTarget {
 
         setSelectedChat(this.api, clientChat, serverChat, messageIndex);
 
-        const { selectedChat, focusMessageIndex, events } = this._liveState;
+        const { selectedChat, focusMessageIndex } = this._liveState;
```
hpeebles, line 1436: The events here are always empty since we've just reset the selected chat, so we're safe to remove them.

frontend/openchat-client/src/openchat.ts
```diff
@@ -371,7 +371,7 @@ export class OpenChat extends EventTarget {
         }
 
         this.refreshAffectedEvents(chat, affectedEvents);
-        this.updateDetails(chat, this._liveState.events);
```
hpeebles, line 374: We're safe to remove the events here since they will have already had their users processed when first loading them within `handleEventsResponse`.

megrogan (approved):

PR #2954, opened 2023-01-12
Remove one-time code only needed for upgrade

hpeebles:


megrogan (dismissed):

hpeebles:
> Does this need a CHANGELOG entry?

I think we can skip adding to the CHANGELOGs for these types of changes given that most upgrades require a bit of clean up

hpeebles:
I think if its more than just removing serde defaults then it probably should have an entry, like my previous commit which removes the one-time initialization of the first local user index.

megrogan (approved):

PR #2955, opened 2023-01-12
Avoid having to wait for inter subnet updates when joining groups

hpeebles:


backend/canisters/local_user_index/impl/src/updates/c2c_notify_user_index_events.rs
```diff
@@ -2,7 +2,7 @@ use crate::guards::caller_is_user_index_canister;
 use crate::{mutate_state, RuntimeState};
 use canister_api_macros::update_msgpack;
 use canister_tracing_macros::trace;
-use local_user_index_canister::c2c_notify_user_index_events::{Args, Response, UserIndexEvent};
+use local_user_index_canister::c2c_notify_user_index_events::{Args, Event, Response};
 use tracing::info;
 use types::{PhoneNumberConfirmed, ReferredUserRegistered, StorageUpgraded, UserEvent, UserSuspended, UsernameChanged};
```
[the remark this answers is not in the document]
hpeebles: I didn't do the name change. I thought this PR was big enough already and so would do it in a subsequent PR (if at all).

backend/libraries/types/src/user_events.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{MessageContent, Milliseconds, PhoneNumber, TimestampMillis, UserId};
```
megrogan, line 1: We should probably define these in the user canister api crate

backend/libraries/types/src/user_events.rs
```diff
@@ -10,6 +10,7 @@ pub enum UserEvent {
     ReferredUserRegistered(Box<ReferredUserRegistered>),
     UserSuspended(Box<UserSuspended>),
     OpenChatBotMessage(Box<MessageContent>),
+    UserJoinedGroup(Box<UserJoinedGroup>),
```
megrogan, line 13: We should probably define these in the user canister api crate

hpeebles: Yeah I actually tried that but the UserJoinedGroup is now part of the LocalUserIndex + UserIndex APIs so it seems nicer leaving it as is.

megrogan: Well I suppose you would define it twice

megrogan (approved):

PR #2956, opened 2023-01-12
reduce re-rendering

julianjelfs:
As discussed, hard to know whether this is really effective but it certainly prevents the constant firing of `afterUpdate` in the ChatMessage component.

### DIFF
--- a/frontend/app/src/components/home/ChatMessage.svelte
+++ b/frontend/app/src/components/home/ChatMessage.svelte
@@ -9,6 +9,7 @@
         Dimensions,
         MessageContent,
         OpenChat,
+        PartialUserSummary,
     } from "openchat-client";
     import EmojiPicker from "./EmojiPicker.svelte";
     import Avatar from "../Avatar.svelte";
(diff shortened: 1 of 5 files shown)


frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -109,11 +108,8 @@
     $: inert = deleted || collapsed;
     $: undeletingMessagesStore = client.undeletingMessagesStore;
     $: undeleting = $undeletingMessagesStore.has(msg.messageId);
-    $: canUndelete =
-        msg.content.kind === "deleted_content" &&
-        msg.content.deletedBy === user.userId &&
-        $now - Number(msg.content.timestamp) < 5 * 60 * 1000 && // Only allow undeleting for 5 minutes
-        !undeleting;
+
+    let canUndelete = true;
```
hpeebles, line 112: Since this will almost always be false, shouldn't it default to false?
Otherwise it'll render as true, then re-render when it changes, I think?

hpeebles: Oh wait doesn't this need to be a reactive variable?
So that when it changes those changes actually take effect.

julianjelfs: No it's fine as is since we are explicitly updating the value. It only needs to be reactive if it is computed from other values.

hpeebles (approved):

PR #2958, opened 2023-01-12
make the right panel dismissible (via esc or clicking away)

julianjelfs:
A previously requested feature.

### DIFF
--- a/frontend/app/src/components/home/Home.svelte
+++ b/frontend/app/src/components/home/Home.svelte
@@ -923,7 +923,7 @@
 </main>
 
 {#if showRight && floatRightPanel}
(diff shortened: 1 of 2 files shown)


frontend/app/src/components/home/Home.svelte
```diff
@@ -916,8 +923,8 @@
 </main>
 
 {#if showRight && floatRightPanel}
-    <Overlay fade={!$mobileWidth}>
-        <div class="right-wrapper" class:rtl={$rtlStore}>
+    <Overlay on:close={closeRightPanel} dismissible={true} fade={!$mobileWidth}>
```
megrogan, line 926: nit: can just do dismissible rather than dismissible={true}

megrogan (dismissed):

PR #2962, opened 2023-01-13
Prime caches by background loading events

hpeebles:


frontend/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+### Added
```
megrogan, line 10: DELETE ME

frontend/openchat-client/src/utils/cachePrimer.ts
```diff
@@ -0,0 +1,149 @@
...
+        if (chat.chatId !== this.selectedChatId && !chat.archived) {
+            this.enqueue([chat]);
+        }
+    }
+
+    async processNext(): Promise<void> {
+        try {
+            const chat = Object.values(this.pending).sort(compareChats)[0];
+            if (chat === undefined) {
+                debug("queue empty");
```
megrogan, line 55: Do we still need all the debug statements?

megrogan (approved):

frontend/openchat-client/src/utils/cachePrimer.ts
hpeebles: I guess I can take them out once I've seen it working for a while on webtest.

megrogan (approved):

frontend/openchat-client/src/utils/cachePrimer.ts
```diff
@@ -0,0 +1,149 @@
...
+            this.selectedChatId = chatId;
+        });
+        debug("initialized");
+    }
+
+    processChatUpdates(previous: ChatSummary[], next: ChatSummary[]): void {
+        const record = toRecord(previous, (c) => c.chatId);
+
+        const updated = next.filter((c) =>
+            c.chatId !== this.selectedChatId &&
```
julianjelfs, line 37: On reflection, does this selectedChatId check really achieve anything? It's only checking at the point where we enqueue the chats to prime. By the time we actually process the queue, the selected chat could be something different couldn't it. So we could either fail to prime a chat that we _should_ or incorrectly prime one that we don't need to.

hpeebles: The reason behind it was that if you're on a chat which then gets updated, the latest events will already get loaded, so if the cache primer also loads the latest events then you'll be double loading them.
However, if you're on a chat but not at the bottom, those new events won't get loaded, which means that if you click the 'go to latest' button it'll need to load the events from the canister.
Because of this I've removed everything to do with the selectedChatId which means clicking the 'go to latest' button will always come from the cache but it does mean we will sometimes be double requesting events.

hpeebles: I think sometimes double loading events is a small price to pay for it always being quick.
We can potentially avoid the double loading, I'll have a play around with that next week.

julianjelfs (approved):

PR #2965, opened 2023-01-13
Leave group optimistically

megrogan:


julianjelfs (approved):

frontend/openchat-client/src/openchat.ts
```diff
@@ -827,13 +827,15 @@ export class OpenChat extends EventTarget {
...
     leaveGroup(chatId: string): Promise<"success" | "failure" | "owner_cannot_leave"> {
+        localChatSummaryUpdates.markRemoved(chatId);
         return this.api
             .leaveGroup(chatId)
             .then((resp) => {
                 if (resp === "success" || resp === "not_in_group" || resp === "group_not_found") {
-                    localChatSummaryUpdates.markRemoved(chatId);
                     return "success";
                 } else {
+                    const chat = this._liveState.chatSummaries[chatId];
```
hpeebles, line 837: Surely this is always undefined because you marked it as removed on line 830.
You probably need to take the server chat summary or take a reference to the client one before removing it.

megrogan: Hmm. I did test the failing scenario and it re-instated the chat as planned. I'll take a closer look on Monday.

PR #2971, opened 2023-01-16
Add SNS1 token support to the website

megrogan:
We need to make some backend changes before this can be deployed


megrogan:
@julianjelfs can you reduce the file size of the SNS1 token image?

julianjelfs (dismissed):

frontend/app/src/components/home/profile/Accounts.svelte
```diff
@@ -130,7 +130,7 @@
                 background-image: url("../assets/sns1_token.png");
             }
             &.ckbtc {
-                background-image: url("../assets/bitcoin_token.png");
```
megrogan, line 129: This is wrong

julianjelfs: 🤦

julianjelfs (approved):

PR #2973, opened 2023-01-16
Allow user to see their own admin deleted messages

megrogan:


megrogan:
Probably want to first introduce a user setting for hard/soft deletion of messages. By default a user's messages would be soft-deleted (regardless of who deletes) but this option can be changed to hard-deleted (again regardless of who deletes).

hpeebles (approved):

PR #2975, opened 2023-01-17
Add SNS1 token to backend

megrogan:


backend/canisters/user/impl/src/model/direct_chats.rs
```diff
@@ -12,12 +12,6 @@ pub struct DirectChats {
 }
 
 impl DirectChats {
-    pub fn remove_old_deleted_message_content(&mut self, now: TimestampMillis) {
-        for chat in self.direct_chats.values_mut() {
-            chat.events.remove_old_deleted_message_content(now);
```
hpeebles, line 17: I think we should delete this from `chat_events`

hpeebles:
Needs a CHANGELOG entry

PR #2982, opened 2023-01-18
Support for assigning nicknames to other users

megrogan:


backend/canisters/user/api/src/updates/set_contact.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub nickname: OptionUpdate<String>,
+}
+
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub enum Response {
+    Success,
+    NoChange,
+    NicknameTooShort(FieldTooShortResult),
+    NicknameTooLong(FieldTooLongResult),
+    NicknameNameAlreadyUsed,
```
hpeebles, line 23: This can go

backend/canisters/user/impl/src/model/contacts.rs
```diff
@@ -0,0 +1,73 @@
...
+    NicknameTooShort(FieldTooShortResult),
+    NicknameTooLong(FieldTooLongResult),
+}
+
+impl Contacts {
+    pub fn set_contact(&mut self, contact: OptionalContact) -> SetContactResponse {
+        match contact.nickname {
+            OptionUpdate::NoChange => SetContactResponse::NoChange,
+            OptionUpdate::SetToNone => {
+                if self.map.remove(&contact.user_id).is_some() {
```
hpeebles, line 32: We need to be careful here that when we add more fields we only remove contacts which are empty.

backend/canisters/user/api/src/updates/set_contact.rs
megrogan: Unsaved change in IDE!

backend/canisters/user/impl/src/model/contacts.rs
```diff
@@ -0,0 +1,73 @@
...
+
+                if length_provided < MIN_NICKNAME_LEN {
+                    return SetContactResponse::NicknameTooShort(FieldTooShortResult {
+                        length_provided,
+                        min_length: MIN_NICKNAME_LEN,
+                    });
+                }
+
+                self.map
+                    .entry(contact.user_id)
```
hpeebles, line 57: You can do `entry.or_default().nickname = Some(nickname);`

hpeebles (approved):

PR #2984, opened 2023-01-18
Retry failed messages

julianjelfs:
When message sending fails this will now capture the failed message in indexeddb (and in a store). 

Failed messages will appear in the timeline and can be retried at any time.


frontend/app/src/components/home/Home.svelte
```diff
@@ -144,7 +144,8 @@
             closeThread();
         } else if (ev instanceof SendMessageFailed) {
             // This can occur either for chat messages or thread messages so we'll just handle it here
-            toastStore.showFailureToast("errorSendingMessage");
+            // TODO - not sure whether we need this now
+            // toastStore.showFailureToast("errorSendingMessage");
```
megrogan, line 148: I think not

frontend/app/src/i18n/en.json
```diff
@@ -225,6 +225,7 @@
     "reply": "Reply",
     "replyPrivately": "Reply privately",
     "deleteMessage": "Delete",
+    "retryMessage": "Retry message",
```
megrogan, line 228: translations

frontend/openchat-client/src/stores/unconfirmed.ts
```diff
@@ -17,20 +17,23 @@ function createUnconfirmedReadByThemStore() {
 function createUnconfirmedStore() {
     const store = writable<UnconfirmedMessages>({} as UnconfirmedMessages);
     let storeValue: UnconfirmedMessages = {};
-    store.subscribe((v) => storeValue = v);
+    store.subscribe((v) => (storeValue = v));
 
     function pruneOldMessages(): void {
```
megrogan, line 22: Should this be deleted?

frontend/openchat-agent/src/services/user/user.client.interface.ts
```diff
@@ -67,15 +68,15 @@ export interface IUserClient {
...
+        event: EventWrapper<Message>,
         replyingToChatId?: string,
         threadRootMessageIndex?: number
     ): Promise<[SendMessageResponse, Message]>;
     sendGroupICPTransfer(
         groupId: string,
         recipientId: string,
         sender: CreatedUser,
-        message: Message,
+        event: EventWrapper<Message>,
```
megrogan, line 79: Seems a little iffy given it is only needed for the caching layer

megrogan (approved):

frontend/openchat-client/src/stores/unconfirmed.ts
julianjelfs: Yikes - no!

frontend/openchat-agent/src/services/user/user.client.interface.ts
julianjelfs: Yes it is a bit iffy - and in fact I'm not convinced I actually _do_ need the whole event now. Might be able to do a bit of a re-jig there.

julianjelfs: Come to think of it- there's no easy way round it. We _do_ need to capture the whole event rather than just the message. Otherwise we won't be able to splice them into the timeline.

megrogan (approved):

PR #2994, opened 2023-01-20
Remove one time fix to user date created

megrogan:


hpeebles (approved):

megrogan:
Once this code is merged the user canisters will need to be upgraded ahead of the local_user_index. So don't merge this in until we're ready to upgrade user canisters.

hpeebles (approved):

PR #2997, opened 2023-01-23
Proposal vote redesign

julianjelfs:


hpeebles (approved):

frontend/app/src/components/home/proposals/ProposalContent.svelte
```diff
@@ -0,0 +1,387 @@
+<script lang="ts">
```
megrogan, line 1: Shame we don't have a diff to the previous ProposalContent

julianjelfs: It _is_ annoying. I'm not sure what dictates whether git views something as moved versus deleted. I was expecting it to be smart enough to tell that this file has merely moved.

PR #3013, opened 2023-01-25
Refactor `chat_events`

hpeebles:


backend/canisters/group/impl/src/lib.rs
```diff
@@ -81,8 +77,10 @@ impl RuntimeState {
 
     pub fn summary(&self, participant: &ParticipantInternal) -> GroupCanisterGroupChatSummary {
         let data = &self.data;
-        let latest_event = data.events.main().last();
+        let min_visible_event_index = participant.min_visible_event_index();
         let min_visible_message_index = participant.min_visible_message_index();
+        let main_events_reader = data.events.main_events_reader_filtered(min_visible_event_index);
```
megrogan, line 82: You decided to stick with _filtered then

hpeebles: Ah good point I'll update that

hpeebles: Updated

megrogan (approved):

PR #3039, opened 2023-01-31
Add sns_proposal.sh

megrogan:


sns_proposal.sh
```diff
@@ -0,0 +1,46 @@
...
+              --proposer $NEURON_ID \
+              --min-participants 100 \
+              --min-icp-e8s 50000000000000 \
+              --max-icp-e8s 100000000000000 \
+              --community_fund_investment_e8s 33333300000000 \
+              --min-participant-icp-e8s 100000000 \
+              --max-participant-icp-e8s 15000000000000 \
+              --swap-due-timestamp-seconds 1209600 \
+              --sns-token-e8s 2500000000000000 \
+              --target-swap-canister-id zcdfx-6iaaa-aaaaq-aaagq-cai \
```
hpeebles, line 42: What's this canisterId? A test SNS canister?

hpeebles: Maybe we should say <SNS_SWAP_CANISTER_ID> so that we know we need to update it once we have deployed the SNS.

megrogan: Probably SNS1. But doesn't matter for now

hpeebles (approved):

PR #3044, opened 2023-02-01
Prize message

megrogan:


backend/libraries/types/can.did
```diff
@@ -1206,4 +1225,28 @@ type MessageIndexRange =
...
+        prizes_remaining: nat32;
+        winners: vec UserId;
+        token: Cryptocurrency;
+        end_date: TimestampMillis;
+        caption: opt text;
+    };
+
+type PrizeWinnerContent =
+    record {
+        token: Cryptocurrency;
```
julianjelfs, line 1249: Since this is a message sent to the group, don't we also need to include the recipient / winner's UserId in the `PrizeWinnerContent`?

megrogan: Yes you're right

backend/canisters/user/impl/src/updates/send_message.rs
```diff
@@ -7,22 +7,24 @@ use chat_events::{PushMessageArgs, Reader};
...
 use types::{
-    CanisterId, CompletedCryptoTransaction, ContentValidationError, CryptoTransaction, MessageContent, MessageId, MessageIndex,
-    UserId,
+    CanisterId, CompletedCryptoTransaction, ContentValidationError, CryptoTransaction, MessageContentInitial, MessageId,
+    MessageIndex, UserId,
 };
 use user_canister::c2c_send_messages;
 use user_canister::c2c_send_messages::{C2CReplyContext, SendMessageArgs};
-use user_canister::send_message::{Response::*, *};
+use user_canister::send_message_v2::{Response::*, *};
```
hpeebles, line 15: Is this meant to be v2?

backend/libraries/chat_events/src/chat_events.rs
```diff
@@ -2,17 +2,19 @@ use crate::expiring_events::ExpiringEvents;
...
     ThreadSummary, TimestampMillis, Timestamped, UserId, VoteOperation,
 };
+use ic_ledger_types::Tokens;
 use itertools::Itertools;
 use search::{Document, Query};
 use serde::{Deserialize, Serialize};
 use std::cmp::Reverse;
 use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::HashMap;
+use types::Cryptocurrency;
```
hpeebles, line 17: This should be in the block on line 3

backend/canisters/user/impl/src/updates/send_message.rs
megrogan: yes

julianjelfs (approved):

PR #3045, opened 2023-02-01
Diamond membership and prize messages (front end)

julianjelfs:


frontend/app/src/components/home/PrizeContent.svelte
```diff
@@ -0,0 +1,208 @@
...
+
+    function claim() {
+        claimsStore.add(messageId);
+        client
+            .claimPrize(chatId, messageId)
+            .then((success) => {
+                if (success) {
+                    content = {
+                        ...content,
+                        winners: [...content.winners, client.user.userId],
```
julianjelfs, line 40: This is not right at the moment. I guess I should be using the local updates mechanism here.

megrogan (approved):

PR #3050, opened 2023-02-03
Remove 'authors' from Cargo.toml files

hpeebles:


megrogan:
Why is the cargo.lock for the integration tests deleted?

hpeebles:
> Why is the cargo.lock for the integration tests deleted?

It isn't used. It's a remnant from when the integration tests used to be in their own workspace.

megrogan (approved):

PR #3069, opened 2023-02-06
Add `pay_for_diamond_membership` to `user_index`

hpeebles:


backend/canisters/local_user_index/impl/src/updates/c2c_notify_events.rs
```diff
@@ -100,6 +101,14 @@ fn handle_event(event: Event, runtime_state: &mut RuntimeState) {
                 })),
             );
         }
+        Event::DiamondMembershipPaymentReceived(ev) => {
+            runtime_state.data.user_event_sync_queue.push(
+                ev.user_id.into(),
+                UserEvent::OpenChatBotMessage(Box::new(MessageContent::Text(TextContent {
+                    text: "Payment received for Diamond membership!".to_string(),
```
megrogan, line 108: Could be a bot more informative. It would be good if we could show a user their OpenChat transactions in general...

megrogan (approved):

megrogan (approved):

PR #3138, opened 2023-02-15
Add `c2c_events_internal` for recovering group events

hpeebles:


megrogan (approved):
We need the same for users of course

PR #3183, opened 2023-02-21
Add blog section to landing pages

julianjelfs:
I think this will reasonably easy to maintain. Adding a new post will just be a matter of creating a component with the markup (straight html) and then adding an entry into a dictionary to create a mapping between a slug and the component.

### DIFF
--- a/frontend/app/src/components/landingpages/blog/ResponsiveDesign.svelte
+++ b/frontend/app/src/components/landingpages/blog/ResponsiveDesign.svelte
@@ -1,0 +1,90 @@
+<h1>The Importance of Responsive Web Design for Your Website</h1>
+
+<h2>Introduction</h2>
+
+<p>
+    When it comes to building a successful website, there are many factors to consider, from the
+    design and layout to the functionality and user experience. One critical aspect that cannot be
+    overlooked is responsive web design. In today's mobile-first world, a website that is not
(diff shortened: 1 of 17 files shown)


frontend/app/src/components/landingpages/blog/ResponsiveDesign.svelte
```diff
@@ -0,0 +1,90 @@
+<h1>The Importance of Responsive Web Design for Your Website</h1>
```
megrogan, line 1: Did ChatGPT write this?

frontend/app/src/components/landingpages/blog/Security.svelte
```diff
@@ -0,0 +1,94 @@
+<h1>Cybersecurity Best Practices: How to Protect Your Business from Cyber Attacks</h1>
```
megrogan, line 1: And this?

megrogan (approved):

frontend/app/src/components/landingpages/blog/ResponsiveDesign.svelte
julianjelfs: Yes!

megrogan: Nice!

PR #3201, opened 2023-02-23
restrict scrolling mods to safari

julianjelfs:
when we add content to the bottom of the viewport Chrome, Firefox and Edge all maintain scroll position correctly and so we don't need to do anything. Safari does not. In that case we need to calculate the new scroll top based on the change in scroll height. This is not perfect, and so we _only_ want to do it if necessary.


frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -42,6 +42,7 @@
     import InitialGroupMessage from "./InitialGroupMessage.svelte";
     import { pathParams } from "../../stores/routing";
     import { push } from "svelte-spa-router";
+    import { isSafari } from "utils/devices";
```
hpeebles, line 45: Path looks wrong

hpeebles (approved):

PR #3208, opened 2023-02-24
tweak scroll and add debug logging that we can leave there

julianjelfs:


frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -287,15 +287,32 @@
 
     function shouldLoadPreviousMessages() {
         morePrevAvailable = client.morePreviousMessagesAvailable(chat.chatId);
-        return !loadingPrev && calculateFromTop() < MESSAGE_LOAD_THRESHOLD && morePrevAvailable;
+        const insideLoadThreshold = calculateFromTop() < MESSAGE_LOAD_THRESHOLD;
+        const result = !loadingPrev && insideLoadThreshold && morePrevAvailable;
+        if (result) {
+            console.debug(
+                "SCROLL: shouldLoadPreviousMessages [loadingPrev] [insideLoadThreshold] [morePrevAvailable]",
```
hpeebles, line 294: Given that we already check that `result` is true, the args we are printing are pointless since we know their values already.

julianjelfs: This is true! but I'm wondering if it isn't still nice to show them so that you don't need to look at the code

hpeebles: SGTM

hpeebles (approved):

PR #3225, opened 2023-02-26
Make it easy to replicate prod setup using `WASM_SRC=prod`

hpeebles:


canister_commit_ids.json
```diff
@@ -0,0 +1,15 @@
+{
```
hpeebles, line 1: This gets updated automatically each time we run `./upgrade-canister-prod.sh`

scripts/download-all-canister-wasms.sh
```diff
@@ -4,24 +4,24 @@ SCRIPT=$(readlink -f "$0")
...
 cd $SCRIPT_DIR
 
-COMMIT_ID=$1
+WASM_SRC=$1 # WASM_SRC is either empty, "latest", "prod" or the commit Id
 
-if [ -z "$COMMIT_ID" ]
+if [ -z "$WASM_SRC" ] || [ $WASM_SRC = "latest" ]
 then
-  COMMIT_ID=$(curl -s https://openchat-canister-wasms.s3.amazonaws.com/latest)
+  WASM_SRC=$(curl -s https://openchat-canister-wasms.s3.amazonaws.com/latest)
```
megrogan, line 11: Is this needed because the same check is done in download-canister-wasm.sh?

hpeebles: Its so that we only grab the latest commit Id once + it means we always download the wasms for the same commit Id.

megrogan (approved):

PR #3240, opened 2023-02-28
enable the chat token

julianjelfs:


frontend/openchat-agent/src/services/common/chatMappers.ts
```diff
@@ -377,8 +377,9 @@ export function apiToken(token: Cryptocurrency): ApiCryptocurrency {
             return { SNS1: null };
         case "ckbtc":
             return { CKBTC: null };
+        case "chat":
+            return { CHAT: null };
     }
-    throw new Error(`Unexpected Cryptocurrency type received - ${token}`);
```
hpeebles, line 381: We should probably have one of these, right?
`throw new UnsupportedValueError("Unexpected ApiCryptocurrency type received", candid);`

julianjelfs: In this case we don't need one because the switch statement covers all possible code paths.

PR #3246, opened 2023-03-01
Add `hot_group_exclusions`

hpeebles:


backend/canisters/user/impl/src/updates/add_hot_group_exclusions.rs
```diff
@@ -0,0 +1,29 @@
...
+#[trace]
+fn add_hot_group_exclusions(args: Args) -> Response {
+    run_regular_jobs();
+
+    mutate_state(|state| add_hot_group_exclusions_impl(args, state))
+}
+
+#[update(guard = "caller_is_owner")]
+#[trace]
+fn add_recommended_group_exclusions(args: Args) -> Response {
```
hpeebles, line 17: This will be removed once the FE switches to the new method

megrogan (approved):

PR #3258, opened 2023-03-02
Fix "groupWithN" translations

hpeebles:


julianjelfs (approved):
I thought the script must be broken but it's not so this must have been like this for _ages_

PR #3265, opened 2023-03-03
New routing

julianjelfs:
Switch over to https://github.com/visionmedia/page.js so that we can have normal routes (rather than hash fragment routing). 

This _should_ allow google to index our site properly 🤞 

Old links rendered in messages will get automatically translated. 

In coming request for hash based routes will get automatically redirected (client side).


frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -377,7 +377,8 @@
     }
 
     function doGoToMessageIndex(index: number): void {
-        push(`/${chat.chatId}`);
+        // TODO - replaces push
```
hpeebles, line 380: Is this comment needed? Isn't it already handled?

frontend/openchat-client/src/openchat.ts
```diff
@@ -326,6 +326,7 @@ export class OpenChat extends EventTarget {
     private _lastOnlineDatesPromise: Promise<Record<string, number>> | undefined;
     private _cachePrimer: CachePrimer | undefined = undefined;
     private _membershipCheck: number | undefined;
+    private _referralCode: string | undefined = undefined;
```
hpeebles, line 329: We've called this `referredBy` everywhere else (FE and BE) so I think it should be the same here.

hpeebles (approved):

megrogan (dismissed):

megrogan (approved):

PR #3288, opened 2023-03-09
Fix bug where direct chats would reappear after being removed

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -1674,6 +1674,13 @@ export class OpenChat extends EventTarget {
     }
 
     removeChat(chatId: string): void {
+        if (this._liveState.myServerChatSummaries[chatId] !== undefined) {
```
julianjelfs, line 1677: surely this breaks the mental model. Shouldn't the call to `localChatSummaryUpdates.markRemoved` deal with this? If it doesn't, does it not imply there is something wrong with the code that merges the server with the local data?

julianjelfs (approved):

PR #3293, opened 2023-03-10
Update files after ProposalsBot release

hpeebles:


megrogan:
You could also delete the one-off from the previous release in this PR. I think that makes sense as a rule.

julianjelfs (approved):

megrogan (approved):

PR #3295, opened 2023-03-10
Ensure job to sync events to local user indexes always runs

hpeebles:


backend/canisters/user_index/impl/src/updates/c2c_notify_events.rs
```diff
@@ -23,17 +23,15 @@ fn c2c_notify_events_impl(args: Args, runtime_state: &mut RuntimeState) -> Respo
...
-                    index,
-                    LocalUserIndexEvent::UserJoinedGroup(UserJoinedGroup {
-                        user_id: ev.user_id,
-                        chat_id: ev.chat_id,
-                        as_super_admin: ev.as_super_admin,
-                        latest_message_index: ev.latest_message_index,
-                    }),
-                );
-            }
+            runtime_state.push_event_to_local_user_index(
```
hpeebles, line 26: The issue was that here the job wasn't being started, but now it is because the method on `RuntimeState` starts the job if needed.

megrogan (approved):

PR #3319, opened 2023-03-16
Remove the super_admin role from groups

megrogan:
Still pending updates to the CHANGELOGs and upgrade_order.md


upgrade_order.md
```diff
@@ -1 +1 @@
-website > local_user_index > group
\ No newline at end of file
+user_index->local_user_index->group->website->user
```
hpeebles, line 1: This isn't right... local user index removes the field and calls into users so it must go after users.

hpeebles: And group c2c_join_group has a field removed so groups must go before local user index too.

hpeebles: I'm pretty sure it should be user_index->group->user->local_user_index->website

frontend/openchat-agent/src/services/group/candid/types.d.ts
```diff
@@ -799,10 +799,7 @@ export interface Participant {
 }
 export interface ParticipantAssumesSuperAdmin { 'user_id' : UserId }
 export interface ParticipantDismissedAsSuperAdmin { 'user_id' : UserId }
-export interface ParticipantJoined {
-  'user_id' : UserId,
-  'as_super_admin' : boolean,
-}
+export interface ParticipantJoined { 'user_id' : UserId }
```
hpeebles, line 802: This needs to go out before the backend

PR #3322, opened 2023-03-17
Remove `affected_events` from event responses

hpeebles:


megrogan:
Any change needed to upgrade_order.md?

megrogan (approved):

hpeebles:
> Any change needed to upgrade_order.md?

I was just thinking of not merging this in until the website goes out. Check the comment in the PR description.

Seems like a nicer way to prevent master being blocked for any canister.

Maybe we should only use upgrade_order.ms when there is a chain?

megrogan (approved):

PR #3338, opened 2023-03-21
Introduce the `exchange_client` canister + `market_maker`

hpeebles:


backend/canisters/exchange_client/api/can.did
```diff
@@ -0,0 +1,61 @@
...
+
+type CancelOrdersArgs =
+    record {
+        exchange_id: nat32;
+        orders: vec CancelOrderRequest;
+    };
+
+type CancelOrderRequest =
+    record {
+        id: text;
```
megrogan, line 48: What is this?

backend/canisters/exchange_client/impl/src/exchanges/icdex.rs
```diff
@@ -0,0 +1,164 @@
...
+            unit_size,
+        }
+    }
+
+    // TODO remove the `unwrap`s
+    pub async fn make_order(&self, caller: Principal, order: MakeOrderRequest) -> CallResult<()> {
+        let subaccount = subaccount(caller);
+        let from_account = AccountIdentifier::new(&self.this_canister_id, &Subaccount(subaccount));
+
+        let get_account_response: CallResult<(ic_icrc1::Account, String, Nat, Subaccount)> = make_c2c_call(
```
megrogan, line 42: @Hindsight said this could be generated - although he didn't say how. Worth chasing up again?

> TxAccount can be generated without calling the interface, it needs 3 parameters:
> 1. the principal of the caller
> 2. nonce
> 3. the canister principal of the trading pair
> Note that the nonce for each account is incremental and not random. If the nonce of a transaction is wrong, you need to get it again. getTxAccount() can query the latest nonce

hpeebles: Yeah we would need to store the last used nonce and put that into the calculation. I think this is a follow on improvement we can make. Making trades is really low volume (currently) so I think its fine to go out using the simpler (but less performant) method.

megrogan (approved):

PR #3346, opened 2023-03-22
Allow plat ops to set upgrade concurrency from FE

megrogan:


hpeebles (dismissed):

frontend/app/src/components/App.svelte
```diff
@@ -149,6 +150,36 @@
...
+        client
+            .setGroupUpgradeConcurrency(value)
+            .then((success) => {
+                if (success) {
+                    console.log("Group upgrade concurrency set", value);
+                } else {
+                    console.log("Failed to set group upgrade concurrency", value);
+                }
+            })
+            .catch((e) => {
```
julianjelfs, line 163: the general pattern is to do the try ... catch in the client so that component can assume that the function cannot fail (only resolve with success or failure)

julianjelfs: In fact - there is already a catch in the client so we can just remove it here.

hpeebles (approved):

PR #3369, opened 2023-03-24
Only retrieve active proposals

hpeebles:
This was already being done for NNS proposals but it was missing from SNS proposals.


megrogan (approved):

medical86:
Good morning. How is the project trying to reach out to it's many users satisfied and still maintain a steady and long lasting liquidity with the current economic challenges?

PR #3382, opened 2023-03-27
switch from affected_events to updated_events

julianjelfs:
this allows us to reinstate thread event caching


frontend/openchat-client/src/openchat.ts
```diff
@@ -1694,6 +1698,14 @@ export class OpenChat extends EventTarget {
         const userIds = this.userIdsFromEvents(resp.events);
         await this.updateUserStore(chatId, userIds);
 
+        if (this._liveState.selectedThreadRootMessageIndex !== undefined) {
```
hpeebles, line 1701: Do we need to check that the selected thread hasn't changed?

julianjelfs: yes probably

hpeebles (approved):

PR #3389, opened 2023-03-29
Expose `push_service_principals` in metrics

hpeebles:


hpeebles:
> We're calling it governance principal now right?

Governance principal can do admin things. `push_service_principal` can just read notifications.

julianjelfs (approved):

PR #3417, opened 2023-04-05
Add `set_message_reminder`

hpeebles:


backend/canisters/user/impl/src/updates/set_message_reminder.rs
```diff
@@ -0,0 +1,46 @@
...
+    let now = state.env.now();
+
+    state.data.timer_jobs.enqueue_job(
+        TimerJob::MessageReminder(MessageReminderJob {
+            chat_id: args.chat_id,
+            thread_root_message_index: args.thread_root_message_index,
+            message_index: args.message_index,
+            notes: args.notes,
+        }),
+        args.remind_at,
```
julianjelfs, line 41: what would happen if this were in the past?

hpeebles: Done. Now it'll return an error if the date is in the past

julianjelfs (approved):

PR #3426, opened 2023-04-11
Remove captcha

hpeebles:


backend/canisters/user_index/impl/src/updates/create_challenge.rs
```diff
@@ -1,19 +1,9 @@
...
-}
-
-fn create_challenge_impl(runtime_state: &mut RuntimeState) -> Response {
-    let now = runtime_state.env.now();
-
-    match runtime_state.data.challenges.create(now, &mut runtime_state.env.rng()) {
-        Some(challenge) => Response::Success(challenge),
-        None => Response::Throttled,
-    }
+    NotRequired
```
megrogan, line 8: Why not simply delete the method?

backend/canisters/user_index/api/can.did
```diff
@@ -314,8 +296,6 @@ type SetNeuronControllerForInitialAirdropResponse =
     };
 
 service: {
-    create_challenge : (EmptyArgs) -> (CreateChallengeResponse);
```
megrogan, line 317: This shouldn't be deleted yet right?

megrogan (dismissed):

PR #3429, opened 2023-04-11
Referral metrics to inform backdated rewards

megrogan:


backend/canisters/user_index/impl/src/queries/referral_metrics.rs
```diff
@@ -0,0 +1,81 @@
...
+    let mut user_referrals_map: HashMap<UserId, ReferralData> = HashMap::new();
+
+    for user in runtime_state.data.users.iter() {
+        if let Some(referred_by) = user.referred_by {
+            let data = user_referrals_map.entry(referred_by).or_default();
+            let icp_raised_for_paid_diamond: u64 =
+                user.diamond_membership_details.payments().iter().map(|p| p.amount_e8s).sum();
+            if icp_raised_for_paid_diamond > 0 {
+                data.paid_diamond += 1;
+                data.icp_raised_for_paid_diamond += (icp_raised_for_paid_diamond / 100_000_000) as u32;
```
hpeebles, line 29: This is going to do integer division, so all values < 1 will be 0. So 1 month and 3 month payments will be 0.

hpeebles: I think you should just leave everything as e8s, since the payments are as low as 0.2 ICP.

hpeebles (approved):

PR #3432, opened 2023-04-12
Invited users 💎 + add referral to group links

megrogan:


frontend/app/src/components/home/profile/ReferUsers.svelte
```diff
@@ -53,6 +53,15 @@
             dispatch("chatWith", viewedUserId);
         }
     }
+
+    function buildUsername(userStore: UserLookup, userId: string): string {
```
hpeebles, line 57: Don't we do this exact same thing in a few places already? It seems like we should be using a shared function for this.

frontend/app/src/components/SelectChatModal.svelte
```diff
@@ -17,6 +17,7 @@
     import SectionHeader from "./SectionHeader.svelte";
     import { _ } from "svelte-i18n";
     import { now } from "../stores/time";
+    import { usernameAndIcon } from "openchat-shared";
```
julianjelfs, line 20: we seem to have a few of these creeping in. The original idea was that the shared lib was only there to share things internally between various front end components but that the actual website code would only ever depend on the openchat-client project. It's easy enough to just expose whatever you need from openchat-shared via the openchat-client package.

frontend/app/src/components/home/profile/ReferUsers.svelte
```diff
@@ -92,7 +93,7 @@
                                 size={AvatarSize.Default} />
                         </div>
                         <LinkButton underline="hover">
-                            {$userStore[userId]?.username ?? $_("unknownUser")}
+                            {usernameAndIcon($userStore[userId])}
```
julianjelfs, line 96: this doesn't quite seem to do the same thing. I think if `$userStore[userId]` is undefined you're going to end up with "undefined" rather than the unknown user string.

PR #3435, opened 2023-04-12
Send diamond payment to sns treasury

megrogan:


backend/canisters/user/impl/src/updates/c2c_charge_user_account.rs
```diff
@@ -3,17 +3,19 @@ use crate::{read_state, run_regular_jobs};
...
 #[update_msgpack(guard = "caller_is_user_index")]
 #[trace]
 async fn c2c_charge_user_account(args: Args) -> Response {
     run_regular_jobs();
 
-    let (user_index_ledger_account, ledger_canister_id) = read_state(|state| {
+    let (sns_governance_ledger_account, ledger_canister_id) = read_state(|state| {
         (
-            state.data.user_index_ledger_account(),
+            default_ledger_account(SNS_GOVERNANCE_CANISTER_ID),
```
hpeebles, line 18: This doesn't actually need to be within the `read_state`, maybe its nicer with it outside?

hpeebles (approved):

PR #3440, opened 2023-04-12
Add new message content types for reminders

hpeebles:


backend/libraries/types/src/message_content.rs
```diff
@@ -557,6 +595,25 @@ pub struct PrizeWinnerContent {
     pub prize_message: MessageIndex,
 }
 
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug)]
+pub struct MessageReminderCreatedContent {
+    pub reminder_id: u64,
```
julianjelfs, line 600: Not sure why we need chat_id, thread_root_message_index & event_index in either of these new content types? Did you have something in particular in mind?

julianjelfs (approved):

PR #3452, opened 2023-04-14
Share diamond membership payment with referrer

megrogan:


backend/canisters/local_user_index/impl/src/updates/c2c_create_user.rs
```diff
@@ -105,8 +105,8 @@ fn commit(
 
     if let Some(referred_by) = referred_by {
         if runtime_state.data.local_users.get(&referred_by).is_some() {
-            runtime_state.data.user_event_sync_queue.push(
-                referred_by.into(),
+            runtime_state.push_event_to_user(
```
hpeebles, line 108: I think we can only push to local users otherwise we need to go via the user index. This is so that each canister can verify the call came from a local user index while only knowing it's own local one.

backend/canisters/user_index/impl/src/lib.rs
```diff
@@ -112,6 +113,52 @@ impl RuntimeState {
...
+                    } else {
+                        data.other += 1;
+                    }
+                }
+            }
+        }
+
+        for (user_id, data) in user_referrals_map {
+            if data.paid_diamond > 0 || data.unpaid_diamond > 0 {
+                let amount: u64 = data.icp_raised_for_paid_diamond_e8s + (data.unpaid_diamond as u64 * 50_000_000);
```
hpeebles, line 145: This should be 0.75ICP

backend/canisters/local_user_index/impl/src/updates/c2c_create_user.rs
megrogan: The only difference here is that it is calling start_job_if_required after pushing to the queue

hpeebles: Ahhh right yeah the user index sends a separate event to the referrer if they're on a different local user index

hpeebles (approved):

PR #3480, opened 2023-04-21
Support for BTC Miami referral codes

megrogan:


backend/canisters/user_index/impl/src/lib.rs
```diff
@@ -182,6 +190,8 @@ struct Data {
     pub internet_identity_canister_id: CanisterId,
     #[serde(default)]
     pub next_user_upgrade_started: bool,
+    #[serde(default)]
+    pub referrer_codes: HashMap<String, ReferralType>,
```
hpeebles, line 194: I think we should store more data about these codes such as when they were created and when the were used.

backend/canisters/user_index/impl/src/model/referral_codes.rs
```diff
@@ -0,0 +1,115 @@
...
+        if code.len() > 100 {
+            return None;
+        }
+
+        if let Some(details) = self.codes.get(code) {
+            match details.claimed {
+                Some(_) => None,
+                None => Some(details.referral_type.clone()),
+            }
+        } else if Principal::from_text(code).is_ok() {
```
hpeebles, line 95: `Principal::from_text(code).ok()`

backend/canisters/user_index/impl/src/model/referral_codes.rs
```diff
@@ -0,0 +1,115 @@
...
+            Some(ReferralType::User)
+        } else {
+            None
+        }
+    }
+
+    pub fn metrics(&self) -> HashMap<ReferralType, ReferralTypeMetrics> {
+        let mut metrics = HashMap::new();
+
+        for (_, details) in self.codes.iter() {
```
hpeebles, line 105: `self.codes.values()`

hpeebles (approved):

PR #3482, opened 2023-04-21
Expose user referral leaderboards

hpeebles:


backend/canisters/user_index/impl/src/updates/pay_for_diamond_membership.rs
```diff
@@ -83,6 +83,8 @@ fn process_charge(
 
     if let Some(diamond_membership) = runtime_state.data.users.diamond_membership_details_mut(&user_id) {
         let now = runtime_state.env.now();
+        let has_ever_been_diamond_member = diamond_membership.has_ever_been_diamond_member();
```
megrogan, line 86: We only pay referral rewards for Diamond payments within the first year of the user joining. Presumably we should do the same for the leaderboard.

julianjelfs (approved):

PR #3486, opened 2023-04-22
correct the infinite loop problem

julianjelfs:
The way scrollToMessageIndex currently works is that it looks for the message's dom element and if it doesn't find it it tries to load it and then recurses. 

This can lead to infinite recursion if the message is _hidden_ because even after the message is loaded, the dom element will still not exist. 

Therefore, when we don't find the DOM element, before we recurse, we check whether we have already loaded the message. If we have already loaded it then there is no point recursing. We still can't scroll to the message because it's not there, but recursion is futile at that point so we just scroll to the bottom (which may not be idea - but there isn't really anything better we can do).

### DIFF
--- a/frontend/app/src/components/home/ChatEventList.svelte
+++ b/frontend/app/src/components/home/ChatEventList.svelte
@@ -370,7 +370,10 @@
                 // but it's still not clear what to do in this circumstance because we cannot
                 // scroll to this message and there is nothing that we can do about it. So where *do* we scroll?
                 // The *next* message?, the bottom?, nowhere?
-                return scrollBottom();
+                const nextMessage = findElementWithMessageIndex(index + 1);
+                return nextMessage
+                    ? scrollToMessageIndex(chatId, index + 1, preserveFocus, filling)
+                    : scrollBottom();
             }
         }
     }
--- a/frontend/app/src/components/home/ChatEventList.svelte
+++ b/frontend/app/src/components/home/ChatEventList.svelte
@@ -358,9 +358,20 @@
                 return Promise.resolve();
             }
         } else if (!destroyed) {
-            // make sure that we short-circuit recursion when unmounted otherwise ∞ 💥
-            await client.loadEventWindow(chatId, index, threadRootEvent);
-            return scrollToMessageIndex(chatId, index, preserveFocus);
+            // check whether we have already loaded the event we are looking for
+            const loaded = findMessageEvent(index);
+            if (loaded === undefined) {
+                // we must only recurse if we have not already loaded the event, otherwise we will enter an infinite loop
+                await client.loadEventWindow(chatId, index, threadRootEvent);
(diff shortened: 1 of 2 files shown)


frontend/app/src/components/home/ChatEventList.svelte
```diff
@@ -358,9 +358,20 @@
...
+                // we must only recurse if we have not already loaded the event, otherwise we will enter an infinite loop
+                await client.loadEventWindow(chatId, index, threadRootEvent);
+                return scrollToMessageIndex(chatId, index, preserveFocus);
+            } else {
+                // if we got here it means that we could not find the DOM element for and event that we
+                // have already loaded. This isn't necessarily an error since a message might have been hidden
+                // but it's still not clear what to do in this circumstance because we cannot
+                // scroll to this message and there is nothing that we can do about it. So where *do* we scroll?
+                // The *next* message?, the bottom?, nowhere?
+                return scrollBottom();
```
hpeebles, line 373: Could we find the next visible element and scroll there?
Cos if you want to find index 15 but it's hidden, then being dropped on index 16 seems like the best option.

julianjelfs: Yeah we could just increment the index and recurse. Certainly simple. My worry is that we have _no_ reason to think that the next index exists. In practise it will ... until it doesn't and then we're in another tailspin. So I ended up thinking that something that may not be exactly right but definitely stops the recursion might be better.

hpeebles: Could we grab all the elements then find the one that is next and if none found only then scroll to bottom?

hpeebles: The most common case will be what you saw where a single message disappeared, so it would be good to make that behave nicely.

hpeebles: If that is simple I think we should do it but if it adds significant complexity then I'm happy to wave this through as is since it's a rare edge case.

julianjelfs: > Could we grab all the elements then find the one that is next and if none found only then scroll to bottom?

I think the main issue with trying to scroll to the next message is that we don't have any particularly simple way to tell whether the next message _should_ exist. We can make sure it's index is not > latestMessageIndex, but that still doesn't prevent us from hypothetically traversing through all the messages from our original target to the end. 

In practice we _know_ that won't happen because the only case where this occurs, the next message index _will_ exist. But it still feels ... dodgy. Perhaps if we only allowed it to recurse once in this circumstance? 

But the more I think about this, the more I think we are fixing the wrong thing. The root cause is that `getFirstUnreadMessageIndex` gives us the wrong answer (and there's no way for it to give us the _right_ answer).

hpeebles (approved):

PR #3488, opened 2023-04-24
allow platform mod to delete a message

julianjelfs:
This allows a more surgical response rather than freezing / deleting a group. 

Not this will also require a corresponding back end change I think.

### DIFF
--- a/frontend/app/src/components/App.svelte
+++ b/frontend/app/src/components/App.svelte
@@ -73,104 +73,113 @@
         calculateHeight();
         window.addEventListener("orientationchange", calculateHeight);
         window.addEventListener("unhandledrejection", unhandledError);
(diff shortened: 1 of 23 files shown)


frontend/app/src/components/App.svelte
```diff
@@ -73,104 +73,113 @@
...
+            }
+        });
+    }
+
+    function deleteMessage(
+        chatId: string,
+        messageId: bigint,
+        threadRootMessageIndex?: number | undefined
+    ): void {
+        client.deleteMessage(chatId, threadRootMessageIndex, messageId).then((success) => {
```
hpeebles, line 146: I think we'll need to pass up an `as_super_admin` flag, otherwise the BE won't know that it should check if the user is a super admin.

PR #3493, opened 2023-04-24
Add each new platform moderator to a moderation group

hpeebles:


backend/canisters/user_index/api/src/updates/assign_platform_moderators_group.rs
```diff
@@ -0,0 +1,29 @@
+use candid::{CandidType, Principal};
```
megrogan, line 1: I thought we only needed to call this method locally

hpeebles: Done

megrogan (approved):

megrogan (approved):

PR #3494, opened 2023-04-25
Set transaction created date on the frontend to prevent duplicates

hpeebles:


backend/libraries/types/can.did
```diff
@@ -1019,6 +1020,7 @@ type SnsPendingCryptoTransaction =
         to: Icrc1Account;
         fee: Tokens;
         memo: opt Memo;
+        created: TimestampNanos;
```
julianjelfs, line 1023: This says `TimestampNanos` but the component initialises it with `Date.now` which is milliseconds. Sounds wrong - unless it's not.

hpeebles: Yeah that's fixed now.

megrogan (approved):

PR #3507, opened 2023-04-26
Guidelines & Moderation

julianjelfs:
* adds the platform guidelines which is available as a landing page (`/guidelines`) or as a modal from the register modal
* fixes race condition on checking username validity on register modal
* adds a report message option to public messages
* message reports will be aggregated and sent to a special moderation group where they can be actioned by platform moderators.


frontend/app/src/components/landingpages/GuidelinesContent.svelte
```diff
@@ -0,0 +1,337 @@
...
+                <strong>
+                    Do not share sexually explicit or sexually suggestive content of other people
+                    without the subject’s knowledge and consent, otherwise known as “revenge porn.”
+                </strong>
+                This includes the non-consensual distribution of intimate media that was created either
+                with or without an individual’s consent.
+            </li>
+
+            <li>
+                <strong>
```
hpeebles, line 107: This should go

PR #3513, opened 2023-04-27
BE - replace invite codes with personal invites

megrogan:


backend/canisters/group/impl/src/lib.rs
```diff
@@ -306,39 +309,32 @@ impl Data {
...
         }
     }
 
-    pub fn min_visible_event_index(&self, caller: Principal, invite_code: Option<u64>) -> Option<EventIndex> {
+    pub fn min_visible_event_index(&self, caller: Principal) -> Option<EventIndex> {
         match self.participants.get_by_principal(&caller) {
             Some(p) => Some(p.min_visible_event_index()),
             None => {
-                if self.is_accessible_by_non_member(invite_code) && self.history_visible_to_new_joiners {
+                if self.is_public && self.history_visible_to_new_joiners {
```
hpeebles, line 325: If history is visible to new joiners then wouldn't we always return EventIndex::default()?

hpeebles: Ahh I guess if it is visible then the min_visible_event_index will be set to default.

backend/canisters/group/impl/src/updates/c2c_join_group.rs
```diff
@@ -93,7 +94,12 @@ fn c2c_join_group_impl(args: Args, runtime_state: &mut RuntimeState) -> Response
         mute_notifications: runtime_state.data.is_public,
```
hpeebles, line 94: The `min_visible_event/message_index` should come from the invitation

megrogan: Good catch

backend/canisters/group/impl/src/updates/decline_invitation.rs
```diff
@@ -0,0 +1,26 @@
...
+use user_index_canister_c2c_client::{lookup_user, LookupUserError};
+
+#[update]
+#[trace]
+async fn decline_invitation(_args: Args) -> Response {
+    run_regular_jobs();
+
+    let (caller, user_index_canister_id) = read_state(|state| (state.env.caller(), state.data.user_index_canister_id));
+
+    match lookup_user(caller, user_index_canister_id).await {
```
hpeebles, line 14: Do we really need to go to the user index? Can't we remove by the principal?

megrogan: Ah yes. I implemented this before I added the principal to the InvitedUsers

hpeebles (approved):

PR #3535, opened 2023-05-04
delete direct message - for me

julianjelfs:
I _think_ this is right - is this right?

Still can't quite shake the feeling that I already implemented this before.


frontend/openchat-client/src/utils/chat.ts
```diff
@@ -821,7 +821,7 @@ export function canDeleteOtherUsersMessages(chat: ChatSummary): boolean {
     if (chat.kind === "group_chat" && !chat.frozen) {
         return isPermitted(chat.myRole, chat.permissions.deleteMessages);
     } else {
-        return false;
+        return true;
```
hpeebles, line 824: If it is a frozen group chat you'll end up returning true

hpeebles (approved):

frontend/openchat-client/src/utils/chat.ts
hpeebles: I just updated it

PR #3536, opened 2023-05-04
handle replies to remote direct chats

julianjelfs:
I _think_ this solves the problem rehydrating reminders to messages in direct chats. We were treating the messages as if they belonged in a group chat because the source chat was not the _current_ chat.


frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -610,21 +607,15 @@ export class OpenChatAgent extends EventTarget {
             events,
             threadRootMessageIndex
         ).asyncMap((key, ctx, idxs) => {
-            if (ctx.chatId === currentChatId && chatType === "direct_chat") {
+            // Note that the latestClientEventIndex relates to the *currentChat*, not necessarily the chat for this messageContext
+            // So only include it if the context matches the current chat
+            // And yes - this is probably trying to tell us something
+            const latestIndex = ctx.chatId === currentChatId ? latestClientEventIndex : undefined;
+            if (chatType === "direct_chat") {
```
hpeebles, line 614: I don't think this is correct, where we resolve the events from is dependent on if the message being replied to is a direct or group message, rather than depending on the current chat type.

hpeebles: So we'll need to convert each chatId to a chat type which probably won't be easy since that data isn't readily available in the agent. Maybe we need to add the chat type to the reply context.

julianjelfs: Hmmm yes right. So it's working accidentally because the bot chat is a direct chat.

julianjelfs: So do we need to add chat type to the reply context to fix this? That would be a bit of pain in the arse.

julianjelfs: I can look up the chat from the cache - it's pretty much guaranteed to be there I guess.

hpeebles (approved):

PR #3553, opened 2023-05-09
Append suffix when registering if username is taken

hpeebles:


julianjelfs (commented):
Don't like this _at all_. Why not give me the opportunity to disambiguate with a username I actually _want_ i.e. the way it already works?

hpeebles:
> Don't like this _at all_. Why not give me the opportunity to disambiguate with a username I actually _want_ i.e. the way it already works?

The next PR will move registration to go via the local user index making the process much quicker.
The downside of this is that we no longer can check if the username is unique until the username propagates to the user index, but we lose all the speed improvements if we block the user during this time.

megrogan (dismissed):

megrogan (approved):

PR #3559, opened 2023-05-10
Miami stuff

julianjelfs:
Landing page
Discover hot groups from chat list
Add wallet icon to crypto content
Inline qr scan icon into the address input when sending crypto
Remove reminders, polls and all crypto from Diamond


frontend/app/src/components/home/upgrade/Features.svelte
```diff
@@ -57,6 +57,36 @@
         </div>
     </Feature>
 
+    <Feature>
```
megrogan, line 60: I would think don't show these features in the comparison table at all

frontend/app/src/components/home/profile/SendCrypto.svelte
```diff
@@ -4,40 +4,42 @@
...
     $: account = token === "icp" ? user.cryptoAccount : user.userId;
     $: cryptoBalance = client.cryptoBalance;
     $: transferFees = cryptoLookup[token].transferFeesE8s;
-    $: valid =
-        validAmount &&
-        amountToSendE8s > BigInt(0) &&
-        targetAccount !== "" &&
-        targetAccount !== account;
     $: symbol = cryptoLookup[token].symbol;
+    $: {
```
megrogan, line 30: Why this change out of interest?

frontend/app/src/components/home/upgrade/Features.svelte
julianjelfs: I think it's quite nice to have them just as a reference to what features exist and how lucky you all are to have access to them at no cost.

frontend/app/src/components/home/profile/SendCrypto.svelte
julianjelfs: because it makes it possible to two-way bind it to the parent component which is now necessary since send button is up a level now and should only be enabled when valid is true.

frontend/app/src/components/home/upgrade/Features.svelte
megrogan: By that reasoning we could add 10 more free features. I think they shouldn't be there.

PR #3590, opened 2023-05-15
convert faq to a landing page

### LINKED ISSUE #3588: Support FAQ via landing page url as well as modal
FAQs should be indexed and accessible whether we are logged in or not. So they should be split out into a landing page in the same way that the guidelines are.

julianjelfs:
fixes #3588


frontend/app/src/i18n/en.json
```diff
@@ -334,7 +334,7 @@
...
         "diamond_q": "What is 💎 Diamond Membership?",
-        "diamond_a": "You can upgrade to Diamond membership to have access to extra or enhanced features within OpenChat. <a href=\"?diamond=\">See here</a> for a feature comparison and to upgrade.",
+        "diamond_a": "You can upgrade to Diamond membership to have access to extra or enhanced features within OpenChat. <a href=\"/?diamond=\">See here</a> for a feature comparison and to upgrade.",
```
megrogan, line 337: This change navigates the user away from wherever they are

julianjelfs: Yes - it's a bit annoying. That link is not going to work from a landing page. It will only happen if you're clicking from the FAQ, but I guess it is annoying. Will have to think about how to make that work.

julianjelfs: Probably the only real solution is to extract the diamond features to work as a landing page too and then link to the landing page version from the faq. Can still keep the modal version for upgrade and for the `Tell me about Diamond` shortcut. Bit of a faff.

julianjelfs: Done

megrogan (approved):

PR #3592, opened 2023-05-15
Add moderator role to backend

megrogan:


backend/canisters/group/impl/src/lib.rs
```diff
@@ -184,7 +184,8 @@ impl RuntimeState {
             public: self.data.is_public,
             date_created: self.data.date_created,
             members: self.data.participants.len(),
-            admins: self.data.participants.admin_count(),
+            moderators: self.data.participants.admin_count(),
```
hpeebles, line 187: Wrong way round.

hpeebles (approved):

PR #3631, opened 2023-05-19
Implemented community::selected_updates

megrogan:


backend/canisters/community/api/src/queries/selected_updates.rs
```diff
@@ -0,0 +1,28 @@
+use candid::CandidType;
+use serde::{Deserialize, Serialize};
+use types::{CommunityMember, EventIndex, GroupRules, TimestampMillis, UserId};
+
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub struct Args {
+    pub updates_since: EventIndex,
```
hpeebles, line 7: This should take a timestamp.
The group one should take a timestamp too, I've been meaning to do that for a while but I haven't gotten round to it.

hpeebles: Because there are some things that may not bump the event index, and if we only pass in the event index then we have to return those updates on every loop whereas if we pass in the timestamp we can be more efficient with what we return.

hpeebles (approved):

PR #3642, opened 2023-05-23
Added community::c2c_delete_community

megrogan:


backend/canisters/group_index/api/src/updates/c2c_delete_community.rs
```diff
@@ -0,0 +1,17 @@
+use candid::CandidType;
+use serde::{Deserialize, Serialize};
+use types::UserId;
+
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub struct Args {
+    pub deleted_by: UserId,
+    pub community_name: String,
+    pub members: Vec<UserId>,
```
hpeebles, line 9: We may need to be careful here if a community has loads of members.
But not an issue for now.

hpeebles (approved):

PR #3672, opened 2023-05-26
Implement community::change_channel_role

megrogan:


backend/canisters/community/impl/src/updates/change_channel_role.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        if let Some(channel) = state.data.channels.get_mut(&args.channel_id) {
+            let now = state.env.now();
+
+            match channel
+                .chat
+                .change_role(member.user_id, args.user_id, args.new_role, false, false, now)
+            {
+                ChangeRoleResult::Success(_) => Success,
+                ChangeRoleResult::CallerNotInGroup => UserNotInChannel,
```
hpeebles, line 32: These should probably be renamed to match `User` and `TargetUser`

hpeebles (approved):

PR #3679, opened 2023-05-30
Implement various community query endpoints

megrogan:


backend/canisters/community/impl/src/queries/rules.rs
```diff
@@ -0,0 +1,21 @@
+use crate::read_state;
+use crate::RuntimeState;
+use canister_api_macros::query_candid_and_msgpack;
+use community_canister::rules::{Response::*, *};
+
+#[query_candid_and_msgpack]
+fn rules(args: Args) -> Response {
+    read_state(|state: &RuntimeState| rules_impl(args, state))
```
hpeebles, line 8: You don't need to specify the type

megrogan: Nor did clippy complain

PR #3689, opened 2023-05-31
Improve search by splitting "terms" into "tokens"

megrogan:


backend/libraries/search/src/document.rs
```diff
@@ -64,28 +93,39 @@ fn score_field(query: &Query, field: &Field) -> f32 {
...
-fn score_field_for_token(token: &Token, field: &str) -> f32 {
-    let field_lower = field.to_lowercase();
-    if field.starts_with(&token.value) {
+fn score_field_for_token(search_token: &Token, field: &Field) -> f32 {
+    let mut max_score: f32 = 0.0;
+
+    // Max of token matches
+    for field_token in &field.tokens {
+        let score = score_token_match(search_token, field_token) * calculate_length_boost(field_token.value.len() as f32);
+        max_score = max_by(max_score, score, |a, b| a.partial_cmp(b).unwrap());
```
hpeebles, line 115: Shouldn't the scores get summed up (or something like that), rather than only caring about the max?

hpeebles: Maybe we sum them up but then reduce scores for longer fields, so that "ABC ABCD" gets a lower score than "ABC" when the user searches "ABC"

PR #3699, opened 2023-06-01
Filter users to notify by mute at community level

megrogan:


backend/canisters/community/impl/src/updates/send_message.rs
```diff
@@ -54,6 +56,22 @@ fn send_message_impl(args: Args, state: &mut RuntimeState) -> Response {
...
+                    let users_to_notify: HashSet<UserId> = result
+                        .users_to_notify
+                        .into_iter()
+                        .filter(|u| {
+                            state
+                                .data
+                                .members
+                                .get((*u).into())
+                                .map_or(false, |m| !m.notifications_muted.value)
+                        })
```
hpeebles, line 69: You could `.chain(result.mentioned_users.into_iter())`

hpeebles (approved):

PR #3701, opened 2023-06-01
Add community members get_by_user_id

megrogan:


backend/canisters/community/impl/src/model/members.rs
```diff
@@ -103,7 +103,7 @@ impl CommunityMembers {
         let mut owner_count = self.owner_count;
         let mut admin_count = self.admin_count;
 
-        let member = match self.get_mut(target_user_id.into()) {
+        let member = match self.get_mut_by_user_id(&target_user_id) {
```
hpeebles, line 106: I think this should be named `get_by_user_id_mut, that feels more Rust-y.

hpeebles (approved):

PR #3704, opened 2023-06-01
Don't send notifications to suspended users

megrogan:


backend/canisters/user/impl/src/updates/c2c_toggle_reaction.rs
```diff
@@ -39,8 +39,10 @@ fn c2c_toggle_reaction_impl(args: Args, state: &mut RuntimeState) -> Response {
         if args.added {
             match chat.events.add_reaction(add_remove_reaction_args) {
                 AddRemoveReactionResult::Success => {
-                    if let Some((recipients, notification)) = build_notification(args, chat, now) {
-                        state.push_notification(recipients, notification);
+                    if !*state.data.suspended {
```
hpeebles, line 42: 1 more! :)

hpeebles (approved):

PR #3705, opened 2023-06-01
Add `expiry` to referral codes + end the BtcMiami promo

hpeebles:


backend/canisters/local_user_index/impl/src/model/referral_codes.rs
```diff
@@ -43,51 +45,75 @@ pub struct ReferralTypeMetrics {
...
-                    None => {
-                        details.claimed = Some(ReferralCodeClaim { when: now, who: user_id });
-                        true
-                    }
-                }
-            }
-            Entry::Vacant(_) => false,
-        }
-    }
+    // pub fn claim(&mut self, code: String, user_id: UserId, now: TimestampMillis) -> bool {
```
megrogan, line 84: Why is this commented out?

hpeebles: I'll just put an `allow(dead_code)` over it.

hpeebles: Actually I've deleted it instead. We can recover it if and when it is needed.

megrogan (approved):

megrogan (approved):

PR #3726, opened 2023-06-05
Miscellaneous fixes

megrogan:


backend/canisters/group/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Move invited_users from Data to GroupChatCore ([#3714](https://github.com/open-ic/open-chat/pull/3714))
+- Move invite_users logic into GroupChatCore ([#3716](https://github.com/open-ic/open-chat/pull/3716))
```
hpeebles, line 17: This is now in the changelog twice

megrogan: Actually two different commits doing slightly different things. The first just moved the invited_users field into GroupChatCore whereas the second moved the common aspects of the invite_users API into GroupChatCore. I could combine into a single message with two commits though if you prefer?

hpeebles (approved):

PR #3731, opened 2023-06-05
Implement community::invite_users_to_channel

megrogan:


backend/canisters/community/impl/src/updates/invite_users_to_channel.rs
```diff
@@ -0,0 +1,113 @@
...
+
+fn build_bot_message_event(
+    invited_by: UserId,
+    user_id: UserId,
+    community_id: CommunityId,
+    channel_id: ChannelId,
+    community_name: &str,
+    channel_name: &str,
+) -> Event {
+    let text = format!("You have been invited to the channel [{channel_name}](/community/{community_id}/channel/{channel_id}) in the community [{community_name}](/community/{community_id}) by @UserId({invited_by}).");
```
hpeebles, line 109: This text is the same for all users so should be pulled out of the inner loop

hpeebles (approved):

PR #3755, opened 2023-06-08
community::handle_activity_notification

megrogan:


backend/canisters/community/impl/src/activity_notifications.rs
```diff
@@ -0,0 +1,42 @@
...
+            timestamp: now,
+            member_count: state.data.members.len(),
+        });
+
+        let args = c2c_mark_community_active::Args {
+            duration: mark_active_duration,
+            public_community_activity,
+        };
+
+        ic_cdk::spawn(call_group_index_canister(state.data.group_index_canister_id, args));
```
hpeebles, line 26: Maybe use fire and forget handler?

backend/canisters/group/impl/src/activity_notifications.rs
```diff
@@ -16,7 +16,7 @@ pub(crate) fn handle_activity_notification(state: &mut RuntimeState) {
         .start_if_required(now, mark_active_duration);
 
     if requires_notification {
-        let public_group_activity = if state.data.chat.is_public { Some(extract_activity(now, &state.data)) } else { None };
+        let public_group_activity = state.data.chat.is_public.then_some(extract_activity(now, &state.data));
```
hpeebles, line 19: This will always run `extract_activity` even if it is never used.
You could use `.then(|| ...)` which would avoid that.

PR #3781, opened 2023-06-13
Favourites, archived, pinned + refactor updates

megrogan:


backend/canisters/user/api/can.did
```diff
@@ -588,6 +614,130 @@ type MigrateUserPrincipalResponse =
...
+    record {
+        summaries: vec DirectChatSummary;
+        pinned: vec ChatId;
+    };
+
+type GroupChatsInitial =
+    record {
+        summaries: vec UserCanisterGroupChatSummary;
+        pinned: vec ChatId;
+        cached: vec GroupChatSummary;
```
hpeebles, line 645: Maybe cached should be an optional record containing summaries + timestamp?

backend/canisters/user/api/src/lib.rs
```diff
@@ -142,3 +144,14 @@ pub struct DiamondMembershipPaymentReceived {
     pub recurring: bool,
     pub send_bot_message: bool,
 }
+
+pub fn map_chats_to_chat_ids(chats: Vec<Chat>) -> Vec<ChatId> {
```
hpeebles, line 148: Probably better to add a chat_id function to Chat which returns Option<ChatId>

hpeebles: Actually from the looks of it this is only a temporary thing needed for the switch from updates_v2

backend/canisters/user/api/can.did
```diff
@@ -565,6 +565,32 @@ type UnArchiveChatResponse =
...
+    record {
+        to_archive: vec Chat;
+        to_unarchive: vec Chat;
+    };
+
+type ArchiveUnarchiveChatsResponse =
+    variant {
+        Success;
+        PartialSuccess: record { chats_not_found: vec Chat; };
+        Failure: record { chats_not_found: vec Chat; };
```
hpeebles, line 590: Failure means everything failed, right?
So surely it doesn't need the list of chats?

backend/canisters/user/impl/src/lib.rs
```diff
@@ -204,18 +218,20 @@ impl Data {
...
-        if self.pinned_chats.value.contains(chat_id) {
-            self.pinned_chats.timestamp = now;
-            self.pinned_chats.value.retain(|pinned_chat_id| pinned_chat_id != chat_id);
-        }
+    // TODO: Legacy - delete me once communities enabled
+    pub fn unpin_chat(&mut self, chat_id: ChatId, now: TimestampMillis) {
+        let chat = if self.direct_chats.get(&chat_id).is_some() { Chat::Direct(chat_id) } else { Chat::Group(chat_id) };
+
+        self.favourite_chats.unpin(&chat, now);
+        self.favourite_chats.remove(chat, now);
```
hpeebles, line 234: `remove` should also `unpin`

backend/canisters/user/impl/src/model/communities.rs
```diff
@@ -28,7 +28,7 @@ impl Communities {
     }
 
     pub fn any_updated(&self, since: TimestampMillis) -> bool {
-        self.communities.values().any(|c| c.last_updated() > since)
+        self.communities.values().any(|c| c.last_updated() > since) || self.removed.iter().any(|g| g.timestamp > since)
```
hpeebles, line 31: The latest one will always be last so you can do this -
`self.removed.last().map(|g| g.timestamp > since)`

backend/canisters/user/impl/src/model/favourite_chats.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    pub fn chats(&self) -> &HashSet<Chat> {
+        &self.chats.value
+    }
+
+    pub fn pinned(&self) -> &Vec<Chat> {
+        &self.pinned.value
+    }
+
+    pub fn chats_since(&self, since: TimestampMillis) -> Option<Vec<Chat>> {
```
hpeebles, line 54: This either returns all of them or None, which the name doesn't really suggest.
Maybe `chats_if_updated` and `pinned_if_updated`?

backend/canisters/user/impl/src/model/direct_chats.rs
```diff
@@ -24,6 +26,14 @@ impl DirectChats {
         self.direct_chats.values().filter(move |c| c.has_updates_since(since))
     }
 
+    pub fn pinned(&self) -> &Vec<ChatId> {
+        &self.pinned.value
+    }
+
+    pub fn pinned_since(&self, since: TimestampMillis) -> Option<Vec<ChatId>> {
```
hpeebles, line 33: `pinned_if_updated`

backend/canisters/user/impl/src/model/group_chats.rs
```diff
@@ -32,6 +51,8 @@ impl GroupChats {
 
     pub fn any_updated(&self, since: TimestampMillis) -> bool {
         self.group_chats.values().any(|c| c.last_updated() > since)
+            || self.pinned.timestamp > since
+            || self.removed.iter().any(|g| g.timestamp > since)
```
hpeebles, line 55: Same here as with communities (`.last()...`)

backend/canisters/user/impl/src/updates/archive_unarchive_chats.rs
```diff
@@ -0,0 +1,94 @@
...
+                false
+            }
+        }
+    };
+
+    if success && archive {
+        state.data.favourite_chats.unpin(chat, now);
+    }
+
+    false
```
hpeebles, line 93: This should return `success` I think?

hpeebles (approved):

PR #3823, opened 2023-06-21
Fix ordering of pinned chats

megrogan:


backend/canisters/user/impl/src/model/favourite_chats.rs
```diff
@@ -1,30 +1,40 @@
...
+            false
+        }
     }
 
     pub fn pin(&mut self, chat: Chat, now: TimestampMillis) -> bool {
         if !self.pinned.value.contains(&chat) {
             self.pinned.timestamp = now;
             self.pinned.value.insert(0, chat);
-            self.chats.value.insert(chat);
+            self.chats.value.insert(0, chat);
```
hpeebles, line 37: You aren't checking for uniqueness here. You should just call `self.add`.

hpeebles (dismissed):

hpeebles (approved):

PR #3826, opened 2023-06-21
group_index::explore_groups + supporting changes

megrogan:


backend/canisters/group/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -1,13 +1,17 @@
+use std::time::Duration;
+
```
hpeebles, line 2: Unnecessary whitespace

backend/canisters/group/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -1,13 +1,16 @@
+use std::time::Duration;
```
hpeebles, line 1: This will fail fmt

backend/canisters/group_index/impl/src/jobs/mod.rs
```diff
@@ -1,7 +1,7 @@
 use crate::RuntimeState;
 
-mod calculate_community_hotness;
 mod calculate_hot_groups;
```
hpeebles, line 3: Can the `calculate_hot_groups` job be deleted now?

megrogan: Once the FE uses explore_groups we can of course delete "search" and "recommended_groups" and all supporting code including the calculate_hot_groups job.

hpeebles (approved):

PR #3840, opened 2023-06-23
Delete group after importing it into community

hpeebles:


backend/canisters/local_group_index/impl/src/updates/c2c_delete_group.rs
```diff
@@ -21,6 +21,7 @@ fn c2c_delete_group_impl(args: Args, state: &mut RuntimeState) -> Response {
     }
 }
 
+// TODO make this retry upon failure
 async fn delete_canister(canister_id: CanisterId) {
```
megrogan, line 25: Can't you just use the fire_and_forget_handler?

hpeebles: It's not quite that simple because it needs to call stop and then delete.
But we should be able to do something like that yeah.

megrogan (approved):

PR #3846, opened 2023-06-23
user::c2c_mark_community_updated_for_user

megrogan:


backend/canisters/user/impl/src/updates/c2c_mark_community_updated_for_user.rs
```diff
@@ -0,0 +1,22 @@
+use crate::{mutate_state, run_regular_jobs, RuntimeState};
+use canister_api_macros::update_msgpack;
+use canister_tracing_macros::trace;
+use types::CommunityId;
+use user_canister::c2c_mark_community_updated_for_user::{Response::*, *};
+
+#[update_msgpack]
```
hpeebles, line 7: Guard?

hpeebles (approved):

PR #3850, opened 2023-06-24
refactor svelte stores

julianjelfs:
Introduce a new layer into the svelte store hierarchy that contains all chats and all communities. Refactor the existing stores such that they are derived from this new state and if we have a community selected, the chat lists will contain the channels for that community. 

Also started to wire up channel level functions (involves making the existing group level functions generic where appropriate). 

Can now create and update communities; create, update and delete channels within communities. 

Still lots of rough edges, but it's coming together.


frontend/app/src/components/home/ChatEvent.svelte
```diff
@@ -61,6 +62,9 @@
 
     let userSummary: UserSummary | undefined = undefined;
 
+    $: level = $_(`level.${chatType === "channel" ? "channel" : "group"}`).toLowerCase() as
```
hpeebles, line 65: This seems weird... Its grabbing the translation of 'channel' or 'group' then casting it to type "channel" | "group" which will be wrong for all languages other than English.

hpeebles (approved):

PR #3851, opened 2023-06-26
Remove invalid replies to old messages in the Feature Requests group

hpeebles:


backend/libraries/chat_events/src/chat_event_internal.rs
```diff
@@ -597,6 +597,8 @@ pub struct ReplyContextInternal {
 pub struct ReplyContextInternalPrevious {
     #[serde(rename = "l")]
     pub event_list_if_other: Option<(ChatId, Option<MessageIndex>)>,
+    #[serde(rename = "c")]
+    pub chat_if_other: Option<(MultiUserChatInternal, Option<MessageIndex>)>,
```
hpeebles, line 601: I added this so that upgrading multiple times doesn't lose any data.
That also makes it safer in case we forget to remove this after the next upgrades.

megrogan (approved):

PR #3854, opened 2023-06-26
Add community::claim_prize and refactor group

megrogan:


backend/libraries/ledger_utils/src/lib.rs
```diff
@@ -4,9 +4,56 @@ pub mod sns;
...
+    PendingCryptoTransaction, TimestampNanos, TransactionHash, UserId,
+};
+
+pub fn create_pending_transaction(
+    token: Cryptocurrency,
+    amount: Tokens,
+    user_id: UserId,
+    now_nanos: TimestampNanos,
+) -> PendingCryptoTransaction {
+    let principal = Principal::from(user_id);
```
hpeebles, line 21: Lifting this out seems unnecessary since it is only used in one place and in one of the branches

hpeebles (approved):

PR #3857, opened 2023-06-27
Use correct notification type for claim_prize

megrogan:


backend/canisters/community/impl/src/updates/claim_prize.rs
```diff
@@ -125,10 +127,12 @@ fn commit(args: Args, winner: UserId, transaction: CompletedCryptoTransaction, s
             // Send a notification to group participants
             let notification_recipients = channel.chat.members.users_to_notify(None).into_iter().collect();
 
-            let notification = Notification::GroupMessageNotification(GroupMessageNotification {
-                chat_id: state.env.canister_id().into(),
+            let notification = Notification::CommunityMessageNotification(CommunityMessageNotification {
```
hpeebles, line 130: This should probably be called ChannelMessageNotification in case we ever have community messages in the future

hpeebles (approved):

PR #3871, opened 2023-06-28
Support message reminders for channels

megrogan:


backend/canisters/user/impl/src/timer_job_types.rs
```diff
@@ -44,6 +56,19 @@ pub struct MessageReminderJob {
     pub reminder_created_message_index: MessageIndex,
 }
 
+impl From<MessageReminderJobPrevious> for MessageReminderJob {
+    fn from(value: MessageReminderJobPrevious) -> Self {
+        MessageReminderJob {
+            reminder_id: value.reminder_id,
+            chat: Chat::Group(value.chat_id),
```
hpeebles, line 63: It could be a DirectChat... so maybe we keep the previous job types for a week or so until all the old ones are processed.

hpeebles: Oh actually we can detect if it is a direct chat because we hold all the direct chats in this canister

backend/canisters/user/impl/src/updates/set_message_reminder.rs
```diff
@@ -4,15 +4,35 @@ use crate::{mutate_state, openchat_bot, run_regular_jobs, RuntimeState};
...
 #[update(guard = "caller_is_owner")]
 #[trace]
-fn set_message_reminder(args: Args) -> Response {
+fn set_message_reminder(args: set_message_reminder::Args) -> Response {
+    run_regular_jobs();
+
+    mutate_state(|state| {
+        set_message_reminder_impl(
+            Args {
+                chat: Chat::Group(args.chat_id),
```
hpeebles, line 22: Need to detect if this is a direct or group chat.

backend/canisters/user/impl/src/timer_job_types.rs
```diff
@@ -58,9 +58,11 @@ pub struct MessageReminderJobPrevious {
 
 impl From<MessageReminderJobPrevious> for MessageReminderJob {
     fn from(value: MessageReminderJobPrevious) -> Self {
+        let is_direct_chat = read_state(|state| state.data.direct_chats.get(&value.chat_id).is_some());
```
megrogan, line 61: Hmm interesting

hpeebles (approved):

PR #3882, opened 2023-06-30
Consolidate `getUpdates` and `getInitialState`

hpeebles:


frontend/openchat-agent/src/services/user/mappers.ts
```diff
@@ -761,7 +761,7 @@ function mention(candid: ApiMention): Mention {
     };
 }
 
-function groupChatSummary(candid: ApiGroupChatSummary, limitReadByMeUpTo = true): GroupChatSummary {
```
hpeebles, line 764: `limitReadByMeUpTo` was only needed temporarily when we had a bug which had marked some `readUpTo` values way too high

julianjelfs (approved):

PR #3883, opened 2023-06-30
Top level routing stuff

julianjelfs:
and a few fixes


frontend/dev.sh
```diff
@@ -24,8 +24,8 @@ cd openchat-shared
 pm2 start --name "shared" "npm run dev" 
 cd ../openchat-agent
 pm2 start --name "agent" "npm run dev" 
-cd ../openchat-push
-pm2 start --name "push" "npm run dev" 
+# cd ../openchat-push
```
hpeebles, line 27: Do you want this included?

frontend/openchat-client/src/openchat.ts
```diff
@@ -4247,9 +4248,9 @@ export class OpenChat extends OpenChatAgentWorker {
         }
     }
 
-    clearSelectedCommunity(): void {
-        selectedCommunityId.set(undefined);
-    }
+    // clearSelectedCommunity(): void {
```
hpeebles, line 4251: Is this meant to be included?

frontend/dev.sh
julianjelfs: yes it only affects dev and it just puts a bit less strain on my machine

frontend/openchat-client/src/openchat.ts
julianjelfs: Yes - I don't _think_ we need it and it certainly won't cause any immediate problem.

hpeebles (approved):

PR #3888, opened 2023-07-03
More fixes

julianjelfs:
horrible horrible horrible

### DIFF
--- a/frontend/openchat-client/src/openchat.ts
+++ b/frontend/openchat-client/src/openchat.ts
@@ -3454,12 +3454,10 @@ export class OpenChat extends OpenChatAgentWorker {
     }
 
     async threadPreviews(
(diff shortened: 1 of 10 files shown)


frontend/openchat-client/src/openchat.ts
```diff
@@ -3454,12 +3454,10 @@ export class OpenChat extends OpenChatAgentWorker {
     }
 
     async threadPreviews(
-        chatId: ChatIdentifier | undefined,
+        _chatId: ChatIdentifier | undefined,
```
hpeebles, line 3457: Surely this can be removed?

julianjelfs: It's a bit of a hack. It's passed in so that the fn will get called reactively if the chatId changes. 🤮

PR #3901, opened 2023-07-04
Return is_default in channel summary and match

megrogan:


backend/canisters/community/api/can.did
```diff
@@ -1,993 +1,896 @@
 import "../../../libraries/types/can.did";
```
hpeebles, line 1: This entire file has changed...
I think you should put your formatter settings to match how this file used to be since that is how all our existing candid files are formatted.

hpeebles (approved):

PR #3923, opened 2023-07-06
Explore communities with language filter

megrogan:


backend/canisters/group/api/src/updates/convert_into_community.rs
```diff
@@ -6,6 +6,7 @@ use types::{AccessRules, CommunityId, CommunityPermissions};
 pub struct Args {
     pub rules: AccessRules,
     pub permissions: Option<CommunityPermissions>,
+    pub primary_language: Option<String>,
```
hpeebles, line 9: Should this be optional? I feel like it should be mandatory...

megrogan: Yeah I suppose so

hpeebles (dismissed):

hpeebles (approved):

PR #3977, opened 2023-07-18
Update favourites when a group is deleted or imported into a community

hpeebles:


julianjelfs (approved):
Does this also cover a group being _converted_ into a community?

hpeebles:
> Does this also cover a group being _converted_ into a community?

Yup

PR #4023, opened 2023-07-25
Revert "Add `symbol` to pending crypto transactions (#4018)"

hpeebles:


backend/bots/examples/group_prize_bot/impl/src/jobs/send_prizes.rs
```diff
@@ -146,18 +146,17 @@ async fn transfer_prize_funds_to_group(
     now_nanos: TimestampNanos,
 ) -> Result<CompletedCryptoTransaction, String> {
     // Assume ICRC-1 for now
-    let pending_transaction = types::icrc1::PendingCryptoTransaction {
+    let pending_transaction = types::sns::PendingCryptoTransaction {
         ledger: ledger_canister_id,
-        symbol: token.token_symbol().to_string(),
         token,
-        amount: amount as u128,
+        amount: Tokens::from_e8s(amount),
```
megrogan, line 152: These changes shouldn't be reverted though right?

hpeebles: I left them in at first but then I just thought I'd do a proper revert and switch away from SNS to ICRC1 separately.

julianjelfs (approved):

PR #4040, opened 2023-07-27
Prevent private default + some FE response wiring

megrogan:


hpeebles (approved):

hpeebles:
The new event type needs to be handled here - https://github.com/open-chat-labs/open-chat/blob/master/frontend/app/src/components/home/ChatEvent.svelte#L275

hpeebles (approved):

PR #4047, opened 2023-07-27
Hot groups -> Explore groups

megrogan:
Still referred to as 🔥 Whats hot in main menu in legacy mode


julianjelfs (commented):
Why "discover" when communities is "explore"?

julianjelfs (approved):

PR #4060, opened 2023-07-28
Link to OC community in OC Bot welcome messages

megrogan:
Need to use the prod OC community Id when we have it...


frontend/app/src/utils/features.ts
```diff
@@ -2,4 +2,4 @@
 
 import { createLsBoolStore } from "openchat-client";
 
-export const communitiesEnabled = createLsBoolStore("openchat_communities_enabled", true);
+export const communitiesEnabled = createLsBoolStore("openchat_communities_enabled", false);
```
hpeebles, line 5: !

megrogan: > We should get the website out with communities turned off and then create an OpenChat community ahead of time so we can feed the id into the OC Bot welcome messages in the local_user_index...

hpeebles (approved):

PR #4067, opened 2023-07-29
prevent ios problems with clicks inside hoverables

julianjelfs:
in iOS in safari if a parent element has a :hover pseudo selector and a descendent is clicked, the click handler does not fire first time. The first tap will trigger the parent hover state and only a second tap will activate the button. 

To prevent this weird behaviour I'm converting touchstart to click for certain elements where we know this is a problem. 

This is not a great solution. A better solution would be to replace all hover interactions with explicit taps or long presses on touch devices. But that will need a bit more testing so this seems a safer short term hacklet.

### DIFF
--- a/frontend/app/src/components/home/Tweet.svelte
+++ b/frontend/app/src/components/home/Tweet.svelte
@@ -56,7 +56,7 @@
                 4em,
                 2em,
                 var(--button-spinner),
-                "/assets/twitter.svg",
+                "/assets/twitter.svg?x",
                 0.8s,
                 1
             );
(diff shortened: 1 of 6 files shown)


hpeebles (approved):

frontend/app/src/components/home/Tweet.svelte
```diff
@@ -56,7 +56,7 @@
                 4em,
                 2em,
                 var(--button-spinner),
-                "/assets/twitter.svg",
+                "/assets/twitter.svg?x",
```
hpeebles, line 59: Is that meant to be here?

julianjelfs: yes it's just a cache busting hack

hpeebles (approved):

PR #4070, opened 2023-07-31
Fix notification path

megrogan:


frontend/openchat-push/src/index.ts
```diff
@@ -96,7 +96,7 @@ async function handleNotificationClick(event: NotificationEvent): Promise<void>
             path: event.notification.data.path,
         });
     } else {
-        const urlToOpen = `${new URL(self.location.origin)}${event.notification.data.path}`;
+        const urlToOpen = new URL(event.notification.data.path, self.location.origin);
```
hpeebles, line 99: This seems the wrong way round...?

megrogan: Well it might *seem* the wrong way round but it is in fact the right way round

megrogan: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

julianjelfs (approved):

PR #4102, opened 2023-08-01
Reduce randomness in explore_communities

megrogan:


backend/canisters/group_index/impl/src/jobs/calculate_hotness.rs
```diff
@@ -99,5 +99,5 @@ pub fn calculate_hotness(
     let random = random as f64 / u32::MAX as f64;
 
     // Weighted sum of new, popular, hot and random
-    ((0.5 * newness + popularity + hotness + random) * 1_000_000.0) as u32
+    ((0.5 * newness + popularity + hotness + 0.5 * random) * 1_000_000.0) as u32
```
hpeebles, line 102: Is that really enough of a difference?

The groups with 1 member are all currently above the OpenChat community with 11k members.

hpeebles (approved):

PR #4124, opened 2023-08-03
Automate adding SNS tokens to the registry

hpeebles:


backend/canisters/registry/api/src/updates/add_token.rs
```diff
@@ -8,16 +8,17 @@ use types::CanisterId;
 pub struct Args {
     pub ledger_canister_id: CanisterId,
     pub token_standard: TokenStandard,
-    pub info_url: Option<String>,
-    pub how_to_buy_url: Option<String>,
-    pub transaction_url_format: Option<String>,
-    pub logo: Option<String>,
+    pub info_url: Option<String>,               // Not needed for SNS tokens
```
megrogan, line 11: Perhaps we should have a different endpoint.

PR #4127, opened 2023-08-03
Wire up channel notifications

megrogan:


backend/canisters/notifications/api/src/queries/notifications_v1.rs
```diff
@@ -0,0 +1,21 @@
+use candid::CandidType;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use types::{IndexedEvent, NotificationEnvelopeV1, SubscriptionInfo, TimestampMillis, UserId};
+
+#[derive(CandidType, Serialize, Deserialize, Debug)]
```
hpeebles, line 6: Maybe just call it something like `notification_candid_gen` and it takes Empty and return a single Notification.
There's no need for the subscriptions to exist here or in the candid file.

megrogan: Fair enough

hpeebles (approved):

PR #4129, opened 2023-08-03
Browse 'n' other channels

megrogan:


frontend/app/src/i18n/en.json
```diff
@@ -891,6 +891,7 @@
         "joinCommunity": "Join community",
         "preview": "Preview",
         "browseChannels": "Browse channels",
+        "browseOtherChannels": "Browse {n} other channels",
```
hpeebles, line 894: Other languages, also what about when there's only 1 other channel? It will say 'Browse 1 other channels'.
Maybe it should be 'View' rather than 'Browse'?

julianjelfs (approved):

PR #4137, opened 2023-08-04
Public channels have default behaviour

megrogan:


backend/canisters/community/impl/src/model/channels.rs
```diff
@@ -328,7 +290,7 @@ impl Channel {
             events_ttl: updates_from_events.events_ttl,
             gate: updates_from_events.gate,
             membership,
-            is_default: self.is_default.if_set_after(since).copied(),
+            is_default: None,
```
hpeebles, line 293: This should be false if is_public has changed to false.

hpeebles (approved):

PR #4164, opened 2023-08-09
Get token details from the Registry canister

hpeebles:


frontend/app/src/components/home/AccountInfo.svelte
```diff
@@ -4,16 +4,23 @@
     import { toastStore } from "../../stores/toast";
     import ContentCopy from "svelte-material-icons/ContentCopy.svelte";
     import { iconSize } from "../../stores/iconSize";
-    import type { CreatedUser } from "openchat-client";
+    import type { CreatedUser, OpenChat } from "openchat-client";
     import { copyToClipboard } from "../../utils/urls";
+    import { getContext } from "svelte";
+    import { ICP_SYMBOL } from "openchat-shared";
```
julianjelfs, line 10: I'm not sure it really matters but I don't think we generally import from openchat-shared in the app project. We always import from the openchat-client (which re-exports anything needed from shared). This would be nicer if / when we wanted to let third parties use openchat-client.

frontend/app/src/components/home/CryptoContent.svelte
```diff
@@ -14,16 +14,16 @@
...
+    $: logo = $cryptoLookup[content.transfer.ledger].logo;
     $: transferText = client.buildCryptoTransferText($_, user.userId, senderId, content, me);
     $: transactionLinkText = client.buildTransactionLink($_, content.transfer);
 </script>
 
 {#if transferText !== undefined}
     <div class="message">
         <div class="logo-wrapper">
-            <div class={`logo ${token.toLowerCase()}`} />
+            <img class="logo" src={logo} />
```
julianjelfs, line 26: we can remove all the token specific css classes here as well I think.

frontend/app/src/components/home/profile/Accounts.svelte
```diff
@@ -73,29 +76,23 @@
     <div />
 
     {#each crypto as token}
-        <div class={`icon ${token.key.toLowerCase()}`} />
+        <img class="icon" src={token.logo} />
```
julianjelfs, line 79: unused styles that can be deleted here too

julianjelfs (approved):

PR #4170, opened 2023-08-10
Create guide for local setup

0xkookoo:


hpeebles:
Now that running `npm run dev` will install and run the frontend code I think our current instructions are fine as is, so I don't think this PR is needed any more, so I'm going to close it.
Thanks for helping though!

PR #4173, opened 2023-08-11
Add Prerequisites to README + run `npm i` before building

hpeebles:


frontend/package.json
```diff
@@ -3,12 +3,12 @@
   "version": "1.0.0",
   "description": "",
   "scripts": {
-    "build": "turbo run build",
-    "build:ci": "turbo run build:ci",
-    "deploy:prod_test": "turbo run deploy:prod_test",
-    "deploy:prod": "turbo run deploy:prod",
-    "deploy:testnet": "turbo run deploy:testnet",
-    "dev": "turbo run dev",
+    "build": "npm i && turbo run build",
```
julianjelfs, line 6: we should probably use `npm ci` for everything expect dev. 

We should also remove the `npm ci` step from the frontend github actions workflow.

julianjelfs (approved):

PR #4203, opened 2023-08-17
Add root tsconfig.json containing common settings

hpeebles:


frontend/openchat-worker/tsconfig.json
```diff
@@ -1,26 +1,9 @@
...
-        "noUnusedParameters": true,
-        "outDir": "lib",
-        "paths": {},
-        "resolveJsonModule": true,
-        "rootDir": "./src",
-        "sourceMap": true,
-        "strict": true,
-        "target": "es2020",
-        "types": ["node"]
+        "module": "System",
```
julianjelfs, line 4: does this actually still work at runtime (as well as just building)?

hpeebles: Fixed. Now it works using tsc and has the correct file name.

PR #4204, opened 2023-08-17
Fix Svelte validation issues

hpeebles:


frontend/app/src/components/home/ChatSummary.svelte
```diff
@@ -145,7 +144,7 @@
         delOffset = -60;
     }
 
-    $: chat = normaliseChatSummary($now, chatSummary, $typersByContext);
+    $: chat = normaliseChatSummary(chatSummary, $typersByContext);
```
julianjelfs, line 147: this changes the behaviour. By passing in $now we ensure that the function runs when $now changes (even though its value is not being directly used). Not sure if this is essential, but I'm sure that was intentional.

frontend/app/src/components/home/thread/ThreadHeader.svelte
```diff
@@ -31,13 +30,13 @@
 
     $: byContext = client.typersByContext;
     $: userStore = client.userStore;
-    $: chat = normaliseChatSummary($now, chatSummary, $byContext);
+    $: chat = normaliseChatSummary(chatSummary, $byContext);
```
julianjelfs, line 33: another one

frontend/app/src/components/home/ChatSummary.svelte
hpeebles: Ahh ok, it was complaining about it not being used.

hpeebles: Fixed

julianjelfs (approved):

PR #4208, opened 2023-08-18
Avoid outputting `worker.js` and `push_sw.js` to `app/build`

hpeebles:


frontend/app/rollup.config.js
```diff
@@ -359,26 +348,33 @@ export default {
...
-                        transform: (contents, filename) =>
-                            contents
-                                .toString()
-                                .replace("//# sourceMappingURL=", "//# sourceMappingURL=./_/raw/"),
-                    },
-                ],
-                hook: "writeBundle",
-            }),
+        // Pull in the worker and push_sw
+        copy({
```
julianjelfs, line 352: Now that I see this I have a nasty feeling that I tried this and it didn't work in some way (maybe in dev mode or something) and that's why I flipped it round the other direction. Maybe it won't be a problem any more.

julianjelfs (approved):

frontend/app/rollup.config.js
hpeebles: Seems fine to me locally and on prod test.

PR #4216, opened 2023-08-22
memefighter integration

julianjelfs:


frontend/app/src/i18n/en.json
```diff
@@ -554,6 +554,9 @@
     "offline": "Offine",
     "sendTo": "Send to...",
     "sendGif": "Send gif",
+    "makeMeme": "Meme Fighter",
```
hpeebles, line 557: Other languages... although Meme Fighter should probably not be translated.

frontend/app/src/components/home/MemeFighterContent.svelte
```diff
@@ -0,0 +1,226 @@
+<svelte:options immutable />
```
megrogan, line 1: A lot of code common to ImageContent - could this somehow be re-used by having an image withg an empty caption?

julianjelfs: Yes - I wasn't sure how similar they would be but you're right, they are all but the same.

megrogan (approved):

PR #4218, opened 2023-08-22
allow to make community public

julianjelfs:


frontend/app/src/components/home/communities/edit/Edit.svelte
```diff
@@ -45,6 +45,7 @@
     let channelsValid = true;
     let detailsValid = true;
     let rulesValid = true;
+    let canMakePublic = !editing || client.hasOwnerRights(original.membership.role);
```
hpeebles, line 48: What about for making private? It doesn't seem like we check for having ownership rights anywhere for that...

hpeebles: As in, this should probably be called `canChangeVisibility`.

hpeebles: Ahh ok the `canMakePrivate` is worked out inside `VisibilityControl.svelte`.

hpeebles (approved):

PR #4265, opened 2023-08-30
Consolidate and simplify user/group/community name validation

hpeebles:


backend/libraries/utils/src/text_validation.rs
```diff
@@ -0,0 +1,184 @@
...
+const MAX_GROUP_DESCRIPTION_LENGTH: u32 = 1024;
+const MAX_GROUP_RULES_LENGTH: u32 = 1024;
+
+pub enum UsernameValidationError {
+    TooLong(u16),
+    TooShort(u16),
+    Invalid,
+}
+
+pub fn validate_username(username: &str, ascii_only: bool) -> Result<(), UsernameValidationError> {
```
megrogan, line 16: We might easily want to have different min/max lengths for username/displayname. In fact I think the min for display name should be shorter - say 3 - and the max for username should bd reduced to say 18.

hpeebles: This PR didn't change any functionality in that area, so we could decide exactly what we want to do and make that change as a separate PR?

hpeebles: Display name validation is now updated.

megrogan (approved):

PR #4271, opened 2023-08-30
Implement ability to create and update `user_groups`

hpeebles:


upgrade_order.md
```diff
@@ -0,0 +1,6 @@
+LocalGroupIndex -> Community -> Website -> User
```
megrogan, line 1: LocalUserIndex is missing despite having a CHANGELOG entry. Is that because it can go in any order?

hpeebles: LocalUserIndex + GroupIndex can go whenever

megrogan (approved):

PR #4273, opened 2023-08-30
Set `requireInteraction` to true on notifications

hpeebles:


hpeebles:
Let's hold off on getting this in.
I'm going to try setting notification type to Alert for Chrome and see if that is better.

PR #4280, opened 2023-08-31
Display names on the website

megrogan:


frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -891,11 +892,6 @@
         }
     }
 
-    .username {
```
julianjelfs, line 894: Is this intentional - seems like the element with this class is still there

frontend/openchat-push/src/push_sw.ts
```diff
@@ -127,136 +127,136 @@ async function showNotification(notification: Notification, id: string): Promise
...
     let closeExistingNotifications = false;
-    if (notification.kind === "direct_notification") {
+    if (n.kind === "direct_notification") {
         const chatId: DirectChatIdentifier = {
             kind: "direct_chat",
-            userId: notification.sender.userId,
+            userId: n.sender.userId,
         };
-        title = notification.senderName;
+        title = n.senderDisplayName ?? n.senderName;
```
julianjelfs, line 135: why do we send back two things (displayThing ?? thing) when all we just do is decide which one to display. Why can't the back end just send back whatever we are going to display so we don't have to do that?

julianjelfs (commented):
Why do we have to pass display name into all of these service calls. That's pretty horrible. 

Very nice to see the back of `PartialUserSummary` though I can't really follow how that's ok.

frontend/app/src/components/DisplayNameInput.svelte
```diff
@@ -0,0 +1,45 @@
+<script lang="ts">
+    import { isDisplayNameValid } from "openchat-client";
```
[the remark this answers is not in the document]
megrogan: We do in some cases. It just seemed like lots of extra unnecessary plumbing but I take the point. I'll make the change.

frontend/app/src/components/home/ChatMessage.svelte
megrogan: I don't think this will take effect right? There is already css for .username scoped to .message-bubble.

frontend/openchat-push/src/push_sw.ts
megrogan: Well this is moot. We now pass in sender display_name in addition to sender username into various send message functions so that the notification message can be constructed. It so happens the client has decided to only render one thing but it might choose to render both display name and @username in the notifcations. On the other hand we should probably just remove the new display_name args and instead the client can pass (display_name ?? username) into the username field and the notifications can just use username as before.

megrogan:
> Very nice to see the back of `PartialUserSummary` though I can't really follow how that's ok.

The only place the backend returned this was from `user_index::users` but this was for some legacy reason (@hpeebles?). We were always passing `Some(username)` so there was no longer a need for it to be optional and no longer a need for a different type. There is now a `user_index.users_v2` endpoint which returns a `UserSummary`. Once this PR is in we can remove the `users` endpoint and `PartialUserSummary` altogether from the backend.

hpeebles:
> > Very nice to see the back of `PartialUserSummary` though I can't really follow how that's ok.
> 
> 
> 
> The only place the backend returned this was from `user_index::users` but this was for some legacy reason (@hpeebles?). We were always passing `Some(username)` so there was no longer a need for it to be optional and no longer a need for a different type. There is now a `user_index.users_v2` endpoint which returns a `UserSummary`. Once this PR is in we can remove the `users` endpoint and `PartialUserSummary` altogether from the backend.

Yeah it used to be optional because we'd return the last online date + the username. And 99% of the time only the last online date would have changed. But now the last online dates are in a separate canister so we'd only return the user if the username had changed, so it no longer needed to be optional.

hpeebles:
Although now we have username and display name so maybe both should be optional. But then PartialUserSummary still shouldn't really exist on the FE since we'd still always have the username for each user.

frontend/openchat-push/src/push_sw.ts
megrogan: On second thoughts I think it is better to keep it the way it has been done. For direct chats we don't need to send the username/display name because the user canister already knows them and we don't want the user canister to decide how to render displayname+username.

frontend/app/src/components/DisplayNameInput.svelte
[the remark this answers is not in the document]
megrogan: Done

julianjelfs (approved):

PR #4281, opened 2023-08-31
Make mentions work without needing to select the user

hpeebles:


megrogan (approved):

frontend/openchat-client/src/utils/list.ts
```diff
@@ -92,29 +92,27 @@ export function toRecord<T, K extends string | number | symbol>(
     xs: T[],
     keyFn: (x: T) => K
 ): Record<K, T> {
-    return toRecordFiltered(xs, keyFn, _ => true);
+    return toRecord2(xs, keyFn, (x) => x);
```
hpeebles, line 95: There's no logic change to this file.
`toRecord` now calls into `toRecord2` which then calls into `toRecordFiltered`, so the core logic is only implemented once.

PR #4283, opened 2023-08-31
Improve how we build the user lookup for mentions

hpeebles:


julianjelfs (approved):
Looks fine and definitely a good improvement, but I wonder, given that we need to go through all the members and we are doing so in more than one place, it might be still more efficient and simple if the reverse lookup were simply a derived store based on currentChatMembers held in the openchat client? Then the openchat client can just expose a lookup function and the components never have to worry about if or when to build the lookup.

PR #4284, opened 2023-09-01
Store `userLookupForMentions` in openchat-client

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -4571,20 +4573,20 @@ export class OpenChat extends OpenChatAgentWorker {
...
     }
 
-    buildUserLookupForMentions(includeSelf: boolean): Record<string, UserSummary> {
-        const map = {} as Record<string, UserSummary>;
-        const userStore = this._liveState.userStore;
-        const currentUser = this._user?.userId;
-        for (const member of this._liveState.currentChatMembers) {
-            const userId = member.userId;
-            if (includeSelf || userId !== currentUser) {
+    getUserLookupForMentions(): Record<string, UserSummary> {
```
julianjelfs, line 4576: There doesn't seem to be any reason to return the lookup. Even simpler to just have a `lookupUserByUsername` function right?

hpeebles: We need the full lookup to display the users in the mention picker

hpeebles:
Hold on there's a bug still...

frontend/openchat-client/src/openchat.ts
julianjelfs: Only because we don't have a `getUsersByUsernamePrefix` function on the openchat client. Just thinking that the more logic we can "hide" the better.

julianjelfs (approved):

PR #4330, opened 2023-09-06
fix rendering of tweets so the don't jiggle on and off

julianjelfs:


frontend/app/src/components/home/Tweet.svelte
```diff
@@ -34,11 +34,8 @@
                     theme: $themeStore.name,
                 })
                 .then(() => {
-                    // only render the preview if we are *still* intersecting
-                    if (intersecting && !$eventListScrolling) {
```
hpeebles, line 38: But I thought we decided we would only render previews if we still were intersecting?

hpeebles: The issue seems to be caused by it removing the preview and then re-adding it over and over again.

julianjelfs: > But I thought we decided we would only render previews if we still were intersecting?

We are triggering the render only if we are intersecting and not scrolling. The bit that decided to set or not set the rendered flag was pointless because the tweet would have been rendered whether we set the flag or not.

julianjelfs: > The issue seems to be caused by it removing the preview and then re-adding it over and over again.

Yes and that won't happen now

hpeebles (approved):

PR #4335, opened 2023-09-07
another attempt to fix the tweet glitching

julianjelfs:
+ restrict the size of rendered images on link preview because they sometimes look silly on desktop.

### DIFF
--- a/frontend/app/src/components/home/Tweet.svelte
+++ b/frontend/app/src/components/home/Tweet.svelte
@@ -12,8 +12,9 @@
     export let tweetId: string;
 
     let tweetWrapper: HTMLDivElement | undefined;
-    let tweetRendered = false;
     let supported = false;
+
+    let rendering: Promise<any> | undefined = undefined;
 
     onMount(() => {
         supported = (<any>window).twttr !== undefined;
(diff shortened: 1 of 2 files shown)


megrogan (approved):

frontend/app/src/components/home/Tweet.svelte
```diff
@@ -24,21 +25,23 @@
...
-                })
+
+            rendering = (<any>window).twttr?.widgets.createTweet(tweetId, tweetWrapper, {
+                conversation: "none",
+                theme: $themeStore.name,
+            }) as Promise<any>;
+
+            rendering
                 .then(() => {
-                    tweetRendered = true;
```
hpeebles, line 37: What prevents this from rendering multiple times?

julianjelfs: It will only run if `rendering` is falsy which it will only be if rendering a) has not started or b) failed

hpeebles: Ahhh I didn't realise setting `rendering` to undefined only happened if there was an error.

hpeebles (approved):

PR #4345, opened 2023-09-11
New Theme: Tokyo Night

archetypal-one:
This is a theme I made a while back based on the original OC colors from beta, and wanted to share with others.


hpeebles:
Nice!
In order for your theme to be selectable you'll need to add it to this list - https://github.com/open-chat-labs/open-chat/blob/master/frontend/app/src/theme/themes.ts#L16

archetypal-one:
Okay, I think I did it right and finally got the colors correct. Let me know if there's anything else.

julianjelfs:
Great. I will take a look soon and double check if there are any recent ui elements that might be overlooked etc.

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+    base.label = "Tokyo Night";
+    base.burst = true;
+    base.bg = "radial-gradient(circle, rgba(242,154,216,1) 5%, rgba(10,10,10,1) 70%)";
+    base.bd = darkGray;
+    base.txt = white;
+    base["txt-light"] = txtLight;
+    base.icon.txt = darkGray;
+    base.icon.selected = raspberry;
+    base.button.bg = raspberry;
+    base.button.hv = rasperry;
```
julianjelfs, line 27: There is typo here

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+    base.button["hv-txt"] = darkGray;
+    base.button.txt = "rgba(255 255 255 / 80%)";
+    base.button.disabled = silver;
+    base.collapsible.closed.header.txt = txt;
```
julianjelfs, line 31: This `txt` value is used in several places but it's not defined. I'm sort of guessing it should be white.

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+    base.collapsible.open.header.arrow = white;
+    base.accent = raspberry;
+    base.panel.left.bg = veryDarkGray;
+    base.panel.right.modal = lighterDarkGray;
+    base.modal.bd = darkGray;
+    base.modal.bg = lighterDarkGray;
+    base.chatSummary["bg-selected"] = "rgb(24,24,24 / 40%)";
```
julianjelfs, line 38: This is _very_ subtle to the point where I don't think you can really see which chat you have selected.

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+    base.burst = true;
+    base.bg = "radial-gradient(circle, rgba(242,154,216,1) 5%, rgba(10,10,10,1) 70%)";
+    base.bd = darkGray;
+    base.txt = white;
+    base["txt-light"] = txtLight;
+    base.icon.txt = darkGray;
+    base.icon.selected = raspberry;
+    base.button.bg = raspberry;
+    base.button.hv = rasperry;
+    base.button["hv-txt"] = darkGray;
```
julianjelfs, line 28: I think the button hover text would be better as white. Having it dim on hover feels a bit odd to my eye.

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+    base.menu.bd = white;
+    base.menu.bg = lighterDarkGray;
+    base.menu.txt = txt;
+    base.menu.warn = txt;
+    base.menu["disabled-txt"] = white;
+    base.icon.inverted.txt = silver;
+    base.menu.hv = raspberry;
+    base.reaction.me = white;
+    base.reaction.bg = veryDarkGray;
+    base.primary = raspberry;
```
julianjelfs, line 65: Might be a bit of a problem having the primary colour set to the same as the background of various elements. It's quite likely that there will be things that should stand out which won't. One example I can see is that the drop down arrow on select components is not visible because both the arrow and the background are set to `raspberry`. 

Maybe it's fine.

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+    base.bd = darkGray;
+    base.txt = white;
+    base["txt-light"] = txtLight;
+    base.icon.txt = darkGray;
+    base.icon.selected = raspberry;
+    base.button.bg = raspberry;
+    base.button.hv = rasperry;
+    base.button["hv-txt"] = darkGray;
+    base.button.txt = "rgba(255 255 255 / 80%)";
+    base.button.disabled = silver;
```
julianjelfs, line 30: I would suggest adding: `base.button["disabled-txt"] = darkGray;` otherwise the disabled buttons are quite hard to read.

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+const lighterDarkGray = "#101010";
+const txtLight = offWhite;
+const notificationAccent = silver;
+
+export function getTheme(base: Theme): Theme {
+    // it's ok to mutate the theme passed in because it's a clone
+    base.name = "tokyonight";
+    base.label = "Tokyo Night";
+    base.burst = true;
+    base.bg = "radial-gradient(circle, rgba(242,154,216,1) 5%, rgba(10,10,10,1) 70%)";
```
julianjelfs, line 20: I'm not quite sure what's gone wrong with this background but it doesn't seem to look anything like your screenshot. 

Something like this is a lot closer: 

```
    base.bg = "radial-gradient( circle, #69173e 5%, rgb(6 1 22) 100%)";
```

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
+import type { Theme } from "../types";
+
+// Colors
+const white = "#FFFFFF";
+const offWhite = "#DFDFDF";
+const black = "#000000";
```
julianjelfs, line 6: This is not used so should be removed

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,77 @@
...
+const white = "#FFFFFF";
+const offWhite = "#DFDFDF";
+const black = "#000000";
+const raspberry = "#971E57";
+const darkGray = "#696969";
+const silver = "#C0C0C0";
+const veryDarkGray = "#1e2123";
+const lighterDarkGray = "#101010";
+const txtLight = offWhite;
+const notificationAccent = silver;
```
julianjelfs, line 13: This is not used so should be removed

julianjelfs (changes requested):
Couple of changes that we need to make and a couple that we _might_ like to make.

frontend/app/src/theme/community/tokyonight.ts
archetypal-one: This is the only one I couldn't figure out. What would you change here?

frontend/app/src/theme/community/tokyonight.ts
julianjelfs: the `const black` is still there and needs to be removed. It doesn't really matter but it's not going pass the build with unused variables.

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,76 @@
...
+    base.button.txt = "rgba(255 255 255 / 80%)";
+    base.button.disabled = silver;
+    base.collapsible.closed.header.txt = txt;
+    base.collapsible.open.header.arrow = white;
+    base.accent = raspberry;
+    base.panel.left.bg = veryDarkGray;
+    base.panel.right.modal = lighterDarkGray;
+    base.modal.bd = darkGray;
+    base.modal.bg = lighterDarkGray;
+    base.chatSummary["bg-selected"] = "rgb(54,59,62 / 50%)";
```
julianjelfs, line 37: this is actually slightly the wrong syntax so this variable is not working. The commas are the problem. Should be: 

```
rgb(54 59 62 / 50%)
```

frontend/app/src/theme/community/tokyonight.ts
```diff
@@ -0,0 +1,76 @@
...
+    base.chatSummary.hv = "rgb(24,24,24 / 80%)";
```
julianjelfs, line 38: Same here - this needs to be: 

```
rgb(24 24 24 / 80%)
```

julianjelfs (approved):

PR #4355, opened 2023-09-11
Only allow a single media element to play at a time

hpeebles:


julianjelfs (commented):
There is actually an example in the svelte docs to do exactly this: https://learn.svelte.dev/tutorial/sharing-code

hpeebles:
> There is actually an example in the svelte docs to do exactly this: https://learn.svelte.dev/tutorial/sharing-code

That example only works if a single instance of that single component can play media. But in our case there can be multiple AudioContent components and multiple VideoContent components, any of which could be playing.

hpeebles:
> > There is actually an example in the svelte docs to do exactly this: https://learn.svelte.dev/tutorial/sharing-code
> 
> That example only works if a single instance of that single component can play media. But in our case there can be multiple AudioContent components and multiple VideoContent components, any of which could be playing.

Ah no actually I understood it wrong on first look. It may work...

julianjelfs (approved):

julianjelfs:
Just thinking we should probably pause media using an intersection observer as well so they stop when they go off the screen.

PR #4356, opened 2023-09-11
User groups

julianjelfs:
Allow communities to specify user groups. 

Support user groups in the mention pickers.


frontend/app/src/components/home/Markdown.svelte
```diff
@@ -31,14 +32,24 @@
         });
     }
 
+    function replaceUserGroupIds(text: string): string {
+        return text.replace(/@UserGroup\(([\d\w-]+)\)/g, (match, p1) => {
```
hpeebles, line 36: UserGroup Ids are always numbers so we only need the \d rather than \d\w-

frontend/openchat-client/src/openchat.ts
```diff
@@ -4723,30 +4767,51 @@ export class OpenChat extends OpenChatAgentWorker {
...
     }
 
-    getUserLookupForMentions(): Record<string, UserSummary> {
+    // the key might be a username or it might be a user group name
+    getUserLookupForMentions(
+        communityMembers?: Map<string, Member>,
+    ): Record<string, UserOrUserGroup> {
         if (this._userLookupForMentions === undefined) {
-            const lookup = {} as Record<string, UserSummary>;
+            const lookup = {} as Record<string, UserOrUserGroup>;
```
hpeebles, line 4775: This means that if a user group and a user have the same name you can only mention the user.
What if we returned an array instead? I think we always iterate these values so we don't actually need it to be a record.

julianjelfs: Surely it would be crazy to allow user groups and users to have the same name?

frontend/openchat-client/src/openchat.ts
```diff
@@ -4723,30 +4767,51 @@ export class OpenChat extends OpenChatAgentWorker {
         );
     }
 
-    getUserLookupForMentions(): Record<string, UserSummary> {
+    // the key might be a username or it might be a user group name
+    getUserLookupForMentions(
+        communityMembers?: Map<string, Member>,
```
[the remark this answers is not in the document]
julianjelfs: This is the age old problem of reactivity. We have to pass it in otherwise it won't get called if/when `communityMembers` changes.

frontend/openchat-client/src/openchat.ts
hpeebles: But at which point do we prevent it? A user group could be created, then a user with the same name joins the community. I guess it won't really work with name clashes though since we support typing "@name" manually. So maybe supporting duplicate names is unrealistic and we just leave it where user groups take priority.

frontend/openchat-client/src/openchat.ts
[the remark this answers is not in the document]
hpeebles: Ah right.
Could the userLookupForMentions itself be a derived store which derives from all the necessary stores and also a store that says whether or not the mention picker is active?
That way we only recalculate it when it is set to active, it is reactive to when any of its dependencies change, and we avoid having to pass any data into the client.

megrogan (changes requested):
It looks like you've prettified codegen generated js/ts which I don't think we want right?

julianjelfs:
> It looks like you've prettified codegen generated js/ts which I don't think we want right?

but I can't actually run codegen to undo that and it's not part of a single undoable commit so it's not very easy to rectify. And it also doesn't really matter does it?

megrogan (approved):

PR #4360, opened 2023-09-12
Fix marking latest messages read when multiple messages are deleted

hpeebles:


frontend/app/src/components/home/AggregateCommonEvents.svelte
```diff
@@ -85,7 +85,7 @@
                 class="deleted"
                 title={$_("expandDeletedMessages")}
                 bind:this={deletedMessagesElement}
-                data-index={messagesDeleted.join(" ")}
+                data-index={Math.max(...messagesDeleted)}
```
julianjelfs, line 88: Tiny bit worried about whether that has any implications for [this](https://github.com/open-chat-labs/open-chat/blob/master/frontend/app/src/components/home/ChatEventList.svelte#L376)

PR #4383, opened 2023-09-15
FE - Add versioned rules

megrogan:


julianjelfs (approved):

megrogan:
I won't merge until user canisters have been released

frontend/app/src/components/home/Rules.svelte
```diff
@@ -40,6 +48,19 @@
         maxlength={MAX_RULES_LENGTH}
         rows={8}
         placeholder={interpolateLevel("rules.placeholder", level, true)} />
+    {#if editing && rules.enabled}
+        <Toggle
```
julianjelfs, line 52: Not sure I get the idea here? Are we allowing the user to _decide_ whether the change they have made to the rules should represent a new version of the rules which much therefore be accepted by people again?

frontend/app/src/components/home/thread/Thread.svelte
```diff
@@ -28,6 +28,21 @@
     import { randomSentence } from "../../../utils/randomMsg";
     import TimelineDate from "../TimelineDate.svelte";
     import { reverseScroll } from "../../../stores/scrollPos";
+    import AcceptRulesModal from "../AcceptRulesModal.svelte";
+
+    type ConfirmedActionEvent = ConfirmedSendMessage | ConfirmedRetrySendMessage;
```
julianjelfs, line 33: Same question about editing messages here

frontend/openchat-agent/src/services/user/user.client.ts
```diff
@@ -233,8 +232,9 @@ export class UserClient extends CandidService {
 
     createCommunity(
         community: CommunitySummary,
-        rules: AccessRules,
-        defaultChannels: string[]
+        rules: Rules,
+        defaultChannels: string[],
+        defaultChannelRules: Rules,
```
julianjelfs, line 237: Why do we need to pass these in given that there is no way in the UI to set them?

frontend/openchat-client/src/openchat.ts
```diff
@@ -231,98 +231,100 @@ import { getTypingString, startTyping, stopTyping } from "./utils/chat";
...
+    type CryptocurrencyTransfer,
+    type Mention,
+    type SetDisplayNameResponse,
+    type UserGroupDetails,
+    type CreateUserGroupResponse,
+    type UpdateUserGroupResponse,
+    type SetMemberDisplayNameResponse,
+    type UserOrUserGroup,
+    type UpdatedRules,
+    defaultChatRules,
```
julianjelfs, line 327: Would be better to add this to the other non-type import from `openchat-shared`

frontend/openchat-client/src/openchat.ts
```diff
@@ -2691,36 +2703,50 @@ export class OpenChat extends OpenChatAgentWorker {
...
             user: this.user,
             mentioned: [],
             event,
+            threadRootMessageIndex,
+            rulesAccepted,
+            communityRulesAccepted
         })
             .then(([resp, msg]) => {
                 if (resp.kind === "success") {
                     this.onSendMessageSuccess(chatId, resp, msg, undefined);
```
julianjelfs, line 2720: should be passing threadRootMessageIndex here if we want this to work in threads too.

frontend/openchat-client/src/openchat.ts
```diff
@@ -2906,31 +2936,70 @@ export class OpenChat extends OpenChatAgentWorker {
...
+        const chatRules = this._liveState.currentChatRules;
+        const chat = this._liveState.selectedChat;
+        if (chat === undefined || chatRules === undefined) {
+            return false;
+        }
+
+        const communityRules = this._liveState.currentCommunityRules;
+        const community = this._liveState.selectedCommunity;
+
+        console.log(
```
julianjelfs, line 2980: console.debug would be less annoying. With some easily filterable prefix like RULES.

frontend/openchat-shared/src/domain/access.ts
```diff
@@ -61,18 +61,24 @@ export type AccessControlled = {
...
     text: string;
     enabled: boolean;
 };
 
+export type UpdatedRules = Rules & {
+    newVersion: boolean;
+};
 
-export function defaultAccessRules(level: Level): AccessRules {
+export function defaultChatRules(level: Level): VersionedRules {
```
julianjelfs, line 77: Definitely don't get this rename

julianjelfs (commented):
A few detailed thing but the kind of big overall question I have is why do we need to pass the accepted rule version into all of these send message functions? Seems quite messy. Could accepting the rules not just be a separate operation (called by the AcceptRulesModal)? Then we would just call `sendMessage` again without any extra args and the backend would already know what version of the relevant rules that we have accepted. 

Otherwise we end up with another horrible signature where we have these two optional param when optional is not really the right word. They are mandatory when they're mandatory. It just seems a bit nasty.

PR #4392, opened 2023-09-18
Post user canister release

megrogan:


hpeebles:
You should also remove the serde(deserialize_from = ...) bits

hpeebles (approved):

PR #4398, opened 2023-09-19
Optimise `CachePrimer` to skip chats where cache is already up to date

hpeebles:


frontend/openchat-agent/src/utils/caching.ts
```diff
@@ -752,3 +766,22 @@ function makeChatSummarySerializable<T extends ChatSummary>(chat: T): T {
         latestMessage: makeSerialisable(chat.latestMessage, chat.id, true),
     };
 }
+
+async function readAll<Name extends StoreNames<ChatSchema>>(
```
julianjelfs, line 770: What's the idea with using a cursor when we are actually getting _all_ the records. Why not just `(await db).getAll(storeName)`?

frontend/openchat-client/src/openchat.ts
```diff
@@ -4404,16 +4403,15 @@ export class OpenChat extends OpenChatAgentWorker {
...
+                const chats = (chatsResponse.state.directChats as ChatSummary[])
                     .concat(chatsResponse.state.groupChats)
                     .concat(chatsResponse.state.communities.flatMap((c) => c.channels));
 
-                this.updateReadUpToStore(updatedChats);
-                const chats = getAllServerChats(this._liveState.globalState).values();
+                this.updateReadUpToStore(chats);
 
-                this._cachePrimer?.processChatUpdates(chats, updatedChats);
+                this._cachePrimer?.processChats(chats);
```
julianjelfs, line 4412: Does this mean, now that we are not using the global store to `getAllServerChats` that we could potentially move the cache primer entirely into the worker? That has always seemed like where it belongs. And would also be better if and when we can convert the worker to a service worker (so everything could be kept up to date offline).

frontend/openchat-agent/src/utils/caching.ts
hpeebles: That only returns the values. I want both the keys and the values.

frontend/openchat-client/src/openchat.ts
hpeebles: Ooooo yeah that sounds good let me try that.

julianjelfs (approved):

PR #4401, opened 2023-09-19
Enhanced theme selection

julianjelfs:
Allow the user to specify whether they want to system, light or dark and then specify a preferred light theme and a preferred dark theme. 

image


hpeebles (approved):

megrogan:
I wonder if "System" is the best word. Maybe "Auto"?
Light | Dark | Auto
Light | Dark | Alternate

julianjelfs:
> I wonder if "System" is the best word. Maybe "Auto"? Light | Dark | Auto Light | Dark | Alternate

I personally think System is better as it tells you where the value is coming from. Certainly better than alternate because is does not necessarily alternate (you may have your System set to always dark). We previously had the label "Inherit system theme" - we could have that same label as a tooltip as well I guess?

PR #4406, opened 2023-09-20
Remove `openchat-sw`

hpeebles:


julianjelfs (commented):
This seems a bit premature!

hpeebles:
> This seems a bit premature!

Well webtest + prod test are already switched over to the Dfinity service worker and I was thinking of doing prod too now that the website upgrade has gone out.

Any reason not to? Just more time for testing?

julianjelfs:
> 
> Any reason not to? Just more time for testing?

The last time I checked the caching definitely wasn't working correctly. Would like to check it again.

julianjelfs (approved):

PR #4411, opened 2023-09-21
Tidy up handling of environment variables

hpeebles:


frontend/turbo.json
```diff
@@ -1,29 +1,16 @@
...
     "$schema": "https://turborepo.org/schema.json",
     "globalEnv": [
-        "INTERNET_IDENTITY",
-        "NFID_URL",
         "DFX_NETWORK",
-        "IC_URL",
-        "II_DERIVATION_ORIGIN",
-        "CUSTOM_DOMAINS",
-        "STORAGE_INDEX_CANISTER",
-        "LEDGER_CANISTER_ICP",
```
julianjelfs, line 11: It seems to me that we could now change the value of one of the common env vars and turbo would _not_ invalidate the cache.

frontend/turbo.json
```diff
@@ -1,29 +1,16 @@
 {
     "$schema": "https://turborepo.org/schema.json",
     "globalEnv": [
-        "INTERNET_IDENTITY",
```
julianjelfs, line 4: Shouldn't we now have BUILD_ENV in this list?

frontend/turbo.json
hpeebles: It would count as an input file change, so it would still invalidate the cache, but only for the app build, which is exactly what we want.

frontend/turbo.json
hpeebles: Turbo only loads env variables from the .env file in the frontend directory, so it isn't aware of BUILD_ENV.
Also if we change the BUILD_ENV in any of the build scripts that would count as an input file change and so would invalidate the cache.

frontend/app/common_env
```diff
@@ -0,0 +1,7 @@
+export LEDGER_CANISTER_ICP=ryjl3-tyaaa-aaaaa-aaaba-cai
```
julianjelfs, line 1: if these things are common why are they not just hard coded? Then the cache invalidation will just work. If we change them, it's just a code change. It's not like we can modify them without a release _anyway_.

julianjelfs: They can just be constants in the code.

hpeebles: Yeah we could hard code them, I just thought it might be easier to keep them all in a common file

julianjelfs: But they are no less "in a common file" if they are in a constants module. Seems like less unnecessary plumbing.

hpeebles: Sorted.

frontend/turbo.json
```diff
@@ -1,29 +1,16 @@
...
-        "BLOB_URL_PATTERN",
         "NODE_ENV",
         "USERGEEK_APIKEY",
         "METERED_APIKEY",
         "ROLLBAR_ACCESS_TOKEN",
         "TRANSLATE_API_KEY",
         "PUBLIC_TRANSLATE_API_KEY",
         "GIPHY_APIKEY",
-        "FRAME_ANCESTORS"
+        "FRAME_ANCESTORS",
```
julianjelfs, line 12: FRAME_ANCESTORS should not be in here any more.

hpeebles: Once your PR is in I'll pull latest into here and remove it during merge conflict resolution

julianjelfs (approved):

PR #4420, opened 2023-09-22
Implement tipping messages

hpeebles:


backend/canisters/user/api/can.did
```diff
@@ -117,6 +117,28 @@ type RemoveReactionResponse = variant {
     UserSuspended;
 };
 
+type TipMessageArgs = record {
+    chat : Chat;
+    recipient : UserId;
```
julianjelfs, line 122: Why do we need the recipient when we have the messageId?

hpeebles: We use the recipient to validate the transfer is going to the right place, we don't actually need it for anything else so we could remove it, I just left it in their as a safety net in case the FE every passes up the wrong transfer details.

hpeebles: Actually we can extract the userId from the transfer and then validate that matches the message sender, so we can remove it.

julianjelfs: But we are tipping the _message_ - surely the recipient comes from the message and if we verify anything it's that the transfer matches the sender of the message.

hpeebles: Yeah we verify that the recipient matches the message sender, but the message sender isn't known by the user canister in the case of a group or channel message so I was taking the recipient arg in order to verify the transfer is correct before we make the transfer.

hpeebles: But having the recipient doesn't really make things any better as the FE could pass up the incorrect recipient along with a matching incorrect transfer, the user canister would think everything is fine and so make the transfer and forward the tip on the group, and only then would the sender validation work correctly.
So I'll remove the recipient, extract the user_id from the transfer, then only add the tip if the user_id matches the message sender.
That way if someone hacks the FE they could cause the payment to be made, but no tip will actually display unless the transfer is to the message's sender.

julianjelfs:
Does this include the change to the `Message` type to include the tips that it has received? I can't see that anywhere.

backend/libraries/types/can.did
```diff
@@ -528,6 +528,7 @@ type Message = record {
     content : MessageContent;
     replies_to : opt ReplyContext;
     reactions : vec record { text; vec UserId };
+    tips : vec record { CanisterId; vec record { UserId; nat } };
```
hpeebles, line 531: @julianjelfs here

hpeebles: The canisterId is the ledger canister, then for each ledger there is a list of UserId + amount tipped pairs.

julianjelfs: OK so that's like a list of tuples? So if the same user tips twice we get two (and we can either choose to aggregate those or not on the front end)?

hpeebles: Its a list of tuples but it already aggregates on the backend, so you only see the number of unique users.

julianjelfs (approved):
Let's give it a go

PR #4427, opened 2023-09-22
Implement notifications for message tips

hpeebles:


megrogan (approved):

backend/libraries/types/src/notifications.rs
```diff
@@ -223,5 +223,5 @@ fn notification_length() {
 
     let bytes = candid::encode_one(notification).unwrap().len();
 
-    assert!(bytes < 630, "{bytes}");
+    assert!(bytes < 850, "{bytes}");
```
megrogan, line 226: It's growing...

megrogan (approved):

backend/libraries/types/src/notifications.rs
hpeebles: Yeah... not ideal. We'll fix it properly at some point.

PR #4435, opened 2023-09-25
Add follow/unfollow thread to `inspect_message`

megrogan:


hpeebles (approved):

backend/canisters/group/impl/src/lifecycle/inspect_message.rs
```diff
@@ -39,6 +39,8 @@ fn accept_if_valid(state: &RuntimeState) {
             | "change_role"
             | "claim_prize"
             | "edit_message_v2"
+            | "follow_thread"
+            | "unfollow_thread"
```
hpeebles, line 43: Alphabetical

megrogan: We have delete/undelete message one after the other so I assumed the ordering was more subtle

PR #4448, opened 2023-09-27
Fix unfollowing of threads on backend

megrogan:


backend/libraries/group_chat_core/src/members.rs
```diff
@@ -36,6 +36,7 @@ impl GroupMembers {
             notifications_muted: Timestamped::new(false, now),
             mentions: Mentions::default(),
             threads: HashSet::new(),
+            unfollowed_threads: HashSet::new(),
```
hpeebles, line 39: Maybe this would be better as a Vec?
If it was a Vec then we wouldn't have to iterate over the whole thing on each updates loop.
We could iterate backwards and exit once the thread hasn't been updated since whatever timestamp was passed in.

hpeebles (approved):

PR #4449, opened 2023-09-27
Follow/unfollow threads on frontend

megrogan:


frontend/openchat-agent/src/utils/chat.ts
```diff
@@ -72,159 +64,6 @@ function mergeThings<A, U>(
     return Object.values(updated);
 }
 
-export function mergeUpdates(
```
julianjelfs, line 75: Amazing that all this is not being used!

julianjelfs (approved):

PR #4458, opened 2023-09-28
All members can @everyone by default in private cxt

megrogan:
If it's a private community or group then the default permission for @everyone is all members


megrogan:
This isn't right yet...

hpeebles (approved):

hpeebles (approved):

PR #4459, opened 2023-09-28
Profile link component

julianjelfs:
This is a little trick to get @mentions that can be clicked and handled programmatically even though they have been rendered via markdown. 

As a side-effect we can make use of custom event bubbling and move _all_ profile trigger handling to the root component rather than duplicating it in a number of places.


megrogan (approved):
Awesome!

PR #4475, opened 2023-10-02
Set `lastCryptoSent` when sending a tip rather than upon completion

hpeebles:


frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -356,9 +356,7 @@
         const transfer = ev.detail;
         const currentTip = (msg.tips[transfer.ledger] ?? {})[client.user.userId] ?? 0n;
         client.tipMessage(messageContext, msg.messageId, transfer, currentTip).then((resp) => {
-            if (resp.kind === "success") {
-                lastCryptoSent.set(transfer.ledger);
```
megrogan, line 360: Should remove 
$: lastCryptoSent = client.lastCryptoSent;

megrogan (approved):

PR #4477, opened 2023-10-02
Avoid lots of unnecessary callbacks being executed

hpeebles:


frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -139,12 +138,6 @@
     $: lastCryptoSent = client.lastCryptoSent;
     $: tips = msg.tips ? Object.entries(msg.tips) : [];
 
-    afterUpdate(() => {
```
julianjelfs, line 142: I don't quite get where this is being unobserved now?

hpeebles: There are 3 ways a message can be marked read. Via the intersection observer, via 'Mark all as read', and by being marked as read on another device.
As of this PR we only unobserve messages when they are marked as read via the intersection observer (which is by far the most common case). That happens after calling `markMessageRead` in `ChatEventsList.svelte`.
If a message is read one of the other ways it will still be observed, but then when `markMessageRead` is called nothing will happen.
We can potentially optimise to stop observing in those other cases but the main concern was to avoid the masses of callbacks firing on every message on every update.

julianjelfs: ok - I see where it's being done now.

julianjelfs (approved):

PR #4486, opened 2023-10-03
Support submitting proposals from within OpenChat

hpeebles:


backend/canisters/user/impl/src/updates/submit_proposal.rs
```diff
@@ -0,0 +1,80 @@
...
+        Ok(PrepareResult {
+            my_user_id: state.env.canister_id().into(),
+            proposals_bot_canister_id: state.data.proposals_bot_canister_id,
+            transaction: PendingCryptoTransaction {
+                ledger: SNS_LEDGER_CANISTER_ID,
+                token: Cryptocurrency::CHAT,
+                amount: 4_0000_0000, // 4 CHAT
+                to: Account::from(state.data.proposals_bot_canister_id),
+                fee: Cryptocurrency::CHAT.fee().unwrap(),
+                memo: None,
```
megrogan, line 75: The memo is in the args so you might as well wire it up

hpeebles: This is the transfer to the proposals bot to cover the proposal rejection fee, not the transfer from the treasury.

megrogan (approved):

PR #4537, opened 2023-10-09
Ensure events to signal chat updated only affect intended chat

hpeebles:


julianjelfs (approved):
Looks good to me - if it works!

PR #4544, opened 2023-10-10
Fix submit proposal

megrogan:


frontend/app/src/components/home/MakeProposalModal.svelte
```diff
@@ -63,16 +62,17 @@
...
+    $: requiredFunds = proposalCost + transferFee + transferFee;
     $: insufficientFunds = cryptoBalance < requiredFunds;
     $: padding = $mobileWidth ? 16 : 24; // yes this is horrible
     $: left = step * (actualWidth - padding);
     $: token = treasury === "SNS" ? symbol : "ICP";
     $: titleValid = title.length >= MIN_TITLE_LENGTH && title.length <= MAX_TITLE_LENGTH;
     $: urlValid = url.length <= MAX_URL_LENGTH;
     $: summaryValid = summary.length >= MIN_SUMMARY_LENGTH && summary.length <= MAX_SUMMARY_LENGTH;
-    $: amountValid = isAmountValid(amount);
+    $: amountE8s = Number(amount) * Number(Math.pow(10, tokenDetails.decimals));
```
hpeebles, line 74: Shouldn't really be called e8s because there is no guarantee that these tokens have 8 decimals, for example DKP is about to switch to 3 decimals.

hpeebles: I think just call it `amount`

hpeebles (approved):

PR #4545, opened 2023-10-10
Fix `isFollowedByMe` check

hpeebles:


megrogan:
Can you make the same fix for checking whether a user can follow a thread?

megrogan (approved):

megrogan (approved):

PR #4560, opened 2023-10-12
Prize message changes

julianjelfs:
Make the winner message more discrete when you are _not_ the recipient. 

Give the option when creating a prize message to specify that it can only be claimed by diamond users.


frontend/openchat-agent/src/services/common/chatMappers.ts
```diff
@@ -1313,6 +1314,7 @@ export function apiPrizeContentInitial(domain: PrizeContentInitial): ApiPrizeCot
         caption: apiOptional(identity, domain.caption),
         transfer: apiPendingCryptoTransaction(domain.transfer),
         end_date: domain.endDate,
+        // diamond_only: domain.diamondOnly, // TODO once the back end is available
```
hpeebles, line 1317: This can go now that my PR is in

hpeebles (approved):

PR #4581, opened 2023-10-13
Support subaccount for transfer_sns_funds proposal

megrogan:


frontend/app/src/utils/sns.ts
```diff
@@ -11,3 +11,8 @@ export function isPrincipalValid(text: string): boolean {
     }
 }
 
+const SUB_ACCOUNT_REGEX = new RegExp("^[A-Fa-f0-9]{64}$");
```
hpeebles, line 14: Is this saying it has to be exactly 64 chars?
If so that is wrong, you are allowed to trim the 0's from the left.
So subaccount '1' is valid, we simply need to pad it with 0's

frontend/openchat-agent/src/services/user/mappers.ts
```diff
@@ -1071,8 +1073,8 @@ function proposalAction(action: CandidateProposalAction): ProposalToSubmitAction
         case "transfer_sns_funds":
             return { TransferSnsTreasuryFunds: {
                 to: {
-                    owner: Principal.fromText(action.toPrincipal),
-                    subaccount: []
+                    owner: Principal.fromText(action.recipient.owner),
+                    subaccount: apiOptional((hex) => hexStringToBytes(hex), action.recipient.subaccount)
```
hpeebles, line 1077: We need to pad the string with 0's before converting to bytes, on the BE this needs to be exactly 32 bytes.

hpeebles: Or add the 0's after converting to bytes...

megrogan: Yes of course

hpeebles (approved):

PR #4589, opened 2023-10-16
Wallet

julianjelfs:


frontend/app/src/components/home/profile/AccountsModal.svelte
```diff
@@ -11,7 +20,27 @@
         {$_("wallet")}
     </div>
     <div slot="body">
-        <Accounts />
+        <Accounts bind:showZeroBalance bind:zeroCount />
+    </div>
+    <div slot="footer">
+        <div class="footer">
+            <div class="show-more">
+                {#if zeroCount > 0}
```
hpeebles, line 28: Shouldn't this 'if' check be before the "show_more" div?
Or is the div always rendered for some display reason?

megrogan: Yes it is supposed to be like that...

frontend/app/src/components/home/profile/AccountsModal.svelte
```diff
@@ -24,8 +24,8 @@
     </div>
     <div slot="footer">
         <div class="footer">
-            <div class="show-more">
-                {#if zeroCount > 0}
+            {#if zeroCount > 0}
```
megrogan, line 27: This will now mean the close button will be left-aligned when there is no "show more" link.

PR #4613, opened 2023-10-18
Simplify updating a group or channel

hpeebles:


frontend/app/src/components/home/addgroup/NewGroup.svelte
```diff
@@ -334,7 +220,7 @@
             .then((resp) => {
                 if (resp.kind !== "success") {
                     const err = groupCreationErrorMessage(resp, isChannel);
-                    if (err) toastStore.showFailureToast(interpolateError(err));
+                    if (err) toastStore.showFailureToast(interpolateError(err, isChannel ? "channel" : "group"));
```
julianjelfs, line 223: not sure why we are not just using `candidateGroup.level` here?

julianjelfs: ok, but `const level = candidateGroup.level` would still a bit better I think

hpeebles: Done

PR #4634, opened 2023-10-20
Read cached events using an iterator

hpeebles:


frontend/openchat-agent/src/utils/cachedEventsIterator.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    static async create(
+        db: IDBPDatabase<ChatSchema>,
+        messageContext: MessageContext,
+        startIndex: number,
+        ascending: boolean,
+        bound: number,
+    ): Promise<EventsIterator> {
+        if ((ascending && startIndex > bound) || (!ascending && startIndex < bound)) {
+            return new EventsIterator();
```
julianjelfs, line 98: Is this not an error scenario? Should we be throwing?

hpeebles: Yeah it means the args are invalid. I'll make it throw

julianjelfs (approved):

PR #4666, opened 2023-10-29
Fine grained message permissions on FE

megrogan:
This is deployed to prod_test


frontend/app/src/components/home/addgroup/NewGroup.svelte
```diff
@@ -39,12 +39,7 @@
     let step = 0;
     let actualWidth = 0;
     let detailsValid = true;
-    let originalGroup = {
-        ...candidateGroup,
-        rules: { ...candidateGroup.rules },
-        permissions: { ...candidateGroup.permissions },
-        gate: { ...candidateGroup.gate },
-    };
+    let originalGroup = structuredClone(candidateGroup);
```
julianjelfs, line 42: Note: we really need to remember to test anything that uses `structuredClone` in safari (desktop and mobile)

frontend/app/src/components/home/groupdetails/GroupDetails.svelte
```diff
@@ -34,7 +34,8 @@
 
     // capture a snapshot of the chat as it is right now
     $: canEdit = client.canEditGroupDetails(chat.id);
-    $: canSend = client.canSendMessages(chat.id) || client.canReplyInThread(chat.id);
+    $: canSend =
+        client.canSendAnyMessages(chat.id, false) || client.canSendAnyMessages(chat.id, true);
```
julianjelfs, line 38: Note to self: come back to this ...

frontend/app/src/components/home/thread/Thread.svelte
```diff
@@ -151,7 +153,16 @@
     }
 
     function fileSelected(ev: CustomEvent<AttachmentContent>) {
-        draftThreadMessages.setAttachment(threadRootMessageIndex, ev.detail);
+        const content = ev.detail;
```
julianjelfs, line 156: Could we deal with all this inside `Footer` and only bubble the event if the file can actually be attached? That way we don't have this duplicated. Footer already knows which "mode" it is running in.

frontend/openchat-client/src/openchat.ts
```diff
@@ -1268,8 +1271,23 @@ export class OpenChat extends OpenChatAgentWorker {
         }
     }
 
-    canCreatePolls(chatId: ChatIdentifier): boolean {
-        return this.chatPredicate(chatId, canCreatePolls);
+    canSendMessage(chatId: ChatIdentifier, inThread: boolean, permission: MessagePermission): boolean {
```
julianjelfs, line 1274: I don't love the "boolean blindness" here. I think we should at least replace the `boolean` with something like `"thread" | "message"` but we could also go one better and have `"thread" | "message" | "any"`. This would support the scenario where we need to check both which _does_ happens elsewhere _and_ future proof us better given that it is entirely contingent that we only have _two_ possible contexts. 

Another possibility would be to pass in the `MessageContext` here rather than the chatId, then you wouldn't need the `inThread` param. But that wouldn't support the "any context" case.

julianjelfs, line 1274: could it be `permission?: MessagePermission` to cover the case where we don't care which specific type so that we don't actually need `canSendAnyMessages`?

frontend/openchat-client/src/openchat.ts
```diff
@@ -1268,8 +1271,23 @@ export class OpenChat extends OpenChatAgentWorker {
...
+        return this.chatPredicate(chatId, (chat) => canSendMessage(chat, inThread, permission));
+    }
+
+    permittedMessages(chatId: ChatIdentifier, inThread: boolean): Map<MessagePermission, boolean> {
```
julianjelfs, line 1278: same thoughts as above on the `inThread` param.

julianjelfs (changes requested):

frontend/app/src/components/home/MessageActions.svelte
```diff
@@ -111,48 +110,58 @@
...
+        {/if}
+        {#if permittedMessages.get("poll")}
             <div class="poll" on:click|stopPropagation={createPoll}>
                 <HoverIcon title={$_("poll.create")}>
                     <Poll size={$iconSize} color={"var(--icon-txt)"} />
                 </HoverIcon>
             </div>
         {/if}
-        {#if isMultiUser}
+        {#if permittedMessages.get("prize")}
```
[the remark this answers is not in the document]
megrogan: This is taken care of inside client.permittedMessages

frontend/app/src/components/home/addgroup/NewGroup.svelte
megrogan: ok

julianjelfs (approved):

PR #4671, opened 2023-10-30
Disable signals

julianjelfs:
it's a bit rough at the moment so let's keep it hidden for now. Can still be set via the iframe init message.


frontend/app/src/theme/light.ts
```diff
@@ -14,6 +14,7 @@ const disabledTxt = txtLight;
 
 export function lightTheme(): Theme {
     return {
+        hidden: false,
```
hpeebles, line 17: This field is only set on signals and light. Is that intentional?

julianjelfs: Yes, dark inherits light and all the other themes inherit from one of those two.

megrogan (approved):

PR #4675, opened 2023-10-30
Fix rendering of relative preview images + styling

megrogan:


frontend/app/src/components/home/GenericPreview.svelte
```diff
@@ -59,28 +59,36 @@
...
+                {/if}
+            </div>
         {/if}
     {/each}
 </div>
 
 <style lang="scss">
+    .preview {
+        margin-top: $sp3;
+        border-top: 1px solid currentColor;
```
julianjelfs, line 88: This is going to give a border above every preview? Is that what we want?

megrogan: I think so

megrogan:
The latest is on webtest

frontend/app/src/theme/types.ts
```diff
@@ -15,6 +15,7 @@ export interface Theme {
     bg: string;
     txt: string;
     "txt-light": string;
+    "txt-muted": string;
```
julianjelfs, line 18: I think it's confusing and will cause future problems to introduce a  `txt-muted` field that is actually only used for something that is _not_ muted text. I had in mind something specific to what this is actually used for. i.e. something like `currentChat.msg.separator`.  That will definitely cause fewer problems if themes want to do different things.

I think it may also be impossible to get away with a single value. You might need a different value for your own messages and for other users messages. I don't think it's alway going to work to use the same colour in both cases.

julianjelfs (approved):

julianjelfs (approved):

PR #4716, opened 2023-11-07
Disable disappearing messages until backend is upgraded

hpeebles:


megrogan (dismissed):

frontend/app/src/components/home/VisibilityControl.svelte
```diff
@@ -166,26 +166,26 @@
...
-                {#if disappearingMessages}
-                    <DisappearingMessages
-                        {canEditDisappearingMessages}
-                        bind:ttl={candidate.eventsTTL} />
-                {/if}
-            </div>
-        </Checkbox>
-    </div>
-{/if}
+<!--{#if candidate.kind === "candidate_group_chat"}-->
```
megrogan, line 169: It would be simpler just to comment out this #if and add an always false condition

megrogan (approved):

PR #4740, opened 2023-11-09
Improve unread threads handling

### LINKED ISSUE #4738: Make sure that unread thread count contributes to overall unread count for a scope
Currently if you have unread messages in e.g. 3 chats and also 2 unread threads, the overall unread count for that scope (e.g. groups chats) would be 3 and that's what you would see in the left nav because the unread threads are not included in that count. 

This means that if you _only_ have unread threads, you see nothing in the left nav at all and may well miss things that you should be seeing. 

Example
======
image

julianjelfs:
fixes #4738


julianjelfs:
Actually - this is not quite right yet - can be better.

hpeebles (approved):

PR #4744, opened 2023-11-09
Platform mods can del message when not group member

megrogan:


hpeebles:
Maybe worth having an integrtion test to show that the user index can delete a message?

PR #4753, opened 2023-11-10
Reload if user suspension status changes

megrogan:
User canisters need to be upgraded before this is merged


hpeebles (commented):
Are we handling it when they become unsuspended? Doesn't look like it...

hpeebles (approved):

PR #4765, opened 2023-11-14
Add new config option to polls

### LINKED ISSUE #4721: New poll option - permanent vote
Once you have voted, you have voted and you cannot change your vote.

julianjelfs:
fixes #4721


frontend/app/src/i18n/en.json
```diff
@@ -391,6 +391,7 @@
         "oneDay": "1 Day",
         "oneWeek": "1 Week",
         "allowMultipleVotes": "Allow multiple votes",
+        "allowChangeVotes": "Allow user to change vote",
```
hpeebles, line 394: "Allow users* ..." is probably better

frontend/app/src/components/home/PollContent.svelte
```diff
@@ -24,13 +24,16 @@
 
     $: numberOfVotes = totalVotes(content);
 
+    $: cannotVote =
+        content.ended || readonly || (haveIVoted && !content.config.allowUserToChangeVote);
```
hpeebles, line 28: If the config is set to allow multiple then they can still vote on other options.

julianjelfs: yes I thought it made sense that this new setting only applied for single vote polls

julianjelfs: Oh wait - I see what you mean

julianjelfs: although - if allow multiple is true, allow user to change will also always be true

hpeebles (approved):

PR #4775, opened 2023-11-15
`approve_transfer` takes fee from amount

megrogan:


hpeebles:
I think this is the wrong way of doing it. If you approve 2 CHAT then the spender should be able to send 2 CHAT.
We should simply ask to approve the joining fee amount minus the transfer fee.

PR #4805, opened 2023-11-17
few more bits

julianjelfs:


frontend/openchat-agent/src/services/candidService.ts
```diff
@@ -16,12 +16,12 @@ export abstract class CandidService {
...
         factory: IDL.InterfaceFactory,
         canisterId: string,
-        config: AgentConfig
+        config: AgentConfig,
     ): T {
         const host = config.icUrl;
         const agent = new HttpAgent({ identity: this.identity, host, retryTimes: 5 });
         const isMainnet = config.icUrl.includes("icp-api.io");
-        if (!isMainnet) {
+        if (!isMainnet && navigator.onLine) {
```
hpeebles, line 24: If we don't fetch the root key then the agent will be unusable... but given that this is only a test thing it doesn't really matter.

hpeebles (approved):

PR #4816, opened 2023-11-20
Support deleting direct chats (only for the current user)

hpeebles:


backend/canisters/user/api/can.did
```diff
@@ -225,6 +225,15 @@ type UnblockUserResponse = variant {
     UserSuspended;
 };
 
+type DeleteDirectChatArgs = record {
+    user_id : UserId;
```
megrogan, line 229: Perhaps it should take a block: bool too?

hpeebles: Done

megrogan (approved):

PR #4818, opened 2023-11-21
optimise the new message case

julianjelfs:
When we have a single new message (e.g. index 10) we will attempt to load a message window around message index 10. Because we do not have message index 10 in the cache, we will conclude a total cache miss and get *all* messages from the backend even though we actually _may_ have messages 1-9 in the cache. 

This change optimises this specific case and attempts to look up messageIndex - 1 as the center point. This means that we will retrieve messages 1-9 from the cache and only look up message index 10 from the backend. 

In itself, this is unlikely to be earth shattering, but it allows a phase two which is to immediately return the messages that we _have_ been able to get from the cache while loading those we haven't.


frontend/openchat-agent/src/utils/caching.ts
```diff
@@ -414,7 +415,34 @@ export async function getCachedEventsByIndex<T extends ChatEvent>(
...
+    messageIndex: number,
+    iterations = 0,
+): Promise<number | undefined> {
+    const eventIndex = await getCachedEventIndexByMessageIndex(db, context, messageIndex);
+    if (eventIndex === undefined && iterations === 0) {
+        console.debug(
+            "EV: we didn't find the event index for ",
+            messageIndex,
+            " recursing to look for event index for ",
+            messageIndex + 1,
```
hpeebles, line 433: - rather than +

julianjelfs: doh

hpeebles (approved):

PR #4822, opened 2023-11-22
Cached chat state

julianjelfs:
This will load the initial state in two chunks. First it will immediately return the cached state and then it will return the up to date state as soon as possible afterwards.


frontend/openchat-client/src/openchat.ts
```diff
@@ -4512,152 +4513,165 @@ export class OpenChat extends OpenChatAgentWorker {
...
-            });
-
-            if (!init || chatsResponse.anyUpdates) {
-                if (updateRegistryTask !== undefined) {
-                    // We need the registry to be loaded before we attempt to render chats / events
-                    await updateRegistryTask;
-                }
+    private async loadChatsResponse(
+        updateRegistryTask: Promise<void> | undefined,
+        init: boolean,
```
hpeebles, line 4518: I know this wasn't changed in this PR but I think calling this `init` is confusing and I've had to check what it means a few times in the past.
It could mean 'this is doing the initialization', but in our case it means 'things are initialized'.
So maybe we should call it `initialized`?

hpeebles: Or invert it and call it `initialLoad`?

frontend/openchat-client/src/openchat.ts
```diff
@@ -4512,152 +4513,165 @@ export class OpenChat extends OpenChatAgentWorker {
...
+        const updateRegistryTask = !init ? this.updateRegistry() : undefined;
+
+        return new Promise<void>((resolve) => {
+            this.sendStreamRequest({
+                kind: "getUpdates",
+                initialLoad: !init,
+            })
+                .subscribe(async (resp) => {
+                    await this.loadChatsResponse(updateRegistryTask, init, resp as UpdatesResult);
+                    chatsLoading.set(!this._liveState.chatsInitialised);
```
hpeebles, line 4663: Why not just `false`?

hpeebles: Is it something to do with there potentially being more coming?

frontend/openchat-client/src/openchat.ts
```diff
@@ -4512,152 +4513,165 @@ export class OpenChat extends OpenChatAgentWorker {
...
-            const chatsResponse = await this.sendRequest({
-                kind: "getUpdates",
-            });
-
-            if (!init || chatsResponse.anyUpdates) {
-                if (updateRegistryTask !== undefined) {
-                    // We need the registry to be loaded before we attempt to render chats / events
-                    await updateRegistryTask;
-                }
+    private async loadChatsResponse(
```
hpeebles, line 4516: Maybe `processChatsResponse`?
Because at this point the chats response has already been loaded.

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -641,6 +642,48 @@ export class OpenChatAgent extends EventTarget {
         }
     }
 
+    // chatEventsWindow(
```
hpeebles, line 645: Are these meant to be commented out?

hpeebles: Ahh these are newly added... I thought you'd commented out a load of existing code.

These can all be deleted.

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -42,138 +42,139 @@ import {
...
+    type UpdateUserGroupResponse,
+    type DeleteUserGroupsResponse,
+    type SetMemberDisplayNameResponse,
+    type UpdatedRules,
+    type FollowThreadResponse,
+    type CandidateProposal,
+    type SubmitProposalResponse,
+    type AccountTransactionResult,
+    type OptionalChatPermissions,
+    Stream,
```
hpeebles, line 177: We have 2 import blocks from `openchat-shared`, one is all the types and one is everything else. This should be in the block just below this one.

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -1581,6 +1677,28 @@ export class OpenChatAgent extends EventTarget {
         };
     }
 
+    getUpdates(initialLoad: boolean): Stream<UpdatesResult> {
+        return new Stream(async (resolve, reject) => {
+            const cachedState = await getCachedChats(this.db, this.principal);
+            if (cachedState && initialLoad) {
+                resolve(
```
hpeebles, line 1684: This is really cool!

frontend/openchat-client/src/openchat.ts
julianjelfs: just in case the `chatsInitialised` doesn't get set for some reason i.e. `chatsInitialised` is the source of truth.

frontend/openchat-client/src/openchat.ts
julianjelfs: I've gone with `handleChatsResponse` to mirror the `handleEventsResponse` that we already have.

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -771,20 +772,42 @@ export class OpenChatAgent extends EventTarget {
...
-                threadRootMessageIndex,
-                latestKnownUpdate,
-            ),
+            rawEvents,
             threadRootMessageIndex,
             latestKnownUpdate,
         );
     }
 
+    // private groupChatEventsWindow(
```
hpeebles, line 790: This can go

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -1111,6 +1134,39 @@ export class OpenChatAgent extends EventTarget {
         return ev;
     }
 
+    // private async rehydrateEventResponse2<T extends ChatEvent>(
```
hpeebles, line 1137: And here

hpeebles (approved):

hpeebles (approved):

PR #4825, opened 2023-11-22
refine offline tracking

julianjelfs:
this basically redefines offline to mean offline || very low bandwidth but also provides the abstraction to allow us to more easily adjust what being offline _means_

Note that the network information api is not available on iOS so this is only going to behave differently for android.


frontend/openchat-shared/src/constants.ts
```diff
@@ -1 +1,10 @@
 export const ONLINE_THRESHOLD = 120;
+export const MAX_MESSAGES = 100;
+export const MAX_EVENTS = 500;
+export const MAX_MISSING = 30;
+export const OPENCHAT_BOT_USER_ID = "zzyk3-openc-hatbo-tq7my-cai";
+export const OPENCHAT_BOT_USERNAME = "OpenChatBot";
+export const OPENCHAT_BOT_AVATAR_URL = "assets/robot.svg";
+
+// downlink is the effective bandwidth estimate in megabits per second, rounded to the nearest multiple of 25 kilobits per seconds.
+export const MIN_DOWNLINK = 2;
```
megrogan, line 10: I reckon this number should be lower... say 0.5

julianjelfs: It's a bit unclear what it actually _means_ - it doesn't actually seem to mean what I thought it did. I will try to pin it down exactly

megrogan (approved):

megrogan (approved):

PR #4839, opened 2023-11-23
Cache user

julianjelfs:
Store the current user in the cache. 

When the page loads return the cached user first if it is available and then the up to date user from the back end as soon as it's available. This should be another step towards rendering the initial UI as fast as we possibly can.


frontend/openchat-client/src/openchat.ts
```diff
@@ -3743,12 +3744,24 @@ export class OpenChat extends OpenChatAgentWorker {
...
-                selectedAuthProviderStore.init(AuthProvider.II);
-            }
-            return response;
+        return new Promise((resolve, reject) => {
+            let resolved = false;
+            this.sendStreamRequest({ kind: "getCurrentUser" })
+                .subscribe((user) => {
+                    if (user.kind === "created_user") {
+                        userCreatedStore.set(true);
+                        selectedAuthProviderStore.init(AuthProvider.II);
```
hpeebles, line 3753: Why is this hardcoded to II?

megrogan (approved):

frontend/openchat-client/src/openchat.ts
julianjelfs: It's a default that gets set if there is nothing in local storage

PR #4874, opened 2023-11-28
fix mobile

julianjelfs:
bit hacky and I'm not sure why this didn't seem to be a problem before but this seems to work now - deployed to webtest already.

### DIFF
--- a/frontend/openchat-shared/src/domain/crypto.ts
+++ b/frontend/openchat-shared/src/domain/crypto.ts
@@ -67,8 +67,8 @@ export const dollarExchangeRates: Record<string, number> = {
     nua: to2SigFigs(dollarToICP / 0.0147),
     sonic: to2SigFigs(dollarToICP / 0.0274),
     sneed: to2SigFigs(dollarToICP / 22.25),
-    exe: to2SigFigs(dollarToICP / 0.0152),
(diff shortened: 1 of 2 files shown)


hpeebles (approved):

frontend/openchat-shared/src/domain/crypto.ts
```diff
@@ -68,7 +68,7 @@ export const dollarExchangeRates: Record<string, number> = {
     sonic: to2SigFigs(dollarToICP / 0.0274),
     sneed: to2SigFigs(dollarToICP / 22.25),
     exe: to2SigFigs(dollarToICP / 0.0152),
-    taggr: to2SigFigs(dollarToICP / 1.0197),
+    taggr: to2SigFigs(dollarToICP / 0.6565),
```
hpeebles, line 71: This looks wrong... the price on icpcoints is 0.1851 ICP

hpeebles: But I think you picked the TAGGR/USD price rather than the TAGGR/ICP price.
It doesn't look like it has ever been as high as the price you put here.

julianjelfs: 🤦 - corrected

hpeebles (approved):

PR #4880, opened 2023-11-29
Top up NNS neuron when users pay ICP for lifetime Diamond membership

hpeebles:


backend/canisters/user_index/impl/src/lib.rs
```diff
@@ -211,15 +216,22 @@ struct Data {
...
     pub neuron_controllers_for_initial_airdrop: HashMap<UserId, Principal>,
+    #[serde(default = "nns_governance_canister_id")]
+    pub nns_governance_canister_id: CanisterId,
     pub internet_identity_canister_id: CanisterId,
     pub user_referral_leaderboards: UserReferralLeaderboards,
     pub platform_moderators_group: Option<ChatId>,
     pub reported_messages: ReportedMessages,
     pub fire_and_forget_handler: FireAndForgetHandler,
     #[serde(default)]
+    pub nns_neuron: Option<NnsNeuron>,
```
megrogan, line 227: Perhaps it should say nns_neuron_8_year? Or something like that.

hpeebles: Done

megrogan (approved):

PR #4886, opened 2023-11-29
control font-family with css vars

julianjelfs:
This gives theme authors a bit more flexibility. 

Still doesn't support the async loading of custom fonts e.g. google fonts but a bit better.

### DIFF
--- a/frontend/app/src/theme/community/windoge98.ts
+++ b/frontend/app/src/theme/community/windoge98.ts
@@ -31,6 +31,10 @@ export function getTheme(base: Theme): Theme {
     base.accent = magenta;
     base.disabledTxt = txt40;
     base.placeholder = txt60;
+    // base.font =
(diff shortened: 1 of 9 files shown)


frontend/app/src/theme/community/windoge98.ts
```diff
@@ -31,6 +31,10 @@ export function getTheme(base: Theme): Theme {
     base.accent = magenta;
     base.disabledTxt = txt40;
     base.placeholder = txt60;
+    // base.font =
```
hpeebles, line 34: Is this purposely left in?

julianjelfs: Yes - it's there as a suggestion but I don't want to impose too much on someone else's theme.

hpeebles (approved):

PR #4897, opened 2023-11-30
Add support for P2P trades

hpeebles:


backend/canisters/community/impl/src/lib.rs
```diff
@@ -295,6 +298,10 @@ struct Data {
     pub total_payment_receipts: PaymentReceipts,
 }
 
+fn escrow_canister_id() -> CanisterId {
+    CanisterId::from_text("s4yi7-yiaaa-aaaar-qacpq-cai").unwrap()
```
megrogan, line 302: We don't need a separate canister for prod_test?

hpeebles: Yeah we do, I'll first deploy to prod test with the prod test canister Id

megrogan (approved):

megrogan (approved):

PR #4900, opened 2023-12-01
Swap tokens FE

megrogan:


frontend/openchat-client/src/openchat.ts
```diff
@@ -5139,6 +5138,14 @@ export class OpenChat extends OpenChatAgentWorker {
         );
 
         cryptoLookup.set(cryptoRecord);
+
+        window.setTimeout(this.refreshBalancesInSeries, 0);
+    }
+
+    private async refreshBalancesInSeries() {
```
julianjelfs, line 5145: This raised a bit of an eyebrow. Any particular reason that we want / need to do them one at a time?

julianjelfs (approved):

frontend/app/src/i18n/en.json
```diff
@@ -641,6 +643,19 @@
...
     },
+    "tokenSwap": {
+        "title": "Swap",
+        "quote": "Quote",
+        "swapNotAvailable": "Currently it is not possible to swap {tokenIn} to any other token",
+        "noQuotes": "No DEXes provided a quote to swap {tokenIn}",
+        "getTokenSwapsError": "Error getting swaps for {tokenIn}",
+        "quoteError": "Error getting quotes to swap {tokenIn}",
+        "findingAvailableTokens": "finding available tokens to swap...",
+        "quoteInfo": "When you press the `{quote}` button the following decentralized exchanges (DEXes) will be queried on your behalf for a quote to swap your tokens:",
```
hpeebles, line 654: This got removed, right?

frontend/openchat-agent/src/services/dexes/index.ts
```diff
@@ -23,6 +23,23 @@ export class DexesAgent {
         );
     }
 
+    async canSwap(tokens: Set<string>): Promise<Set<string>> {
+        const allPools = await this._icpSwapIndexClient.getPools();
+
+        const available = new Set<string>();
+
+        for (const p of allPools) {
+            if (tokens.has(p.token0)) {
```
hpeebles, line 32: I think this should be `if (tokens.has(p.token0) && tokens.has(p.token1)) { // add both }`

hpeebles: Since otherwise a token may only be swappable with another token which we don't support.

frontend/openchat-client/src/openchat.ts
megrogan: Yes, so as not to block the browser connection pool with low priority calls

hpeebles (approved):

PR #4922, opened 2023-12-04
Skip refreshing balances when in anonymous mode

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -5155,12 +5162,14 @@ export class OpenChat extends OpenChatAgentWorker {
 
         cryptoLookup.set(cryptoRecord);
 
-        window.setTimeout(this.refreshBalancesInSeries, 0);
+        if (!this._liveState.anonUser) {
+            window.setTimeout(() => this.refreshBalancesInSeries(), 0);
```
hpeebles, line 5166: I had to add the `() => ...` otherwise `this` wasn't the expected value.

megrogan (approved):

PR #4926, opened 2023-12-05
Refactor browse channels

### LINKED ISSUE #4892: Refactor "Browse Channels"
It's too easy to miss new channels at the moment. Convert the "Browse Channels" button into a collapsible card that contains all other (public) channels that the user is not a member of already. This card should be open by default, but we remember the setting if the user specifically chooses to close it. The search function in the left panel should apply to both channels I am in and the ones I am not. When a channel I am not in is clicked on we enter preview mode for that channel and it should no longer show up in the list of other channels until or unless I dismiss it from preview.

julianjelfs:
fixes #4892


frontend/app/src/i18n/en.json
```diff
@@ -1068,7 +1068,7 @@
         "convert": "Convert to community",
         "convertButton": "Convert",
         "channels": "Channels",
-        "otherChannels": "Other channels",
+        "otherChannels": "Other public channels",
```
hpeebles, line 1071: Other languages need updating

PR #4970, opened 2023-12-08
Fix ordering of crypto tokens

hpeebles:


frontend/app/src/components/home/CryptoSelector.svelte
```diff
@@ -16,12 +16,13 @@
     let selecting = false;
     let ignoreClick = false;
 
-    $: cryptoLookup = client.enhancedCryptoLookup;
-    $: crypto = Object.values($cryptoLookup).filter((t) => filter(t));
+    $: cryptoLookup = client.cryptoLookup;
```
hpeebles, line 19: This is only used to look up the symbol

PR #4984, opened 2023-12-11
Added support for message filters in registry

megrogan:


backend/canisters/registry/impl/src/model/message_filters.rs
```diff
@@ -0,0 +1,72 @@
...
+            self.last_updated = now;
+            Some(id)
+        }
+    }
+
+    pub fn remove(&mut self, index: u64, now: TimestampMillis) -> bool {
+        let index = index as usize;
+        if index < self.filters.len() {
+            let filter = &mut self.filters[index];
+            filter.regex = None;
```
hpeebles, line 37: Seems a bit strange to remove the text but leave the Id.
Maybe use a map to store them and hold the latest index separately?
Or set a flag on the entry to mark it removed?

hpeebles: We could have `added: Timestamp, removed: Option<Timestamp>`

backend/canisters/registry/impl/src/model/message_filters.rs
```diff
@@ -0,0 +1,72 @@
...
+    last_updated: TimestampMillis,
+}
+
+impl MessageFilters {
+    pub fn add(&mut self, regex: String, now: TimestampMillis) -> Option<u64> {
+        let regex = Some(regex);
+        if self.filters.iter().any(|f| f.regex == regex) {
+            None
+        } else {
+            let id = (self.filters.len() - 1) as u64;
```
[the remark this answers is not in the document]
megrogan: Duh. This was in the right place and then I subsequently moved the declaration above the push!

PR #4987, opened 2023-12-11
Filter messages based on regular expressions

megrogan:
Not actually applying the filters yet

### DIFF
--- a/frontend/openchat-client/src/openchat.ts
+++ b/frontend/openchat-client/src/openchat.ts
@@ -439,6 +439,7 @@ import { isDisplayNameValid, isUsernameValid } from "./utils/validation";
 import type { DraftMessage } from "./stores/draftMessageFactory";
 import { verifyCredential } from "./utils/credentials";
 import { offlineStore } from "./stores/network";
+import { messageFilters } from "./stores/messageFilters";
 
 const UPGRADE_POLL_INTERVAL = 1000;
 const MARK_ONLINE_INTERVAL = 61 * 1000;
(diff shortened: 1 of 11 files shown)


hpeebles (dismissed):

frontend/openchat-client/src/openchat.ts
```diff
@@ -5180,23 +5196,35 @@ export class OpenChat extends OpenChatAgentWorker {
     }
 
     private async updateRegistry() {
-        const registry = await this.sendRequest({
+        const [registry, updated] = await this.sendRequest({
             kind: "updateRegistry",
         });
 
-        const cryptoRecord = toRecord(registry.tokenDetails, (t) => t.ledger);
+        if (updated) {
```
hpeebles, line 5203: I think this will fail on first load, because the stores won't have been set, but the registry won't have had any updates.

hpeebles: Maybe pass in an `initialLoad` flag or something like that

frontend/openchat-client/src/utils/chat.ts
```diff
@@ -1350,9 +1380,9 @@ function mergeLocalUpdates(
     }
 
     if (
-        senderBlocked &&
         localUpdates?.blockedMessageRevealed !== true &&
```
hpeebles, line 1383: `blockedMessageRevealed` -> `hiddenMessageRevealed`

hpeebles (approved):

PR #5007, opened 2023-12-14
Pass any failing filter to "send message" calls

megrogan:


frontend/openchat-client/src/openchat.ts
```diff
@@ -3282,6 +3286,8 @@ export class OpenChat extends OpenChatAgentWorker {
 
         const canRetry = this.canRetryMessage(msg.content);
 
+        const messageFilterFailed = doesMessageFailFilter(msg, get(messageFiltersStore), this._liveState.user.userId);
```
hpeebles, line 3289: This will always return false because the sender is the current user

hpeebles: I think the sender check should be done separately

hpeebles (approved):

PR #5018, opened 2023-12-15
Reset scroll position when changing scope

hpeebles:


frontend/app/src/components/home/ChatList.svelte
```diff
@@ -181,31 +161,19 @@
 
     let chatListElement: HTMLElement;
 
-    onMount(() => {
-        onScopeChanged();
+    beforeUpdate(() => {
+        if (previousScope === $chatListScope && previousView !== view && previousView === "chats") {
```
hpeebles, line 165: This probably looks a bit strange, but I didn't want to just check `view === "threads"` in case we ever added more views.
I want this to trigger whenever the view has changed and the previous view was `"chats"`

megrogan (dismissed):

frontend/app/src/components/home/ChatList.svelte
megrogan: if (previousScope === $chatListScope && view !== "chats" && previousView === "chats")

megrogan (approved):

PR #5035, opened 2023-12-18
Get USD & ICP token exchange rates from icpcoins

megrogan:


canister_ids.json
```diff
@@ -16,6 +16,10 @@
     "ic": "4ijyc-kiaaa-aaaaf-aaaja-cai",
     "ic_test": "7kifq-3yaaa-aaaaf-ab2cq-cai"
   },
+  "icpcoins": {
```
hpeebles, line 19: Surely just hardcode it in the client?

hpeebles (dismissed):

frontend/openchat-shared/src/domain/crypto.ts
```diff
@@ -130,3 +103,8 @@ export type AccountTransactions = {
 };
 
 export type AccountTransactionResult = Failure | (Success & AccountTransactions);
+
+export type TokenExchangeRates = {
+    to_icp: number | undefined;
```
julianjelfs, line 108: Slightly eccentric casing for js-land

megrogan: To much going back and forth between languages!

julianjelfs (approved):

PR #5045, opened 2023-12-19
don't render the menu unless we are in the viewport

julianjelfs:
I would like to understand a bit better _why_ this makes such a difference, but it's definitely worth doing.


frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -422,234 +431,240 @@
...
-                        on:initiateThread
-                        on:deleteFailedMessage
-                        on:replyPrivately={replyPrivately}
-                        on:editMessage={editMessage}
-                        on:tipMessage={tipMessage}
-                        on:reportMessage={() => (showReport = true)}
-                        on:cancelReminder={cancelReminder}
-                        on:remindMe={() => (showRemindMe = true)} />
+                    {#if showChatMenu}
+                        {#if true}
```
hpeebles, line 576: This can go

hpeebles: Ahh is this meant to be `#if intersecting`?

julianjelfs: lol -yes! literally the whole point. I was just experimenting with something else. Doh.

hpeebles (approved):

PR #5049, opened 2023-12-19
In wallet toggle tokens, usd, icp + show totals

megrogan:
You can see this on webtest


hpeebles (approved):

frontend/app/src/components/MultiToggle.svelte
```diff
@@ -0,0 +1,63 @@
...
+            {label}
+        </div>
+    {/each}
+</div>
+
+<style lang="scss">
+    .toggle-wrapper {
+        display: flex;
+        align-items: center;
+        border: 1px solid var(--bd);
```
julianjelfs, line 38: better to use `var(--bw)` rather than `1px` so that other themes look right.

frontend/app/src/components/MultiToggle.svelte
```diff
@@ -0,0 +1,63 @@
...
+        border-radius: 4px;
```
julianjelfs, line 39: as above - better to use `var(--rd)` for all the radius settings in here.

frontend/app/src/components/MultiToggle.svelte
```diff
@@ -0,0 +1,63 @@
...
+        cursor: pointer;
+
+        .toggle {
+            border: none;
+            border-right: 1px solid var(--bd);
+            padding: 6px 10px 5px 9px;
```
julianjelfs, line 45: that's some eccentric padding?

frontend/app/src/components/MultiToggle.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+            &:hover,
```
julianjelfs, line 47: button has a different background colour variable for hover so maybe we should use the same here?

frontend/app/src/components/MultiToggle.svelte
megrogan: Ah yes - left and riught should be the same!

frontend/app/src/components/home/profile/Accounts.svelte
```diff
@@ -198,21 +235,40 @@
...
+
+            &.total {
+                @include font(bold, normal, fs-100, 22);
+                padding-right: 30px;
+                text-align: right;
+            }
+        }
+
+        tr.total {
+            color: var(--disabledTxt);
```
julianjelfs, line 266: I'd say `var(--txt-light)` is a better. They are probably the same, but seems safer for theme-ing.

PR #5059, opened 2023-12-20
Re-jig of menu positioning

julianjelfs:
This is mostly based on this lib: https://github.com/simonwep/nanopop

But I had to tweak it to allow it to force a selection in the case of a complete miss while truncating the height / width. 

So in the case of the crypto selector for example it will now try to position the menu, find that there is no variant that fits and then recurse with a force flag to just take the preferred alignment and _make_ it fit.


frontend/app/src/components/MenuIcon.svelte
```diff
@@ -57,7 +56,8 @@
     }
 
     .menu {
-        position: absolute;
+        // position: absolute;
```
megrogan, line 59: Does this need to hang around?

frontend/app/src/components/home/BigDebugMenu.svelte
```diff
@@ -0,0 +1,82 @@
+<script lang="ts">
```
megrogan, line 1: Does this need to hang around?

frontend/app/src/stores/menu.ts
```diff
@@ -31,41 +26,40 @@ function close(menu: HTMLElement | undefined): HTMLElement | undefined {
...
-                menu.style.setProperty("--override-height", `${dim.h}px`);
+            if (centered && get(mobileWidth)) {
+                positionInCenter(menu);
+            } else {
+                const pos = reposition(triggerEl, menu, {
+                    position: `${position}-${align}`,
+                    margin: gutter,
+                });
+                if (!pos) {
+                    reposition(triggerEl, menu, {
```
megrogan, line 56: Couldn't this retry happen within reposition itself? Then also shared by tooltip.

megrogan (dismissed):

frontend/app/src/components/home/BigDebugMenu.svelte
julianjelfs: I think it's quite useful for testing so I wouldn't mind leaving it. Shouldn't do any harm.

frontend/app/src/components/MenuIcon.svelte
julianjelfs: nope. gone.

frontend/app/src/stores/menu.ts
julianjelfs: yes I guess so. I will change it.

megrogan (approved):

PR #5175, opened 2024-01-12
Use "swap" instead of "trade" in vars and types

megrogan:


backend/libraries/types/src/group_roles.rs
```diff
@@ -75,7 +75,7 @@ pub struct OptionalMessagePermissions {
     pub crypto: OptionUpdate<GroupPermissionRole>,
     pub giphy: OptionUpdate<GroupPermissionRole>,
     pub prize: OptionUpdate<GroupPermissionRole>,
-    pub p2p_trade: OptionUpdate<GroupPermissionRole>,
+    pub p2p_swap: OptionUpdate<GroupPermissionRole>,
```
hpeebles, line 78: We'll have to be careful with the upgrade order here

hpeebles (approved):

hpeebles (approved):

PR #5220, opened 2024-01-19
Implement ability to update user principals

hpeebles:


backend/canisters/local_user_index/impl/src/lib.rs
```diff
@@ -215,6 +218,10 @@ struct Data {
     pub rng_seed: [u8; 32],
 }
 
+fn identity_canister_id() -> CanisterId {
+    CanisterId::from_text("6klfq-niaaa-aaaar-qadbq-cai").unwrap()
```
megrogan, line 222: Use test flag to set this to prod_test id

hpeebles: The test flag isn't easily available at this stage.
I can just upgrade it first on prod test with the value set to the prod test canisterId.

backend/canisters/user_index/impl/src/lib.rs
```diff
@@ -231,6 +240,10 @@ struct Data {
     pub diamond_membership_fees: DiamondMembershipFees,
 }
 
+fn identity_canister_id() -> CanisterId {
+    CanisterId::from_text("6klfq-niaaa-aaaar-qadbq-cai").unwrap()
```
megrogan, line 244: Init prod_test too

backend/canisters/user_index/impl/src/updates/c2c_update_user_principal.rs
```diff
@@ -0,0 +1,81 @@
...
+    new_principal: Principal,
+}
+
+fn prepare(args: Args, state: &RuntimeState) -> PrepareResult {
+    let user_id = state.data.users.get_by_principal(&args.old_principal).unwrap().user_id;
+
+    PrepareResult {
+        user_id,
+        old_principal: args.old_principal,
+        new_principal: args.new_principal,
```
megrogan, line 46: Seems a bit weird to echo back the args

megrogan (approved):

megrogan (approved):

PR #5223, opened 2024-01-19
Fix input amount display in p2p swaps

hpeebles:


megrogan:
CHANGELOG entries needed for groups and communities too

hpeebles:
> CHANGELOG entries needed for groups and communities too

Change is only in User canisters.

PR #5264, opened 2024-01-25
Sync user principals to the Identity canister

hpeebles:


backend/canisters/identity/impl/src/lib.rs
```diff
@@ -53,6 +60,8 @@ struct Data {
     pub governance_principals: HashSet<Principal>,
     pub user_index_canister_id: CanisterId,
     pub cycles_dispenser_canister_id: CanisterId,
+    pub user_principals: UserPrincipals,
+    pub legacy_principals: HashSet<Principal>,
```
megrogan, line 64: Maybe legacy_principals should be part of UserPrincipals?

backend/canisters/user_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -24,4 +24,15 @@ fn post_upgrade(args: Args) {
     init_state(env, data, args.wasm_version);
 
     info!(version = %args.wasm_version, "Post-upgrade complete");
+
+    mutate_state(|state| {
```
megrogan, line 28: Perhaps the canister should remember if it has synced already so it doesn't try again (if we forget to remove the code)

And perhaps we should label the code to be removed

PR #5273, opened 2024-01-26
Improve layout of p2pSwap messages

megrogan:


frontend/app/src/components/home/P2PSwapContent.svelte
```diff
@@ -208,14 +208,14 @@
         {/if}
         <div class="coins">
             <div class="coin">
-                <SpinningToken logo={fromDetails.logo} spin={false} />
+                <SpinningToken logo={fromDetails.logo} spin={false} size="medium" />
                 <div class="amount">{fromAmount} {content.token0.symbol}</div>
             </div>
 
-            <div class="swap-icon"><SwapIcon size={"2.5em"} /></div>
+            <div class="swap-icon"><SwapIcon size={"2.5rem"} /></div>
```
julianjelfs, line 215: I think there is a reason that we do not use `rem` for the svg icons. IIRC safari doesn't allow it.

megrogan: I am still using 2.5rem for the top calculation because that is based on half the coin height (5rem)

julianjelfs (approved):

PR #5297, opened 2024-01-30
Consolidate `DirectChatEvent` and `GroupChatEvent` types

hpeebles:


frontend/openchat-shared/src/domain/chat/chat.ts
```diff
@@ -661,10 +667,9 @@ export type LocalMessageUpdates = {
 
 export type EventsResponse<T extends ChatEvent> = "events_failed" | EventsSuccessResult<T>;
```
julianjelfs, line 668: Do `EventsResponse` and `EventsSuccessResult` even _need_ to be generic any more?

julianjelfs (approved):

PR #5319, opened 2024-02-02
Send periodic OC bot messages to translators

megrogan:


backend/canisters/translations/impl/src/model/translations.rs
```diff
@@ -162,6 +168,47 @@ impl Translations {
...
+    pub fn collate_decision_summaries(&self, since: TimestampMillis) -> HashMap<UserId, DecisionSummary> {
+        self.translations
+            .iter()
+            .filter(|t| match &t.status {
+                TranslationStatus::Approved(s) => s.attribution.when > since,
+                TranslationStatus::Rejected(s) => s.attribution.when > since,
+                TranslationStatus::Deployed(s) => s.deployed > since,
+                _ => false,
+            })
+            .group_by(|t| t.proposed.who)
```
hpeebles, line 180: This will only group consecutive values, so probably isn't what you want.

hpeebles: I've just seen this which may work here - https://docs.rs/itertools/0.8.2/itertools/trait.Itertools.html#method.into_group_map

megrogan: Ah thanks. Missed that

backend/canisters/translations/impl/src/model/translations.rs
```diff
@@ -58,37 +62,43 @@ impl Translations {
     }
 
     pub fn approve(&mut self, id: u64, user_id: UserId, now: TimestampMillis) -> ApproveResponse {
-        if let Some(translation) = self.translations.get_mut(id as usize) {
+        let translation = if let Some(translation) = self.translations.get(id as usize) {
```
hpeebles, line 65: You could use the relatively new `let-else` pattern here.

`let Some(translation) = self.translations.get(id as usize) else { return ApproveResponse::NotFound };`

PR #5328, opened 2024-02-06
Implement ability for Identity canister to create key delegations for users

hpeebles:


backend/canisters/identity/api/src/queries/mod.rs
```diff
@@ -1 +1,2 @@
 pub mod check_auth_principal;
+pub mod get_delegation;
```
megrogan, line 2: We've tended to avoid using "get" as a prefix for query methods

backend/canisters/identity/impl/src/updates/prepare_delegation.rs
```diff
@@ -0,0 +1,41 @@
+use crate::{delegation_signature_msg_hash, mutate_state, Delegation, RuntimeState};
+use canister_tracing_macros::trace;
+use ic_cdk_macros::update;
+use identity_canister::prepare_delegation::{Response::*, *};
+use types::Milliseconds;
+use utils::time::{DAY_IN_MS, NANOS_PER_MILLISECOND};
+
+const DEFAULT_EXPIRATION_PERIOD: Milliseconds = 30 * DAY_IN_MS;
+const MAX_EXPIRATION_PERIOD: Milliseconds = 90 * DAY_IN_MS;
```
megrogan, line 9: I suppose we can think about this in a future iteration

megrogan (approved):

PR #5331, opened 2024-02-07
Tool to merge translation corrections

megrogan:


scripts/merge-latest-translations-prod.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
julianjelfs, line 1: it seems a bit dubious that any version of this _other than_ prod exists.  We would never want to update the files with corrections from anywhere other than prod (would we)?

hpeebles (approved):

PR #5337, opened 2024-02-08
Push anonymised `user_online` events to the `event_sink` canister

hpeebles:


backend/canisters/event_relay/impl/src/lib.rs
```diff
@@ -46,20 +55,44 @@ impl RuntimeState {
...
             cycles_dispenser_canister_id,
+            salt: [0; 32],
             rng_seed: [0; 32],
             test_mode,
         }
     }
+
+    pub fn obfuscate_user(&self, user: String) -> String {
+        // We only want to obfuscate userId principals, so if the string is not a principal we return it as is
+        if Principal::from_text(&user).is_err() {
```
megrogan, line 86: Possibly an unnecessary restriction

hpeebles: Yeah I was just thinking if we ever say user = ProposalsBot or something like that

hpeebles: I'll remove the restriction for now since we may never need it

megrogan (approved):

PR #5415, opened 2024-02-21
Fix `access_token` endpoint + integration test

megrogan:


julianjelfs (approved):

julianjelfs (dismissed):

backend/canisters/local_user_index/impl/src/queries/access_token.rs
```diff
@@ -48,11 +52,22 @@ fn get_user(state: &RuntimeState) -> Option<(UserId, bool)> {
     })
 }
 
-fn build_token(user_id: UserId, chat: Chat, start_call: bool, state: &RuntimeState) -> Response {
+fn build_token(user_id: UserId, args: Args, state: &RuntimeState) -> Response {
     if let Some(secret_key_der) = state.data.oc_secret_key_der.as_ref() {
-        let claims = VideoCallClaims::new(user_id, chat, start_call, state.env.now());
+        let now = state.env.now();
+
+        let mut rng = seed_rng(&state.data.rng_seed, state.env.caller(), now);
```
hpeebles, line 59: Use `state.env.rng` here instead. `rng_seed` is only used across upgrades so that we don't have to reseed the rng every time.

megrogan: yes whoops

hpeebles (dismissed):

PR #5445, opened 2024-02-27
Video call support

julianjelfs:
This is constrained to only be available to platform operators for the time being so that we can get things merged more regularly and try it out between us.


frontend/app/src/components/home/ChatList.svelte
```diff
@@ -200,6 +201,7 @@
     }
 </script>
 
+<!-- svelte-ignore missing-declaration -->
```
hpeebles, line 204: Is this needed? Doesn't seem like it would be needed based on the additions...

hpeebles (dismissed):

hpeebles (approved):

PR #5449, opened 2024-02-28
More testing around access_token functionality

megrogan:


backend/integration_tests/src/env.rs
```diff
@@ -15,11 +16,19 @@ pub struct TestEnvManager {
...
 impl TestEnvManager {
     pub fn get(&self) -> TestEnvWrapper {
+        self.get_internal(None)
+    }
+
+    pub fn get_with_seed(&self, seed: Hash) -> TestEnvWrapper {
+        self.get_internal(Some(seed))
+    }
+
+    fn get_internal(&self, seed: Option<Hash>) -> TestEnvWrapper {
```
hpeebles, line 26: This isn't going to set the seed in the case where it reuses an existing env.

hpeebles (approved):

PR #5451, opened 2024-02-28
fix group webrtc connections

julianjelfs:


hpeebles (approved):

frontend/openchat-client/src/utils/chat.ts
```diff
@@ -136,7 +136,8 @@ export function getUsersToMakeRtcConnectionsWith(
     }
 
     const activeUsers = getRecentlyActiveUsers(chat, events, MAX_RTC_CONNECTIONS_PER_CHAT);
-    return activeUsers.has(myUserId) ? Array.from(activeUsers).filter((u) => u !== myUserId) : [];
+    activeUsers.delete(myUserId);
```
hpeebles, line 139: Actually I think this is right. You only make connections if you are a recently active user. Otherwise everyone viewing a chat will attempt to make webrtc connections to the same few users.

julianjelfs: But that seems flawed. If I visit a chat that people are actively using I'm not going to get any webrtc messages. Why is that good?

hpeebles: If you're not an active participant then it doesn't matter if you get messages 5 seconds later

hpeebles: But maybe we can improve the recently active members check. Especially for small groups.

PR #5455, opened 2024-02-29
Put back code to sync secret key with LocalUserIndexes

megrogan:


hpeebles (dismissed):

backend/canisters/user_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -36,5 +37,11 @@ fn post_upgrade(args: Args) {
 
             crate::jobs::sync_legacy_user_principals::start_job_if_required(state);
         }
+
```
hpeebles, line 40: We should wrap this in an `if data.oc_key_pair.is_initialised()`

hpeebles (approved):

PR #5472, opened 2024-03-04
Reject NNS proposal just before deadline if neuron not already voted

hpeebles:


backend/canisters/proposals_bot/impl/src/jobs/retrieve_proposals.rs
```diff
@@ -110,16 +119,41 @@ fn handle_proposals_response<R: RawProposal>(governance_canister_id: CanisterId,
...
-
             mutate_state(|state| {
+                // TODO Remove this!
+                // Temp hack for Dragginz
+                // Dfinity are fixing a bug in their governance canister which is causing it to
+                // return old proposals
+                let dragginz_governance_canister_id: CanisterId = CanisterId::from_text("zqfso-syaaa-aaaaq-aaafq-cai").unwrap();
+                if governance_canister_id == dragginz_governance_canister_id {
+                    proposals.retain(|p| p.id() > 36)
+                } else if governance_canister_id == state.data.nns_governance_canister_id {
```
megrogan, line 130: Maybe have this as the first if clause then put the TODO comment before the else if so it is clear which block is referred to by the comment

megrogan (approved):

megrogan (approved):

PR #5573, opened 2024-03-19
Prevent bot users from being able to get video call access tokens

hpeebles:


backend/canisters/community/api/src/queries/c2c_can_issue_access_token_for_channel.rs
```diff
@@ -5,6 +5,8 @@ use types::{AccessTokenType, ChannelId, UserId};
 pub struct Args {
     pub user_id: UserId,
     pub is_diamond: bool,
+    #[serde(default)]
+    pub is_bot: bool,
```
megrogan, line 9: Surely just check is_bot in `access_token` directly

megrogan (approved):

PR #5621, opened 2024-04-02
Top up all canisters registered with the SNS

hpeebles:


backend/external_canisters/sns_root/api/src/queries/get_sns_canisters_summary.rs
```diff
@@ -10,6 +10,7 @@ pub struct Response {
     pub governance: Option<CanisterSummary>,
     pub ledger: Option<CanisterSummary>,
     pub swap: Option<CanisterSummary>,
+    pub dapps: Vec<CanisterSummary>,
```
megrogan, line 13: Is there a better name I wonder?

hpeebles: That's the name coming from the SNS root canister

megrogan (approved):

PR #5629, opened 2024-04-03
Allow platform operators to enable/disable the principal migration job

hpeebles:


backend/canisters/identity/api/src/main.rs
```diff
@@ -8,6 +8,7 @@ fn main() {
     generate_candid_method!(identity, generate_challenge, update);
     generate_candid_method!(identity, migrate_legacy_principal, update);
     generate_candid_method!(identity, prepare_delegation, update);
+    generate_candid_method!(identity, set_principal_migration_job_enabled, update);
```
megrogan, line 11: I wondering why you haven't implemented the function? It seems like it would be very quick.

hpeebles: It's already implemented. Just wasn't wired up on the frontend

megrogan (approved):

PR #5648, opened 2024-04-08
Select chat modal should not re-order chats

megrogan:


hpeebles (dismissed):

frontend/app/src/components/SelectChatModal.svelte
```diff
@@ -74,8 +74,10 @@
     $: selectedChatId = client.selectedChatId;
     $: globalState = client.globalStateStore;
 
+    let initialGlobalState = structuredClone($globalState);
```
julianjelfs, line 77: would this not work just as well without the `structuredClone`?

megrogan: Yes structuredClone was wrong

hpeebles (approved):

PR #5677, opened 2024-04-15
Markdown toggle

### LINKED ISSUE #5656: Make block level markdown an opt-in feature
Block level markdown is a power-user feature and it _can_ have unintended consequences so it is probably better that it has to be specifically enabled for a message. 

This requires some UI work to enable (and probably preview a markdown message) and also a back end change to capture a flag indicating whether the message should be rendered with block level markdown or not.

julianjelfs:
The detects if the user has entered block level markdown content in a text message and if they have, will display a discrete toggle to decide whether to render with block level markdown or not (this will be remembered in local storage). 

This won't actually _work_ unless / until we add the required flag to the relevant `send_message` implementations and to the `Message` type definition. 

fixes #5656


julianjelfs:
This currently doesn't work for editing messages

megrogan (approved):

PR #5680, opened 2024-04-16
Add `block_level_markdown` flag to messages

hpeebles:


julianjelfs (dismissed):

backend/canisters/community/api/src/updates/c2c_send_message.rs
```diff
@@ -13,10 +13,16 @@ pub struct Args {
...
     pub forwarding: bool,
+    #[serde(default = "bool_true")]
+    pub block_level_markdown: bool,
     pub community_rules_accepted: Option<Version>,
     pub channel_rules_accepted: Option<Version>,
     #[serde(default)]
     pub message_filter_failed: Option<u64>,
 }
 
+fn bool_true() -> bool {
```
megrogan, line 24: We can do better than this.
1. Either all messages before the blockdown feature were added are false
2. Or we detect blockklevel markdown in the message like wot the FE does

megrogan (approved):

PR #5695, opened 2024-04-18
Trigger incoming call via cache primer

julianjelfs:


frontend/openchat-client/src/openchat.ts
```diff
@@ -3112,7 +3111,7 @@ export class OpenChat extends OpenChatAgentWorker {
                         threadRootMessageIndex,
                     });
                 }
-                if (unconfirmed.delete(context, messageId)) {
+                if (unconfirmed.contains(context, messageId)) {
```
hpeebles, line 3114: This has reverted my revert

julianjelfs: Well spotted - not sure how that happened.

hpeebles (approved):

hpeebles:
Are we checking if the chat is muted anywhere? We probably shouldn't ring for chats that are muted.

julianjelfs:
> Are we checking if the chat is muted anywhere? We probably shouldn't ring for chats that are muted.

We aren't at the moment. I'm in two minds about that. It's only for private chats. And I think there's probably be a difference between not wanting notifications and not wanting to receive calls.

hpeebles (approved):

PR #5717, opened 2024-04-22
make sure that the messageId is definitely a bigint

julianjelfs:
Note: this is an educated guess at what the problem is but I'm fairly confident. We should double check on webtest.

### DIFF
--- a/frontend/openchat-client/src/events.ts
+++ b/frontend/openchat-client/src/events.ts
@@ -33,7 +33,7 @@ export class RemoteVideoCallStartedEvent extends CustomEvent<{
     messageId: bigint;
 }> {
     constructor(chatId: ChatIdentifier, userId: string, messageId: bigint) {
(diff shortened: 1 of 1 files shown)


frontend/openchat-client/src/events.ts
```diff
@@ -33,7 +33,7 @@ export class RemoteVideoCallStartedEvent extends CustomEvent<{
     messageId: bigint;
 }> {
     constructor(chatId: ChatIdentifier, userId: string, messageId: bigint) {
-        super("openchat_event", { detail: { chatId, userId, messageId } });
+        super("openchat_event", { detail: { chatId, userId, messageId: BigInt(messageId) } });
```
hpeebles, line 36: We should just do this within `rtc.ts` where we receive the message.
We convert other BigInt fields in there too.

hpeebles (approved):

PR #5726, opened 2024-04-25
Mark principals as migrated after being updated

hpeebles:


backend/canisters/user_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -24,4 +26,23 @@ fn post_upgrade(args: Args) {
     init_state(env, data, args.wasm_version);
 
     info!(version = %args.wasm_version, "Post-upgrade complete");
+
+    let already_migrated = vec![
+        "3skqk-iqaaa-aaaaf-aaa3q-cai",
+        "27eue-hyaaa-aaaaf-aaa4a-cai",
+        "2yfsq-kaaaa-aaaaf-aaa4q-cai",
+    ];
```
megrogan, line 34: Need to add Steffen now

megrogan (approved):

PR #5752, opened 2024-05-01
don't check this here - it'll fail later

julianjelfs:
It's better not to have this check here.


frontend/app/src/utils/pin.ts
```diff
@@ -0,0 +1,23 @@
+import { writable } from "svelte/store";
```
megrogan, line 1: You should delete this

megrogan (approved):

PR #5776, opened 2024-05-07
replace bio

julianjelfs:


frontend/app/src/components/landingpages/Team.svelte
```diff
@@ -26,7 +26,7 @@
             imageUrl={"/assets/team/steffen2.png"}
```
megrogan, line 26: I think the photo ios different too

julianjelfs: It is but I have strategically ignored the photo because I've already done it and it's the wrong shape.

megrogan (approved):

PR #5790, opened 2024-05-09
Keep retrying to get user's country code

hpeebles:


julianjelfs (commented):
I'm not convinced by this. We don't have much reason to think this is the problem and also this is just going to keep trying forever even when there is some fatal error. The users complaining (hysterically) have tried restarting after all.

PR #5802, opened 2024-05-14
Enable rolling storage for all users

hpeebles:


backend/canisters/storage_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -27,4 +27,10 @@ fn post_upgrade(args: Args) {
     init_state(env, data, args.wasm_version);
 
     info!(version = %args.wasm_version, "Post-upgrade complete");
+
+    mutate_state(|state| {
```
megrogan, line 31: Which users don't have this enabled currently?

hpeebles: It looks like we never enabled it for existing users. So it is set to false for users who registered before Feb 2023.

megrogan (approved):

PR #5821, opened 2024-05-20
CHIT leaderboard backend

megrogan:


hpeebles (dismissed):

backend/libraries/types/src/chit.rs
```diff
@@ -15,3 +15,9 @@ pub enum ChitEarnedReason {
     DailyClaim,
     Achievement(String),
 }
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug)]
+pub struct ChitUserBalance {
+    pub user_id: UserId,
+    pub balance: i32,
```
hpeebles, line 22: Maybe this should be a u32? And we exclude negatives.

hpeebles: Otherwise if the first user in a season gets a deduction, they'll still appear top.

megrogan:
Not ready yet... still need to remove a user if there balance has gone negative

megrogan:
Latest commit means the leaderboard could have < 50 users even though there are at last 50 postive balances but this is very unlikely in practise

hpeebles (approved):

PR #5834, opened 2024-05-21
Claim daily CHIT, streak badges, balance

megrogan:


frontend/app/src/components/home/profile/Badges.svelte
julianjelfs, line 1: It's a tiny bit unfortunate that this is _not_ quite self-contained in that (presumably) it is assumed that it is used within a flexbox container.

julianjelfs (approved):

frontend/app/src/components/home/profile/Badges.svelte
megrogan: Indeed. I tried to wrap these two components in a flexbox container to control the centering and gap but it then went wrong in the chat header for some reason.

PR #5836, opened 2024-05-22
Throttle messages

julianjelfs:
Handle message throttling a bit better: 

* increase the limits a bit
* store the timestamps in local storage so you can't circumvent by simply refreshing
* indicate that messages are being throttled in the footer and show a timer when the next message can be sent


frontend/openchat-client/src/stores/throttling.ts
```diff
@@ -0,0 +1,62 @@
+import { writable } from "svelte/store";
+
+const localStorageKey = "openchat_most_recent_sent_message_times";
+
+const countdown = writable<number>(0);
+export const throttleCountdown = {
```
megrogan, line 6: Could just store the timestamp and pass that to ThrottleCountdown and use $now500?

julianjelfs: Yes I think that would be better. This all evolved before I did the UI part, so I think you're right it can be simpler.

megrogan (approved):

PR #5840, opened 2024-05-22
CHIT leaderboard

megrogan:


frontend/app/src/i18n/en.json
```diff
@@ -425,6 +425,7 @@
     "guestUser": "Guest User",
     "halloffame": {
         "allTime": "All time",
+        "balance": "CHIT",
```
hpeebles, line 428: I know its the same in every language but we should still add the translations, otherwise they'll get added next time we run the script and they may get incorrectly translated.

megrogan: There is a bunch of CHIT related stuff not translated (using sync_lang) yet because it is really placeholder text and it will definitely need deleting from all the translation files which is a pain.

frontend/app/src/components/home/ChitHallOfFame.svelte
```diff
@@ -0,0 +1,270 @@
...
+
+    function getData() {
+        client
+            .chitLeaderboard()
+            .then((result) => {
+                leaders = result.slice(0, 10);
+            })
+            .finally(() => {
+                if (leaders.length < 10) {
+                    for (let i = leaders.length; i < 10; i++) {
```
julianjelfs, line 63: `leaders = [...leaders, ...Array(10 - leaders.length).fill(blankLeader)]`

way cooler

julianjelfs (approved):

PR #5869, opened 2024-05-30
Avoid waiting for tip to be confirmed

hpeebles:


hpeebles:
> This is just going to swallow any error though. If there is a failure, the modal will be closed and so there is nothing there to display the error. We need to use the error toast.

Right yeah I thought that may be the case. I guess this should've been a draft (or not a PR at all), as I was just throwing my untested changes over to Matt who is going to pick this up.

hpeebles:
I'll close this and Matt can create a new PR once its all ready.

PR #5896, opened 2024-06-03
Job to update users whose streak ended yesterday

megrogan:


backend/canisters/user_index/impl/src/jobs/update_user_streaks.rs
```diff
@@ -0,0 +1,42 @@
...
+const MS_IN_DAY: Milliseconds = 1000 * 60 * 60 * 24;
+
+thread_local! {
+    static TIMER_ID: Cell<Option<TimerId>> = Cell::default();
+}
+
+pub(crate) fn start_job_if_required(state: &RuntimeState) -> bool {
+    if TIMER_ID.get().is_none() {
+        let now = state.env.now();
+        let run_at = time_until_midnight(now);
```
hpeebles, line 17: `run_at` is the delay, so the name isn't correct.

backend/canisters/user_index/impl/src/model/streak.rs
```diff
@@ -45,6 +45,14 @@ impl Streak {
         }
     }
 
+    pub fn expired_yesterday(&self, now: TimestampMillis) -> bool {
+        if let Some(today) = Streak::timestamp_to_day(now) {
```
hpeebles, line 49: This will result in loads of unnecessary calls to `timestamp_to_day`.
We could just pass in the day index.

hpeebles: Maybe just `expired`? rather than having to be a specific day.

hpeebles (approved):

hpeebles (approved):

PR #5917, opened 2024-06-07
Enable Sonic for swapping

megrogan:


frontend/openchat-agent/src/services/dexes/sonic/swaps/sonic.swaps.client.ts
```diff
@@ -8,7 +8,7 @@ import { Principal } from "@dfinity/principal";
 
 const SONIC_INDEX_CANISTER_ID = "3xwpq-ziaaa-aaaah-qcn4a-cai";
 const TEN_MINUTES = 10 * 60 * 1000;
-const ENABLED: boolean = false;
+const ENABLED: boolean = true;
```
hpeebles, line 11: Just remove this flag

megrogan: Yes but maybe if there was an issue it would be convenient to rollback

hpeebles (approved):

PR #5940, opened 2024-06-14
fix chat state caching problem

julianjelfs:
+ a couple of changes to experiment with improving the output from the rollbar logging.

### DIFF
--- a/frontend/openchat-agent/src/utils/caching.ts
+++ b/frontend/openchat-agent/src/utils/caching.ts
@@ -48,7 +48,7 @@ import type { CryptocurrencyContent } from "openchat-shared";
 import type { PrizeContent } from "openchat-shared";
 import type { P2PSwapContent } from "openchat-shared";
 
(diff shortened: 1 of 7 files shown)


frontend/openchat-agent/src/utils/caching.ts
```diff
@@ -48,7 +48,7 @@ import type { CryptocurrencyContent } from "openchat-shared";
 import type { PrizeContent } from "openchat-shared";
 import type { P2PSwapContent } from "openchat-shared";
 
-const CACHE_VERSION = 103;
+const CACHE_VERSION = 104;
```
megrogan, line 51: So there was no way to fix this more surgically?

julianjelfs: I think I'd rather tackle that when there is no immediate urgency because it's something that we _really_ don't want to fuck up.

megrogan (approved):

PR #5941, opened 2024-06-14
Add a CAPTCHA to register with ETH and SOL wallets

megrogan:


frontend/app/src/components/ModalContent.svelte
```diff
@@ -28,14 +28,22 @@
...
 
     // if your modal *definitely* overflows on mobile you might need to set height explicitly
     export let overflows: boolean = false;
 
     let divElement: HTMLElement;
+    let style = "";
 
     $: useAlignTo = alignTo !== undefined && !$mobileWidth;
-    $: style = useAlignTo ? "visibility: hidden;" : "visibility: visible;";
+    $: {
```
julianjelfs, line 40: this seems quite fishy. I would have thought that it's actually _not_ the modal that is fixed width, it is (probably always) the modal's content. So you're actually saying, the captcha should be 400px wide and the modal should just be as wide as it needs to be to show its contents.

frontend/app/src/components/home/ChallengeModal.svelte
```diff
@@ -0,0 +1,127 @@
...
+    }
+    .body {
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
+        align-items: center;
+        gap: $sp6;
+
+        img {
+            width: 100%;
```
julianjelfs, line 121: As per previous comment feels like this should just be: 

```
img {
    width: 400px;
    @include mobile() {
        width: 100%;
      }
   }
   ```

File: frontend/openchat-client/src/openchat.ts [hunk 3]
```diff
@@ -755,27 +781,38 @@ export class OpenChat extends OpenChatAgentWorker {
...
+    async submitChallenge(challengeAttempt: ChallengeAttempt): Promise<boolean> {
+        if (this._authPrincipal === undefined) {
+            return false;
+        }
 
-        if (this._authPrincipal !== undefined) {
-            this._ocIdentity = await this._ocIdentityStorage.get(this._authPrincipal);
-        } else {
-            await this._ocIdentityStorage.remove();
+        const resp = await this.sendRequest({
```
julianjelfs, line 789: should have a catch on this so that there is no chance of unhandled exception

frontend/openchat-client/src/openchat.ts
```diff
@@ -930,6 +967,14 @@ export class OpenChat extends OpenChatAgentWorker {
         return this._liveState.userCreated && identity !== null;
     }
 
+    generateIdentityChallenge(): Promise<GenerateChallengeResponse> {
+        return this.sendRequest({
```
julianjelfs, line 971: should have a catch on this to avoid chance of unhandled exception

julianjelfs (commented):
There's a fair bit of this around connecting to and initialising the worker that I'm not sure I really understand.

julianjelfs (approved):

PR #5945, opened 2024-06-18
Mitigate performance issues for large communities

julianjelfs:
When a channel in a large community is selected on a cold cache there are problems caused by trying to load thousands of users in one go. In fact this operation is so slow that it can end up being performed twice because it has not finished by the time it is requested again. 

To mitigate this, we extra only the members with elevated roles and up to 1000 other member to add to the userstore. This ensures that we load the key people and enough other users to make the members panel look populated. 

When the user opens the members panel of a very large group they will almost certainly search for the user they are interested in anyway. 

With this update, when they search for a user, if the list of resolved users is smaller than the list of members we will also call the global user search at the same time to fill in the cache with any members that we are missing. 

This is not perfect as there _could_ be more members of the group that match the search term than the max results of the search. This is quite unlikely to be a serious issue as the user can always refine their search. 

This is not a permanent solution as we are still loading _all_ of the members of a community / group each time it is selected. Though mysteriously this seems to be much less of a problem than loading all of the users. 

Then end result: it is still kind of sluggish when loading on an empty cache, but it should at least not crash.


frontend/openchat-shared/src/utils/logging.ts
```diff
@@ -55,3 +55,7 @@ export function debug<T>(data: T, msg?: string): T {
     }
     return data;
 }
+
+export function logDuration(msg: string, started: number): void {
```
megrogan, line 59: We're not referencing this function in the PR. I suppose it doesn't really matter - it might be prove useful later.

megrogan (approved):

frontend/openchat-shared/src/utils/logging.ts
julianjelfs: Yes I thought I would just leave it there since I have deleted it and recreated it a couple of times recently.

PR #5946, opened 2024-06-19
Retry unfinished ICPSwaps + some extra logging

megrogan:
During the NNS subnet incident on Friday 14th the ICP ledger was unavailable for some of the time. There is some evidence that if a user was swapping to ICP during this time, ICPSwap might see their balance as zero and fail to deposit or withdraw the funds.

For an example see this user's canister logs at timestamp 1718396682164:
https://2enzf-kyaaa-aaaaf-avdva-cai.raw.icp0.io/logs

Also see the transaction history of ICP transferred into their account shows no record on the Friday so the withdrawal definitely failed:
https://dashboard.internetcomputer.org/transactions?p=7&m=12439383&t=259&to=550b8e0b7f5291b84cdd8c196f90c0756a2eb66f2ed79be378e5d550ea9380b3

### DIFF
--- a/backend/canisters/user/impl/src/lifecycle/post_upgrade.rs
+++ b/backend/canisters/user/impl/src/lifecycle/post_upgrade.rs
@@ -1,6 +1,7 @@
 use crate::lifecycle::{init_env, init_state};
 use crate::memory::get_upgrades_memory;
-use crate::{mutate_state, Data};
+use crate::timer_job_types::{ProcessTokenSwapJob, TimerJob};
+use crate::{mutate_state, Data, RuntimeState};
 use canister_logger::LogEntry;
 use canister_tracing_macros::trace;
 use ic_cdk::post_upgrade;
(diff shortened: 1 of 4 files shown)


julianjelfs (approved):

backend/canisters/user/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -49,3 +53,24 @@ fn mark_user_canister_empty() {
         );
     })
 }
+
+fn retry_token_swaps(state: &mut RuntimeState) {
```
hpeebles, line 57: The swap could still be in progress. I think you can call is_complete or something like that

megrogan: Oh yes. Thanks for checking. I had in fact meant to look at the swap creation time and only do this if > 30 mins since creation.

PR #5951, opened 2024-06-19
Post release

megrogan:


backend/canisters/identity/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -24,6 +24,4 @@ fn post_upgrade(args: Args) {
     init_state(env, data, args.wasm_version);
 
     info!(version = %args.wasm_version, "Post-upgrade complete");
-
-    mutate_state(|state| state.data.user_principals.populate_originating_canisters());
```
megrogan, line 28: I presume this is a one time job @hpeebles?

hpeebles (approved):

PR #5962, opened 2024-06-25
Achievements BE part 1

megrogan:


backend/libraries/types/src/achievement.rs
```diff
@@ -0,0 +1,39 @@
...
+    Streak3,
+    Streak7,
+    Streak14,
+    Streak30,
+}
+
+impl Achievement {
+    pub fn chit_reward(&self) -> u32 {
+        match self {
+            Achievement::JoinedGroup => 500,
```
megrogan, line 24: We need to agree the CHIT rewards

julianjelfs (approved):

PR #5972, opened 2024-06-28
Move ownership of chit balance and streak to users

megrogan:


julianjelfs (approved):

backend/canisters/user/impl/src/updates/claim_daily_chit.rs
```diff
@@ -0,0 +1,94 @@
...
+    if streak >= 30 {
+        state.data.award_achievement(Achievement::Streak30, now);
+    }
+
+    state.data.notify_user_index_of_chit(now);
+
+    state.data.event_store_client.push(
+        EventBuilder::new("user_claimed_daily_chit", now)
+            .with_user(user_id.to_string(), true)
+            .with_source(user_id.to_string(), false)
```
hpeebles, line 58: This needs to be anonymised too otherwise you can immediately tie the user to the anonymized Id

backend/canisters/user_index/api/src/updates/c2c_notify_chit.rs
```diff
@@ -0,0 +1,15 @@
+use serde::{Deserialize, Serialize};
+use types::TimestampMillis;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Args {
+    pub chit_balance: i32,
```
hpeebles, line 6: I think this needs a timestamp, because they may not arrive in order, so the user index needs to be able to ignore values that aren't the latest.

hpeebles (approved):

PR #5973, opened 2024-07-01
Improve cache

### LINKED ISSUE #5966: Refine service worker caching
Currently the service worker will only cache a resource that supplies the X-Cacheable-Resource header which is an attempt to mitigate the occasional fallback html responses we get during upgrade. 

This is a bit annoying as it means everything that serves content to the front end needs to add these headers. 

We want to modify this so that only resources requested from the oc.app origin require this header. Then stuff that's not coming form the asset canister will carry on working as before.

julianjelfs:
by using a custom workbox plugin which ensures that response code is 200 (like the existing plugin) and also that the response content-type is _not_ text/html

fixes #5966

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -58,10 +58,11 @@
         "turbo": "^1.10.13"
     },
     "dependencies": {
-        "workbox-cacheable-response": "^7.0.0",
-        "workbox-expiration": "^7.0.0",
-        "workbox-recipes": "^7.0.0",
-        "workbox-routing": "^7.0.0",
-        "workbox-strategies": "^7.0.0"
+        "workbox-cacheable-response": "^7.1.0",
+        "workbox-core": "^7.1.0",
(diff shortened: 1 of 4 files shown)


frontend/package.json
```diff
@@ -58,10 +58,11 @@
         "turbo": "^1.10.13"
     },
     "dependencies": {
-        "workbox-cacheable-response": "^7.0.0",
-        "workbox-expiration": "^7.0.0",
-        "workbox-recipes": "^7.0.0",
-        "workbox-routing": "^7.0.0",
-        "workbox-strategies": "^7.0.0"
+        "workbox-cacheable-response": "^7.1.0",
```
hpeebles, line 61: Can we remove this dependency now?

hpeebles (approved):

hpeebles (approved):

PR #5982, opened 2024-07-03
Add ZKP to XFrame ancestors

puresamari:
Adding the current ZKP url (https://bxnm7-oqaaa-aaaag-albpq-cai.icp0.io/) to xframe.ts and ic-assets.json .


julianjelfs (changes requested):
I notice that you have a trailing / in both cases. Not 100% sure whether that will cause a problem but given that none of the others have that I would suggest that we remove it.

puresamari:
Heh yeah my bad, i removed them @julianjelfs

julianjelfs (approved):

PR #5984, opened 2024-07-03
Prevent previewing messages of gated chats

hpeebles:


megrogan:
Could we just make the FE changes?

megrogan (approved):

julianjelfs (approved):

PR #5988, opened 2024-07-03
Support composite access gates

hpeebles:


backend/libraries/types/src/gated_groups.rs
```diff
@@ -75,6 +78,19 @@ pub struct TokenBalanceGate {
...
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct CompositeGate {
+    pub left: Box<AccessGate>,
+    pub right: Box<AccessGate>,
+    pub and: bool,
+}
+
+impl CompositeGate {
+    pub fn synchronous(&self) -> bool {
+        self.left.synchronous() && self.right.synchronous()
```
megrogan, line 90: need to do || too

backend/libraries/types/can.did
```diff
@@ -1521,6 +1521,11 @@ type AccessGate = variant {
     SnsNeuron : SnsNeuronGate;
     Payment : PaymentGate;
     TokenBalance : TokenBalanceGate;
+    Composite : record {
```
julianjelfs, line 1524: I thought we decided to opt for a list of gates with a bool to say whether it was any or all in the end?

backend/libraries/gated_groups/src/lib.rs
```diff
@@ -136,6 +145,21 @@ fn check_verified_credential_gate(
     CheckIfPassesGateResult::Success
 }
 
+fn check_composite_gate_synchronously(gate: CompositeGate, args: CheckGateArgs) -> CheckIfPassesGateResult {
+    let count = gate.inner.len();
+    for (index, inner) in gate.inner.into_iter().enumerate() {
+        let result = check_if_passes_gate_synchronously(inner, args.clone());
```
julianjelfs, line 151: Given the types, this ends up being recursive (as a composite gate can contain composite gates) so we should probably have definite termination to that recursion (even if we don't anticipate the front end creating recursive gates) like a max depth?

hpeebles: I've added validation to disallow any composite gates within a composite gate so it shouldn't ever happen but yeah I'll see if I can protect against that in a nice way here too.

hpeebles: Updated. It'll be fairly easy to switch this to allow some nesting if we ever decide to do so.

julianjelfs (approved):

PR #6008, opened 2024-07-08
Calculate streak on demand for the current user

julianjelfs:
One thing to note is that the streak value will not _itself_ be reactive with this approach. I think this will not cause a problem in practice, but if it does we can pass in one of the $now stores.

### DIFF
--- a/frontend/openchat-agent/src/utils/caching.ts
+++ b/frontend/openchat-agent/src/utils/caching.ts
@@ -49,7 +49,7 @@ import type { CryptocurrencyContent } from "openchat-shared";
 import type { PrizeContent } from "openchat-shared";
 import type { P2PSwapContent } from "openchat-shared";
 
-const CACHE_VERSION = 105;
+const CACHE_VERSION = 106;
 const FIRST_MIGRATION = 104;
 const MAX_INDEX = 9999999999;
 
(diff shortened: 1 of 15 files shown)


frontend/openchat-agent/src/utils/caching.ts
```diff
@@ -136,6 +136,11 @@ const migrations: Record<number, MigrationFunction<ChatSchema>> = {
         const store = tx.objectStore("chats");
         store.delete(key);
     },
+    106: async (principal, tx) => {
```
hpeebles, line 139: These 2 could be deduplicated

hpeebles (approved):

frontend/openchat-agent/src/utils/caching.ts
```diff
@@ -130,12 +130,18 @@ type MigrationFunction<T> = (
     transaction: IDBPTransaction<T, StoreNames<T>[], "versionchange">,
 ) => Promise<void>;
 
+async function clearChatsStore(
```
megrogan, line 133: should be clearChitStore

julianjelfs: Nope. The chit is stored in the chat store so the migration clears the _chat_ store.

PR #6016, opened 2024-07-10
Don't ring for incoming video call if already in the call

hpeebles:


megrogan (approved):

frontend/app/src/components/home/Home.svelte
```diff
@@ -293,13 +293,14 @@
...
 
     function remoteVideoCallStarted(ev: RemoteVideoCallStartedEvent) {
-        // Check user is not already in the call and it started less than an hour ago
-        if (
-            !ev.detail.currentUserIsParticipant &&
-            Number(ev.detail.timestamp) > Date.now() - 60 * 60 * 1000
-        ) {
-            incomingVideoCall.set(ev.detail);
+        // If current user is already in the call, or has previously been in the call, or the call started more than an hour ago, exit
+        if (($selectedChatStore?.id === ev.detail.chatId && $selectedChatStore.videoCallInProgress !== undefined) ||
```
julianjelfs, line 297: What do you think that: 

`($selectedChatStore?.id === ev.detail.chatId && $selectedChatStore.videoCallInProgress !== undefined)`

means? _I_ think it means that the call is for the currently selected chat and the selected chat has a call in progress. I don't get why we would want to _not_ ring in that circumstance.

hpeebles: Ahh I thought it meant the user is in the call. I'll fix that.

PR #6034, opened 2024-07-15
Support for a bunch more achievements

megrogan:


frontend/package.json
```diff
@@ -9,7 +9,7 @@
         "build:prod_test": "npm ci && dotenv -- turbo run build:prod_test --cache-dir .turbo",
         "build:testnet": "npm ci && dotenv -- turbo run build:testnet --cache-dir .turbo",
         "deploy:prod_test": "npm i && dotenv -- turbo run deploy:prod_test --cache-dir .turbo",
-        "deploy:prod": "npm ci && dotenv -- turbo run deploy:prod --cache-dir .turbo",
+        "deploy:prod": "npm i && dotenv -- turbo run deploy:prod --cache-dir .turbo",
```
hpeebles, line 12: .

hpeebles:
I think you should run codegen and set the `new_achievement` field to false everywhere.
Because we need that to be done on the FE before any of this can go out on the BE.

hpeebles (approved):

PR #6043, opened 2024-07-16
Handle new achievements on FE

megrogan:


frontend/openchat-shared/src/domain/chit.ts
```diff
@@ -23,71 +23,74 @@ export type MemeContestWinner = {
...
-    "set_avatar",
-    "joined_group",
-    "joined_community",
-    "sent_direct_message",
-    "received_direct_message",
-    "upgraded_to_diamond",
-    "set_display_name",
-    "upgrade_to_gold_diamond",
+    "appointed_group_mod",
+    "changed_theme",
```
julianjelfs, line 80: Curious about `changed_theme` - theme is device specific and is not recorded on the back end at all. So how are we able to track that?

hpeebles (approved):

PR #6052, opened 2024-07-17
Fix `local_user_index_canister` function

hpeebles:


megrogan:
What about the tests that weren't running?

megrogan (approved):

PR #6078, opened 2024-07-19
Return proof as JWT after paying for Diamond membership

hpeebles:


backend/canisters/user_index/impl/src/updates/pay_for_diamond_membership.rs
```diff
@@ -239,7 +240,16 @@ fn process_charge(
                 .push((args.token, args.expected_price_e8s as u128));
         }
 
-        Success(result)
+        let claims = Claims::new(now + HOUR_IN_MS, "diamond_membership".to_string(), result.clone());
```
julianjelfs, line 243: Doesn't this need to include the type of subscription? Otherwise someone could offer proof for a lifetime gate by only buying 1 month sub?

hpeebles: Everything is based off the expiry date. For lifetime diamond membership then expiry date is simply 1000 years in the future.
Bit hacky, but was the simplest way to handle lifetime diamond.

hpeebles: All that the LocalUserIndexes know is the expiry date per user.

julianjelfs (approved):

PR #6087, opened 2024-07-22
Store CHIT balances per month

hpeebles:


megrogan:
I think we should return total_chit_balance too

backend/canisters/user_index/impl/src/model/user.rs
```diff
@@ -130,17 +139,22 @@ impl User {
...
+                .get(&MonthKey::from_timestamp(now))
+                .copied()
+                .unwrap_or_default(),
             streak: self.streak(now),
             is_unique_person: self.unique_person_proof.is_some(),
         }
     }
 
-    pub fn to_summary_v2(&self, now: TimestampMillis) -> UserSummaryV2 {
+    pub fn to_summary_v2(&self, now: TimestampMillis, month_key: MonthKey) -> UserSummaryV2 {
```
megrogan, line 153: No need to pass in month_key. Can derive it from now.

backend/canisters/user_index/impl/src/model/user.rs
```diff
@@ -156,9 +170,10 @@ impl User {
         }
     }
 
-    pub fn to_summary_volatile(&self, now: TimestampMillis) -> UserSummaryVolatile {
+    pub fn to_summary_volatile(&self, now: TimestampMillis, month_key: MonthKey) -> UserSummaryVolatile {
```
megrogan, line 173: No need to pass in month_key. Can derive it from now.

megrogan (dismissed):

megrogan (approved):

PR #6120, opened 2024-07-26
Ensure a lottery ticket can only win once

megrogan:


backend/canisters/airdrop_bot/impl/src/model/airdrops.rs
```diff
@@ -123,49 +121,55 @@ impl Airdrops {
...
-                .collect(),
-        };
+                    },
+                )
+            })
+            .collect();
+
+        let mut lottery_winners = Vec::new();
+
+        for i in 0..prizes {
```
hpeebles, line 154: We should break if `ticket_holders` is empty

backend/canisters/airdrop_bot/impl/src/model/airdrops.rs
```diff
@@ -123,49 +121,61 @@ impl Airdrops {
...
+
+            lottery_winners.push((
+                winner,
+                Prize {
+                    chat_won: config.lottery_prizes[i],
+                    block_index: None,
+                },
+            ));
+
+            if ticket_holders.is_empty() {
```
hpeebles, line 167: This needs to be at the start of the loop because ticket_holders could start empty

hpeebles (approved):

PR #6123, opened 2024-07-26
Fix and simplify month in airdrop messages

megrogan:


backend/canisters/airdrop_bot/impl/src/jobs/process_pending_actions.rs
```diff
@@ -7,14 +7,14 @@ use icrc_ledger_types::icrc1::transfer::{TransferArg, TransferError};
 use rand::Rng;
 use std::cell::Cell;
 use std::time::Duration;
-use time::macros::format_description;
```
hpeebles, line 10: Can we remove the dependency on `time`?

hpeebles (approved):

PR #6152, opened 2024-07-30
Message visibility to non-members of public groups

megrogan:


backend/canisters/community/impl/src/updates/create_channel.rs
```diff
@@ -75,6 +75,9 @@ fn create_channel_impl(
         return CommunityFrozen;
     }
 
+    let messages_visible_to_non_members =
+        if args.is_public { args.messages_visible_to_non_members.unwrap_or_default() } else { false };
```
hpeebles, line 79: This should be based on the existence of a gate

backend/canisters/group/impl/src/lifecycle/init.rs
```diff
@@ -23,6 +23,7 @@ fn init(args: Args) {
         args.subtype,
         args.avatar,
         args.history_visible_to_new_joiners,
+        args.is_public && args.messages_visible_to_non_members.unwrap_or(true),
```
hpeebles, line 26: This should be based on the existence of a gate

hpeebles (approved):

hpeebles (approved):

hpeebles (approved):

PR #6157, opened 2024-07-31
Post release

megrogan:


backend/libraries/group_chat_core/src/lib.rs
```diff
@@ -1572,7 +1571,6 @@ impl GroupChatCore {
 
         if public_changed || message_visbility_changed {
             let event = GroupVisibilityChanged {
-                now_public: self.is_public.value,
```
hpeebles, line 1575: Let's change the frontend over in time for the next release

hpeebles: Actually we probably want the frontend out before groups are finished

hpeebles (approved):

backend/libraries/group_chat_core/src/lib.rs
megrogan: Yes just spoken to Julian about it

PR #6179, opened 2024-08-02
Remove deleted users from OnlineUsers canister

hpeebles:


backend/canisters/online_users/api/src/updates/c2c_remove_user.rs
```diff
@@ -0,0 +1,12 @@
+use candid::{CandidType, Principal};
+use serde::{Deserialize, Serialize};
+
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub struct Args {
+    pub principal: Principal,
```
megrogan, line 6: You don't want to batch them?

hpeebles: This happens so infrequently that it doesn't make sense to batch them. I won't be pushing backdated ones.

megrogan (approved):

PR #6185, opened 2024-08-02
Use total CHIT earned for the leaderboard

hpeebles:


megrogan:
Also need to fix existing leaderboard of course

hpeebles:
> Also need to fix existing leaderboard of course

Done

megrogan (approved):

PR #6194, opened 2024-08-02
Support updating add members permission

megrogan:

PR #6195, opened 2024-08-02
Add members to channels in private communities FE

megrogan:
I have deployed this to prod_test. Seems to be working well. One thing missing is aggregation of add member events.

This depends on this PR and communities being upgraded:
https://github.com/open-chat-labs/open-chat/pull/6194

@julianjelfs it seems I didn't need to do a DB migration after all.


frontend/app/src/components/home/groupdetails/InviteUsers.svelte
```diff
@@ -7,52 +7,103 @@
...
+                        break;
+                    }
+                    default: {
+                        toastStore.showFailureToast(i18nKey("group.addMembersFailed"));
+                        break;
+                    }
+                }
+            })
+            .catch((err) => {
+                client.logError("AddMembersFailed", err);
```
julianjelfs, line 82: this will almost certainly already have been logged.

frontend/openchat-client/src/utils/chat.ts
```diff
@@ -380,6 +380,8 @@ export function mergeLocalSummaryUpdates(
                     current.latestEventIndex,
                 );
 
+                console.log(`updated.kind: ${updated.kind}`, `current.kind: ${current.kind}`);
```
julianjelfs, line 383: looks like some left over debug?

julianjelfs:
> @julianjelfs it seems I didn't need to do a DB migration after all.

This seems ... surprising. 

It can be the case that it doesn't cause an error, but that doesn't necessarily mean that it isn't wrong. It seems to me that we will get an undefined value for the new permission which isn't necessarily what we would get if we loaded the data from the server. So my inclination would be that we _should_ bump the cache and do the minimal migration.

julianjelfs (approved):

PR #6215, opened 2024-08-07
Link Internet Identity account

julianjelfs:
When manually verifying humanity or accessing a credential gated group (including humanity gated) we will now do the following: 

* load the user's auth principles and check whether they have at least one that is II
* if not we prompt them to link an II principle by
      - authenticating with the II
      - authenticating with the original auth method
      - linking the two identities together


frontend/openchat-worker/src/worker.ts
```diff
@@ -62,18 +62,18 @@ async function initialize(
...
-        return { kind: "success", identity: ocIdentity };
-    }
-
     identityAgent = new IdentityAgent(
         authProviderIdentity as SignIdentity,
         identityCanister,
         icUrl,
     );
 
+    authPrincipalString = authPrincipal.toString();
```
hpeebles, line 71: Why was this moved later? It looks unchanged so it may as well be earlier so that we can exit earlier if we have the OC identity.

julianjelfs: Because I need the identityAgent to exist so that I can look up a user's auth principals.

hpeebles (approved):

PR #6221, opened 2024-08-08
Use single shared `HttpAgent` instance rather than 1 per canister

hpeebles:


frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -276,32 +278,50 @@ export class OpenChatAgent extends EventTarget {
...
-        this._dexesAgent = new DexesAgent(config);
+        this._dexesAgent = new DexesAgent(this._agent);
         this._groupInvite = config.groupInvite;
     }
 
     private get principal(): Principal {
         return this.identity.getPrincipal();
     }
 
+    private createHttpAgent(): HttpAgent {
```
julianjelfs, line 313: might be nice if this `createHttpAgent` were just a shareable function rather than a method since it is also used (or something very similar in the identity agent and we'd want it as difficult as possible to get wrong (the whole fetchRootKey thing has been wrong in past).

hpeebles: Done

julianjelfs (approved):

PR #6225, opened 2024-08-11
Expose JSON versions of UserIndex APIs + switch frontend to new versions

hpeebles:


julianjelfs (commented):
How much control do we have over the type generation? Would be nice if we could automatically move to typescript style naming conventions. 

Also the way the union types are derived still seems a bit unfortunate. 

These things only really matter if we are striving to actually try to remove our mapping layer.

backend/canisters/user_index/api/src/main.rs
```diff
@@ -38,6 +40,43 @@ fn main() {
...
     generate_candid_method!(user_index, update_diamond_membership_subscription, update);
 
+    let directory = env::current_dir()
+        .unwrap()
+        .join("frontend/openchat-agent/tsBindings/userIndex");
+    if directory.exists() {
+        std::fs::remove_dir_all(&directory).unwrap();
+    }
+
+    generate_ts_method!(user_index, check_username);
```
megrogan, line 50: Anyway to automate this so it extracts all queries and updates without having to explicitly list them?

backend/canisters/user_index/api/src/queries/check_username.rs
```diff
@@ -1,12 +1,15 @@
 use candid::CandidType;
 use serde::{Deserialize, Serialize};
+use ts_gen::ts_export;
 
 #[derive(CandidType, Serialize, Deserialize, Debug)]
+#[ts_export(user_index, check_username)]
```
megrogan, line 6: Is it possible to do this with a macro which extracts user_index and check_username from context rather than having to state them explicitly?

hpeebles: No. I spent a while digging into that. If a macro could detect the file it was in then it would be possible, but it can't

backend/canisters/user_index/api/src/main.rs
hpeebles: Yeah we probably could. But there are some that we don't want to include which makes it harder to automate.

frontend/openchat-agent/src/utils/mapping.ts
```diff
@@ -5,9 +5,13 @@ export function optional<A, B>(candid: [] | [A], mapper: (a: A) => B): B | undef
     return candid[0] !== undefined ? mapper(candid[0]) : undefined;
 }
 
+export function optionalJson<A, B>(json: A | null, mapper: (a: A) => B): B | undefined {
+    return json != undefined ? mapper(json) : undefined;
```
julianjelfs, line 9: Weird that the input type is `A | null` and then we are checking if `json != undefined`

megrogan (approved):

PR #6238, opened 2024-08-14
New lottery algorithm for next airdrop

megrogan:


backend/canisters/user_index/impl/src/queries/users_chit.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+fn chit_for_user(user_id: &UserId, month_key: MonthKey, state: &RuntimeState) -> Chit {
+    state
+        .data
+        .users
+        .get_by_user_id(user_id)
+        .map(|user| {
+            let balance = user.chit_per_month.get(&month_key).copied().unwrap_or_default();
+            let streak = user.streak(state.env.now());
```
hpeebles, line 27: Pull `state.env.now()` out of the loop

backend/integration_tests/src/client/user_index.rs
```diff
@@ -233,26 +233,26 @@ pub mod happy_path {
         }
     }
 
-    pub fn chit_balances(
+    pub fn chit(
```
hpeebles, line 236: This should be `users_chit` if we're matching the name exposed by the canister

hpeebles (approved):

hpeebles (approved):

PR #6242, opened 2024-08-14
Set/store user's wallet configuration

megrogan:


hpeebles (approved):

backend/canisters/user/api/can.did
```diff
@@ -1189,6 +1191,27 @@ type RetrieveBtcResponse = variant {
...
+    config : WalletConfig;
+};
+
+type WalletConfig = variant {
+    Auto : AutoWallet;
+    Manual : ManualWallet;
+};
+
+type AutoWallet = record {
+    min_cents_visible : nat32;
```
hpeebles, line 1204: Nit: I think this should just be `min_cents`

julianjelfs (approved):

PR #6249, opened 2024-08-16
Wallet refactor part one

julianjelfs:
Adds a wallet config which allows you to choose auto mode whereby all tokens that meet a minimum dollar value threshold are displayed + the default tokens (the default will be auto mode with min dollar balance set to $1). You can also choose manual mode where you can manually select just the tokens that you want. 

This should help organise the wallet into something a bit more manageable.


megrogan:
Don't merge this until user canisters have been released

megrogan (approved):

PR #6303, opened 2024-09-02
Trim poll answers and check for case insensitive uniqueness

hpeebles:


frontend/app/src/components/home/PollBuilder.svelte
```diff
@@ -88,7 +91,7 @@
 
     function deleteAnswer(answer: string) {
         poll.pollAnswers.delete(answer);
```
julianjelfs, line 93: This should now be `poll.pollAnswers.delete(answer.toUpperCase())` shouldn't it?

julianjelfs (approved):

PR #6305, opened 2024-09-02
Add metrics about recently joined users

megrogan:


backend/canisters/user_index/impl/src/lib.rs
```diff
@@ -272,7 +274,47 @@ impl RuntimeState {
...
+
+                if user.diamond_membership_details.is_lifetime_diamond_member() {
+                    stats.lifetime_diamond += 1;
+                }
+
+                if user.total_chit_earned() >= 5000 {
+                    stats.more_than_5k_chit += 1;
+                }
+
+                if user.total_chit_earned() >= 20000 {
```
hpeebles, line 301: Maybe 25k? Since the lottery was 25k.

hpeebles (approved):

PR #6317, opened 2024-09-03
Support community referrals

megrogan:


backend/canisters/local_user_index/api/src/updates/join_channel.rs
```diff
@@ -10,6 +10,8 @@ pub struct Args {
     pub community_id: CommunityId,
     pub channel_id: ChannelId,
     pub invite_code: Option<u64>,
+    #[serde(default)]
```
hpeebles, line 13: No need for serde default on Options

hpeebles (approved):

PR #6319, opened 2024-09-04
Community referrals

julianjelfs:


frontend/app/src/components/home/groupdetails/GroupDetails.svelte
```diff
@@ -178,6 +180,9 @@
                 open={$groupInviteUsersOpen}
                 headerText={i18nKey("invite.inviteWithLink", undefined, chat.level, true)}>
                 <InviteUsersWithLink container={chat} />
+                {#if chat.level === "channel"}
```
megrogan, line 183: Does it make sense to show them here?

julianjelfs: My thinking was that you may have explicitly shared the channel url and so you might be surprised if it _doesn't_ show here. But ... not sure.

megrogan: It seems weird because there can be users that you have referred that aren't in the selected channel

julianjelfs: Yes probably - I'll take it out of there

PR #6329, opened 2024-09-06
Fix removing of community referrals

megrogan:


backend/canisters/community/impl/src/queries/selected_updates.rs
```diff
@@ -78,7 +79,7 @@ fn selected_updates_impl(args: Args, state: &RuntimeState) -> Response {
         match &event_wrapper.event {
             CommunityEventInternal::MembersRemoved(e) => {
                 for user_id in e.user_ids.iter() {
-                    let referral = my_referrals.map_or(false, |rs| rs.contains(user_id));
+                    let referral = my_user_id.is_some() && e.referred_by.get(user_id).copied() == my_user_id;
```
hpeebles, line 82: You should just dereference rather than using `copied`.

`*e.referred_by.get(user_id) == my_user_id;`

hpeebles (approved):

PR #6331, opened 2024-09-06
Airdrop warning

julianjelfs:
This should be ok with hard-coded values until we get onto October.


frontend/app/src/i18n/en.json
```diff
@@ -100,6 +100,7 @@
     "addedBy": "{changedBy} added {changed} to the {level}",
     "admin": "admin",
     "advanced": "Advanced",
+    "airdropWarning": "In order to qualify for the next Airdrop, you must be a member of the **{channelName}** channel in the **{communityName}** community. [Click here to join.]({url})",
```
hpeebles, line 103: "In order to qualify for the next Airdrop" should probably say "In order to qualify for the next "CHIT for CHAT" airdrop". Because otherwise just saying "... the next Airdrop" is ambiguous.

julianjelfs: I don't really think that's necessary. It's already there in the community name and I don't think there is any doubt about what it means. And I don't want the message to be any longer than it needs to be.

hpeebles: Oh yeah it'll be in the community name

hpeebles (approved):

PR #6335, opened 2024-09-06
Remove Zod generated types + add some Typebox types + a few fixes

hpeebles:


Cargo.toml
```diff
@@ -259,3 +259,4 @@ debug = false
 [patch.crates-io]
 ic-cdk-macros = { git = "https://github.com/hpeebles/cdk-rs", rev = "f20bca1b4210b37effcb3c5d82adbb05c6647cbd" }
 rmp-serde = { git = "https://github.com/hpeebles/msgpack-rust", rev = "832a3f48e67eea56c869715ae6e1045583dd011b" }
+ts-rs = { git = "https://github.com/hpeebles/ts-rs", rev = "7b402398a066b33f2bc27e1771fb2c2318f402c4" }
```
hpeebles, line 262: This is to generate Records from HashMaps rather than using the `[key: string]` syntax.

julianjelfs (approved):

frontend/openchat-client/src/utils/chat.ts
```diff
@@ -414,6 +414,8 @@ export function mergeLocalSummaryUpdates(
                             notificationsMuted:
                                 updated.notificationsMuted ?? current.membership.notificationsMuted,
                             archived: updated.archived ?? current.membership.archived,
+                            rulesAccepted:
```
hpeebles, line 417: This wasn't being populated in this arm of the if statement, but it should be

PR #6350, opened 2024-09-09
Add support for external achievements

megrogan:


backend/canisters/user_index/impl/src/model/external_achievements.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        if !achievement.awarded.insert(user_id) {
+            return AwardResult::AlreadyAwarded;
+        }
+
+        achievement.remaining_chit_budget = achievement.remaining_chit_budget - achievement.chit_reward;
+
+        if achievement.remaining_chit_budget < achievement.chit_reward {
+            achievement.budget_exhausted = Some(now);
+            achievement.awarded = HashSet::new();
```
hpeebles, line 78: Do we want to clear this so eagerly? It may be useful for debugging things

megrogan: Maybe not. I guess we can remove them in a future upgrade. I'll remove this.

backend/canisters/user_index/api/src/queries/external_achievements.rs
```diff
@@ -0,0 +1,32 @@
...
+#[derive(CandidType, Debug)]
+pub struct SuccessResult {
+    pub timestamp: TimestampMillis,
+    pub achievements_added: Vec<ExternalAchievement>,
+    pub achievements_removed: Vec<ExternalAchievement>,
+}
+
+#[ts_export(user_index, external_achievements)]
+#[derive(CandidType, Debug)]
+pub struct ExternalAchievement {
```
hpeebles, line 28: Expiry date?
Because then the backend doesn't need to return any that have expired since the frontend will already know that.

backend/canisters/user_index/impl/src/queries/external_achievements.rs
```diff
@@ -0,0 +1,43 @@
...
+fn external_achievements_impl(args: Args, state: &RuntimeState) -> Response {
+    let mut achievements_added = Vec::new();
+    let mut achievements_removed = Vec::new();
+
+    for achievement in state.data.external_achievements.iter() {
+        let add = achievement.registered > args.updates_since;
+        let remove = achievement.expires > args.updates_since
+            || achievement.budget_exhausted.map_or(false, |ts| ts > args.updates_since);
+
+        if add ^ remove {
```
hpeebles, line 19: This should be `||`, right?
Rather than `^`, which is XOR.

backend/canisters/user_index/impl/src/queries/external_achievements.rs
```diff
@@ -0,0 +1,43 @@
...
+                achievements_removed.push(a);
+            }
+        }
+    }
+
+    if achievements_added.is_empty() && achievements_removed.is_empty() {
+        SuccessNoUpdates
+    } else {
+        Success(SuccessResult {
+            timestamp: state.env.now(),
```
hpeebles, line 38: The query caching will work better if you return the timestamp of the latest update. That will cause everyone to send the same request.

backend/canisters/user_index/api/src/queries/external_achievements.rs
megrogan: Simpler without

hpeebles (approved):

hpeebles (approved):

PR #6353, opened 2024-09-10
strip meta data from images

julianjelfs:
As hard as I have tried, I cannot get dynamic resizing to work with gifs. There is a lib for reading and writing gif data but it's completely undocumented and just doesn't seem to work without mangling the gifs.


megrogan (approved):

frontend/openchat-client/src/utils/media.ts
```diff
@@ -231,8 +221,8 @@ export async function messageContentFromFile(
             if (isImage) {
                 const extract = await extractImageThumbnail(blobUrl, mimeType);
 
-                if (data.byteLength > maxSizes.image) {
-                    data = (await resizeImage(blobUrl, mimeType, isDiamond)).data;
+                if (!isGif || data.byteLength > maxSizes.image) {
+                    data = (await stripMetaDataAndResize(blobUrl, mimeType)).data;
```
hpeebles, line 225: But we should always do this, right? Not just when the byteLength exceeds the max.

julianjelfs: We do it always _if it's not a gif_ because I think it's less likely gifs will have meta data and I don't want to break _all_ gifs.

hpeebles: Ahhh woops yup I misread that if statement

hpeebles (approved):

PR #6356, opened 2024-09-10
add is_diamond to start video claim

julianjelfs:


hpeebles (approved):

hpeebles:
We can delete the old `StartVideoCall` variant actually. Might as well do that here while we're in that area.

megrogan (approved):

PR #6364, opened 2024-09-10
Leaderboard returns all_time|this_month|last_month

megrogan:


backend/canisters/user_index/impl/src/updates/c2c_notify_chit.rs
```diff
@@ -23,7 +23,10 @@ fn c2c_notify_chit_impl(args: Args, state: &mut RuntimeState) -> Response {
         now,
     ) {
         if let Some(user) = state.data.users.get_by_user_id(&user_id) {
-            state.data.chit_leaderboard.update_position(user_id, user.total_chit_earned());
+            state
+                .data
+                .chit_leaderboard
+                .update_position(user_id, user.total_chit_earned(), args.chit_balance);
```
hpeebles, line 29: This should be based on the `args.timestamp` rather than just taking the current month, otherwise if someone earns some CHIT right at the end of the month that'll be assigned to the new month incorrectly.

backend/canisters/user_index/impl/src/model/chit_leaderboard.rs
```diff
@@ -1,54 +1,122 @@
...
+            this_month: Vec::new(),
+            last_month: Vec::new(),
+            this_month_key: MonthKey::from_timestamp(now),
+        }
+    }
+
+    pub fn switch_months(&mut self, now: TimestampMillis) {
+        let mk = MonthKey::from_timestamp(now);
+
+        if mk == self.this_month_key.previous() {
```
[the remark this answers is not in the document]
hpeebles: Oh we can just use `next` rather than `previous`

backend/canisters/user_index/impl/src/model/chit_leaderboard.rs
```diff
@@ -1,54 +1,122 @@
...
-                balance: latest_balance,
-            });
+            leaderboard.push(ChitUserBalance { user_id, balance: chit });
         }
 
-        self.list.sort_unstable_by_key(|i| Reverse(i.balance));
+        leaderboard.sort_unstable_by_key(|i| Reverse(i.balance));
+    }
+
+    pub fn all_time(&self) -> &Vec<ChitUserBalance> {
```
hpeebles, line 110: Use a slice rather than a reference to a vec

hpeebles (approved):

PR #6372, opened 2024-09-11
Reset all cached chat data

hpeebles:


julianjelfs (commented):
Why do we want to do this? 

It seems wrong to me. 

There should _always_ be a migration otherwise there would be no need to update the cache version. 

IMO we should _explicitly not_ give the option to splat the DB because it implies that we haven't understood what actually needs to be done. i.e. what the correct migration is. 

And we shouldn't worry about running through _all of the migrations_ because people typically won't because they will be on the previous version.

PR #6382, opened 2024-09-12
Referred by gate

julianjelfs:
You might notice that I am not actually checking the gate at all client side. I _could_ but it would be a bit of a pain in the arse. I would prefer not to on the basis that it doesn't really matter and it's pretty doubtful that anyone is really going to use this gate (other than Multi).


hpeebles (approved):

megrogan:
I'd like to discuss this

PR #6430, opened 2024-09-20
Display tick once messages have been accepted by the canister

hpeebles:


frontend/openchat-agent/src/services/candidService.ts
```diff
@@ -88,12 +89,18 @@ export abstract class CandidService {
                 this.agent.call(this.canisterId, {
                     methodName: methodName + "_msgpack",
                     arg: payload,
+                    callSync: onRequestAccepted === undefined,
```
hpeebles, line 92: If there is no `onRequestAccepted` action then we should use the synchronous endpoint which will have slightly faster responses than the async polling version.

julianjelfs (approved):

julianjelfs (approved):

PR #6452, opened 2024-09-24
Remove the unused `use_for_new_canisters` field from upgrade args

hpeebles:


megrogan:
You say "unused". Has it never been used?

megrogan (approved):

PR #6453, opened 2024-09-24
Support upgrading to large wasms by uploading in chunks

hpeebles:


backend/canisters/group_index/impl/src/updates/upload_wasm_chunk.rs
```diff
@@ -0,0 +1,21 @@
+use crate::{mutate_state, RuntimeState};
+use canister_tracing_macros::trace;
+use group_index_canister::upload_wasm_chunk::{Response::*, *};
+use ic_cdk::update;
+
+#[update]
```
megrogan, line 6: No guard?

megrogan (approved):

megrogan (approved):

PR #6459, opened 2024-09-26
Allow changing PIN number if signed in within last 5 minutes

hpeebles:


backend/canisters/user/impl/src/updates/set_pin_number.rs
```diff
@@ -20,30 +22,50 @@ fn set_pin_number(args: Args) -> Response {
...
-        if let Some(new) = args.new.as_ref() {
-            let length = new.len();
-            if length < MIN_LENGTH {
-                return TooShort(FieldTooShortResult {
-                    length_provided: length as u32,
-                    min_length: MIN_LENGTH as u32,
-                });
+    if state.data.pin_number.enabled() {
+        match args.verification {
+            PinNumberVerification::None => return PinRequired,
```
julianjelfs, line 27: Why does this variant exist?

megrogan: If you set it for the first time you don't need an existing pin

megrogan (dismissed):

megrogan (approved):

PR #6479, opened 2024-09-30
Video fixes

### LINKED ISSUE #6456: Couple of video call issues
First
====

Sometimes the call message in OpenChat gets stuck and never flips to the ended state. There are a couple of possible explanations. Either the call from the video bridge to OC fails (unlikely) or it is never made. This could be because the call is so short that Daily does not consider it a meeting and therefore never fires the relevant hook. Even so the fallback polling should catch it so it's not quite clear what causes it. 

Second
======

Duplicate frame error that we sometimes see. Work out what specifically causes that and do something about it. 

Third
=====

You can see the meeting message and join the meeting _before_ the incoming call modal pops up, then the incoming call modal opens after you have already joined. (This one _really_ seems impossible).

julianjelfs:
fixes #6456 

Should hopefully get rid of the Duplicate iframe error that we see quite regularly.


hpeebles (approved):

frontend/app/src/i18n/cn.json
```diff
@@ -647,10 +647,10 @@
         "joined_group": "加入群组",
         "nothingDone": "你还有很多工作要做！",
         "nothingLeftToDo": "放松！你没什么可做的了",
-        "owned_group_with_1000_diamond_members": "拥有一个拥有 1000 名经过验证的成员的群组/社区",
-        "owned_group_with_100_diamond_members": "拥有一个拥有 100 名经过验证的成员的群组/社区",
-        "owned_group_with_10_diamond_members": "拥有一个拥有 10 名经过验证的成员的群组/社区",
         "owned_group_with_1_diamond_member": "拥有一个拥有 1 名经过验证的成员的群组/社区",
+        "owned_group_with_10_diamond_members": "拥有一个拥有 10 名经过验证的成员的群组/社区",
```
megrogan, line 651: Balls. Looks like the "merge-latest-translations" function orders these differently to sync_lang

PR #6482, opened 2024-09-30
clear everything from the cache

julianjelfs:


frontend/openchat-agent/src/utils/caching.ts
```diff
@@ -202,14 +210,15 @@ const migrations: Record<number, MigrationFunction<ChatSchema>> = {
     111: clearChatsStore,
     112: clearCommunityDetails,
     113: createExternalAchievementsStore,
+    114: clearEverything,
```
megrogan, line 213: Is it necessary to keep these migrations. Surely we only need to do a nuke?

julianjelfs: We can remove them. Though doing so slightly bothers me for reasons I'm not 100% sure about.

megrogan (approved):

PR #6552, opened 2024-10-10
Push activity to users using GroupedTimerJobQueue

megrogan:


backend/libraries/timer_job_queues/src/grouped_timer_job_queue.rs
```diff
@@ -197,6 +197,12 @@ where
     }
 }
 
+impl<T: TimerJobItemGroup> Default for GroupedTimerJobQueue<T> {
+    fn default() -> Self {
+        GroupedTimerJobQueue::new(5, true)
```
hpeebles, line 202: That's quite an opinionated default. Previously the dev would have to think out the max concurrency which I think is better than having it set for them.

backend/canisters/community/impl/src/updates/accept_p2p_swap.rs
```diff
@@ -33,15 +34,37 @@ async fn accept_p2p_swap(args: Args) -> Response {
...
+                                message_index: message.message_index,
+                                activity: MessageActivity::P2PSwapAccepted,
+                                timestamp: state.env.now(),
+                                user_id: Some(user_id),
+                            }),
+                        );
+                    }
+                }
+
+                state.data.user_event_sync_queue.flush();
```
hpeebles, line 66: Lets just call `handle_activity_notification(state);` again

backend/canisters/community/impl/src/updates/start_video_call.rs
```diff
@@ -56,7 +56,7 @@ fn start_video_call_impl(args: Args, state: &mut RuntimeState) -> Response {
             .collect(),
         }),
         None,
-        Vec::new(),
+        &Vec::new(),
```
hpeebles, line 59: &Vec is a bit weird. Can this just be &[]?

hpeebles (approved):

PR #6573, opened 2024-10-14
Ensure members marked as lapsed in updates queries

megrogan:


backend/canisters/community/impl/src/queries/selected_updates.rs
```diff
@@ -110,6 +113,14 @@ fn selected_updates_impl(args: Args, state: &RuntimeState) -> Response {
...
+    for (user_id, update) in state.data.members.iter_latest_updates(args.updates_since) {
+        match update {
+            MemberUpdate::Lapsed | MemberUpdate::Unlapsed => {
+                user_updates_handler.mark_member_updated(&mut result, user_id, false, false);
+            }
+        }
+    }
+
     for member in data.members.iter() {
         if member.display_name().timestamp > args.updates_since {
```
hpeebles, line 125: Let's put display name updates into the `user_updates` BTreeSet so that we can avoid iterating over every each time we get updates

hpeebles (dismissed):

hpeebles (approved):

PR #6593, opened 2024-10-16
Update canisters post release

hpeebles:


backend/canisters/community/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -30,25 +29,6 @@ fn post_upgrade(args: Args) {
         finalize_group_import(group_id);
     }
```
megrogan, line 31: Also remove the serde default and pub from updates field of CommunityMembers

megrogan (approved):

PR #6604, opened 2024-10-17
Mark a group/community active if a member lapses

megrogan:


julianjelfs (approved):

megrogan:
> What about when they unlapse? It should be the same, right?

They don't unlapse in a job - they unlapse when a user rejoins or a gate is removed and in both these cases there is already a call to "handle_activity_notification".

hpeebles (approved):

PR #6671, opened 2024-10-24
Trigger migrating events to stable memory using canister timer job

hpeebles:


backend/canisters/local_group_index/impl/src/jobs/migrate_events_to_stable_memory.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    state
+        .data
+        .canisters_pending_events_migration_to_stable_memory
+        .get(random)
+        .copied()
+}
+
+async fn migrate_events(canister_id: CanisterId) {
+    if let Ok(true) = group_canister_c2c_client::c2c_migrate_events_to_stable_memory(canister_id, &Empty {}).await {
```
hpeebles, line 45: Groups and communities expose the same interface, so its fine to just use the `group_canister_c2c_client` here.

A response of `true` means the migration is complete.

megrogan (approved):

PR #6672, opened 2024-10-24
Allow external achievements to be updated

megrogan:


backend/canisters/user_index/impl/src/model/external_achievements.rs
```diff
@@ -24,11 +24,12 @@ pub struct ExternalAchievementInternal {
 }
 
 impl ExternalAchievements {
+    // This will either a register a new external achievement or update an existing entry with the same id
```
hpeebles, line 27: Typo

megrogan: I had fixed in VS code and not saved the file!

hpeebles (approved):

PR #6700, opened 2024-11-04
Update the canister creation fee to 0.5T

hpeebles:


megrogan (commented):
What about storage index?

megrogan (approved):

megrogan (approved):

PR #6703, opened 2024-11-04
Set the derivation origin when checking verifiable credentials

hpeebles:


megrogan:
We need to get this is (having tested it somehow!)

julianjelfs (approved):

PR #6716, opened 2024-11-05
Allow marking the message activity feed read when marking messages read

hpeebles:


megrogan:
This is now redundant right?

PR #6720, opened 2024-11-05
Take payment before registering ext achievement

megrogan:


backend/canisters/user_index/impl/src/updates/register_external_achievement.rs
```diff
@@ -1,30 +1,109 @@
...
+use candid::Principal;
 use canister_api_macros::proposal;
 use canister_tracing_macros::trace;
+use icrc_ledger_types::icrc2::transfer_from::TransferFromArgs;
+use tracing::error;
+use types::{CanisterId, Cryptocurrency, TimestampNanos};
 use user_index_canister::register_external_achievement::{Response::*, *};
 use user_index_canister::ExternalAchievementInitial;
 
+const CHIT_PER_CHAT: u128 = 5000;
```
hpeebles, line 12: Maybe we should rename this because it makes it seem like they are convertible.
Something like `CHAT_FEE_PER_CHIT_AWARD` and it would be 1/5000th of a CHAT.

hpeebles (approved):

PR #6761, opened 2024-11-07
Stream message hydration for activity feed

julianjelfs:


frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -3799,74 +3800,138 @@ export class OpenChatAgent extends EventTarget {
...
         }, new AsyncMessageContextMap<number>());
 
-        // look up all the messages correlated with the activity events (mostly this will come from the cache)
+        const [withCachedMessages, missing] = await this.getMessagesByMessageContext(
+            eventIndexesByMessageContext,
+            activityEvents,
+            "cached",
+        );
+
+        const numberOfMissing = [...missing.values()].reduce((total, s) => total + s.size, 0);
```
hpeebles, line 3831: We don't need to calculate `numberOfMissing`, we just need to know if there are any.

hpeebles: Maybe something like `const anyMissing = [...missing.values()].some((s) => s.size > 0);`?

hpeebles (dismissed):

frontend/openchat-agent/src/services/openchatAgent.ts
julianjelfs: Yes that's a bit nicer

hpeebles (approved):

PR #6764, opened 2024-11-07
Wire up `freeze_community` in `inspect_message`

hpeebles:


julianjelfs (commented):
Do we need unfreeze community added somewhere too?

hpeebles:
> Do we need unfreeze community added somewhere too?

Ah yes! I'll add that

julianjelfs (approved):

PR #6811, opened 2024-11-13
Require a listing fee to add non-SNS tokens

megrogan:


backend/canisters/registry/impl/src/model/tokens.rs
```diff
@@ -53,6 +54,7 @@ impl Tokens {
                 added: now,
                 enabled: true,
                 last_updated: now,
+                payments: if let Some(p) = payment { vec![p] } else { Vec::new() },
```
hpeebles, line 57: I think you should be able to do `payment.into_iter().collect()`

backend/canisters/registry/api/src/updates/add_token.rs
```diff
@@ -2,11 +2,12 @@ use crate::TokenStandard;
 use candid::CandidType;
 use human_readable::{HumanReadablePrincipal, ToHumanReadable};
 use serde::{Deserialize, Serialize};
-use types::CanisterId;
+use types::{CanisterId, UserId};
 
 #[derive(CandidType, Serialize, Deserialize, Debug)]
 pub struct Args {
     pub ledger_canister_id: CanisterId,
+    pub submitted_by: Option<UserId>,
```
hpeebles, line 10: When processing the token it gets rejected if this isn't set, so this should probably be a required field and we just get the website out first

hpeebles:
You need to update the encoded proposal on the frontend with the `payer` field

hpeebles (approved):

PR #6831, opened 2024-11-15
Add new prize message criteria on backend

megrogan:


hpeebles (approved):

hpeebles:
You should add a changelog entry for the LocalUserIndex too since the messages flow through that

PR #6847, opened 2024-11-19
Simplify `inspect_message`

hpeebles:


backend/canisters/cycles_dispenser/impl/src/lifecycle/inspect_message.rs
```diff
@@ -5,12 +5,6 @@ use ic_cdk::inspect_message;
 fn inspect_message() {
     let method_name = ic_cdk::api::call::method_name();
 
-    // 'inspect_message' only applies to ingress messages so calls to c2c methods should be rejected
-    let is_c2c_method = method_name.starts_with("c2c");
```
hpeebles, line 9: I removed the handling of Frozen so that anyone calling gets back the correct response rather than the call getting rejected.

backend/canisters/community/impl/src/lifecycle/inspect_message.rs
```diff
@@ -11,9 +11,7 @@ fn accept_if_valid(state: &RuntimeState) {
 
     // 'inspect_message' only applies to ingress messages so calls to c2c methods should be rejected
     let is_c2c_method = method_name.starts_with("c2c") || method_name == "wallet_receive";
-    let is_frozen = state.data.frozen.value.is_some();
```
hpeebles, line 14: I removed the handling of Frozen so that anyone calling gets back the correct response rather than the call getting rejected.

megrogan (approved):

PR #6877, opened 2024-11-22
Avoid iterating all users when determining who to notify of new message

hpeebles:


backend/canisters/community/impl/src/updates/send_message.rs
```diff
@@ -214,20 +214,6 @@ fn process_send_message_result(
             let message_index = message_event.event.message_index;
             let message_id = message_event.event.message_id;
             let expires_at = message_event.expires_at;
-
-            // Exclude suspended members and bots from notification
-            let users_to_notify: Vec<UserId> = result
```
hpeebles, line 219: This is already handled

megrogan (approved):

PR #6970, opened 2024-12-04
Check bot + user permissions when issuing JWT

megrogan:


backend/canisters/community/impl/src/lib.rs
```diff
@@ -686,6 +691,38 @@ impl Data {
                 .push(user_id, CommunityCanisterEvent::Achievement(achievement));
         }
     }
+
+    pub fn get_bot_permissions(&self, bot_user_id: &UserId) -> Option<&SlashCommandPermissions> {
+        self.bot_permissions.get(bot_user_id)
+    }
+
+    pub fn get_user_permissions(&self, user_id: &UserId, channel_id: &ChannelId) -> Option<SlashCommandPermissions> {
```
hpeebles, line 699: `get_user_permissions` returns `Option<SlashCommandPermissions>`, that doesn't seem right.

Maybe `SlashCommandPermissions` should be renamed to something which applies to users and bots.

hpeebles: Ahh right, this is getting the user permissions specifically for executing bot commands.
So the return type is fine, but the method name should be clearer.
`get_user_permissions_for_bot_commands` or something.

backend/libraries/group_chat_core/src/roles.rs
```diff
@@ -179,4 +184,46 @@ impl GroupRoleInternal {
     fn has_admin_rights(&self) -> bool {
         self.is_admin() || self.is_owner()
     }
+
+    pub fn permissions(&self, role_permissions: &GroupPermissions) -> HashSet<GroupPermission> {
+        let permissions = vec![
```
hpeebles, line 189: You can remove the `vec!` and just leave it as an array.

backend/libraries/group_chat_core/src/roles.rs
```diff
@@ -179,4 +184,46 @@ impl GroupRoleInternal {
...
+        ];
+
+        permissions
+            .into_iter()
+            .filter_map(|(rp, p)| self.is_permitted(rp).then_some(p))
+            .collect()
+    }
+
+    pub fn message_permissions(&self, role_permissions: &MessagePermissions) -> HashSet<MessagePermission> {
+        let permissions = vec![
```
hpeebles, line 209: One more!

hpeebles: I'm surprised clippy isn't picking these up...

PR #7004, opened 2024-12-06
Add msgpack and ts support to `register_bot`

megrogan:


backend/canisters/user_index/impl/src/updates/register_bot.rs
```diff
@@ -18,7 +18,7 @@ const MAX_AVATAR_SIZE: usize = 250_000;
 const MAX_DESCRIPTION_LEN: usize = 10_000;
 const MAX_COMMANDS: usize = 100;
 
-#[proposal(guard = "caller_is_governance_principal")]
+#[proposal(guard = "caller_is_governance_principal", msgpack = true)]
```
hpeebles, line 21: The proposal macro isn't wired up to support msgpack.
We can add it though.

PR #7039, opened 2024-12-11
Update README.md

RohanCyberOps:
readme is not nicely represented


hpeebles:
Sorry for being slow on this! We've been swamped with other priorities but we'll try to get round to this later this week

hpeebles:
Thanks for putting time into this, but I'm going to close this since it isn't the voice we are going for with the readme.

PR #7052, opened 2024-12-12
Bot name uniqueness check

julianjelfs:


frontend/app/src/components/bots/BotBuilder.svelte
```diff
@@ -33,12 +38,37 @@
...
+                return errors;
+            },
             300,
             new ValidationErrors(),
         ),
     );
 
+    function checkUsername(value: string): Promise<ValidationErrorMessages> {
+        return client
+            .checkUsername(value, true)
```
hpeebles, line 55: We need a bot specific version of this because bot names can clash with usernames.
We had to allow this because the register bot proposals can take a few days and anyone in that time could register the same name and cause them to fail.

megrogan: @hpeebles checkUsernme takes a bool now which means it can be used for users or bots

megrogan (approved):

PR #7073, opened 2024-12-17
Add `update_bot` endpoint only callable by bot owner

megrogan:


Cargo.toml
```diff
@@ -189,7 +189,7 @@ ic-agent = "0.39.1"
 ic-canister-sig-creation = "1.1.0"
 ic-captcha = "1.0.0"
 ic-cbor = "2.5.0"
-ic-cdk = "0.17.0"
+ic-cdk = { version = "0.17.0", features = ["transform-closure"] }
```
hpeebles, line 192: I think the feature should be enabled just where its needed rather than at the workspace level. Otherwise it'll always build with that feature unnecessarily.

hpeebles (approved):

PR #7203, opened 2025-01-10
Partnrship addition to OpenChat Frame Ancestors

megalinker:
Hi Openchat Team, Jesús from the Partnrship team here.

Partnrship is an upcoming atomic matchmaking and management tool for projects and workers, mainly focused on the ICP ecosystem. This platform will include a kanban board-styled webapp with tasks for workers to complete. Each task will have an OpenChat dedicated channel, and we'd like to integrate your xframe to embed these channels directly within our platform.

This pull request adds our frontend canister to the frame ancestors to allow said functionality.

Best Regards.


hpeebles (approved):

megalinker:
Hi again,

The link was merged into the ancestors list about 5 days ago, and the merge was successful. However, the xframe console log still shows the old ancestors list without the new link, which is causing a Content Security Policy violation because the new link hasn't been incorporated.

Could this be due to caching, an update schedule, or something else? Is there a typical delay before such changes are reflected?

Thanks for everything!

PR #7243, opened 2025-01-16
Quick reactions

ivan-jukic:
Each message should now get a set of three quick reactions and a new menu icon.

For quick reactions we take top three from the user's favourites. If there are no favourite icons yet (i.e. as would be the case for new users), we show three default ones.

Quick reactions are not yet available on mobile.

### DIFF
--- a/frontend/app/src/components/home/ChatMessage.svelte
+++ b/frontend/app/src/components/home/ChatMessage.svelte
@@ -620,6 +620,12 @@
                         <pre>expiresAt: {expiresAt}</pre>
                         <pre>thread: {JSON.stringify(msg.thread, null, 4)}</pre>
                         <pre>botContext: {JSON.stringify(botContext, null, 4)}</pre>
+                        <pre>inert: {inert}</pre>
+                        <pre>canRevealDeleted: {canRevealDeleted}</pre>
+                        <pre>canlRevealBlocked: {canRevealBlocked}</pre>
+                        <pre>readonly: {readonly}</pre>
+                        <pre>showChatMenu: {showChatMenu}</pre>
+                        <pre>intersecting: {intersecting}</pre>
                     {/if}
 
                     {#if showChatMenu && intersecting}
@@ -652,6 +658,8 @@
                             {crypto}
                             translatable={canTranslate}
                             {translated}
+                            {canReact}
+                            createdUser={user}
                             on:collapseMessage
                             on:forward
                             on:reply={reply}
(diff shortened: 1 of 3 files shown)


frontend/app/src/components/home/ChatMessageMenu.svelte
```diff
@@ -260,13 +290,66 @@
...
+                    }
+                });
+        }
+    }
 </script>
 
-<div class="menu" class:rtl={$rtlStore}>
+<div class="menu" class:inert={inert} class:rtl={$rtlStore}>
+    {#await loadedReactions }
+        <div></div>
```
julianjelfs, line 323: Given that we don't have any meaningful placeholder while waiting here we should just use the shorter syntax: 

```
{#await loadedReactions then [fst, snd, thrd]}
```

frontend/app/src/components/home/ChatMessageMenu.svelte
```diff
@@ -260,13 +290,66 @@
...
+                    </div>
+                </HoverIcon>
+            {/if}
+        {/if}
+    {/await}
     <MenuIcon bind:this={menuIcon} centered position={"right"} align={"end"}>
         <div class="menu-icon" slot="icon">
             <HoverIcon compact>
-                <ChevronDown size="1.6em" color={me ? "#fff" : "var(--icon-txt)"} />
+                <DotsVertical size="1.4em" color="#fff" />
```
julianjelfs, line 352: This colour will probably cause theme issues and I would suggest that we use `var(--menu-txt)` rather than `#fff`.

frontend/app/src/components/home/ChatMessageMenu.svelte
```diff
@@ -512,24 +595,64 @@
...
+        }
+
+        z-index: 1;
+        opacity: 0;
+        display: flex;
         position: absolute;
-        top: -4px;
-        right: $offset;
+        width: fit-content;
+        background-color: var(--reaction-bg);
```
julianjelfs, line 619: Again, likely to cause theme issues and I would instinctively want to use `var(--menu-bg)` rather than `var(--reaction-bg)`.

ivan-jukic: I'm using the `var(--menu-bg)` for the border colour around the quick actions menu.

julianjelfs: Seems like you are using `var(--menu-bd)` for the border. In the theme `bd` is shorthand for border and `bg` is shorthand for background. So I'm suggesting we use `var(--menu-bg)` for the background and `var(--menu-bd)` for the border. At least I think that is the combo which is least likely to have theme problems.

ivan-jukic: That's actually correct; I changed it locally to test something, and then forgot about it 🙈

frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -722,29 +730,36 @@
...
-        }
+    :global(.message-bubble .menu) {
+        z-index: -1;
+        opacity: 0;
+    }
 
-        :global(.me .menu-icon:hover .wrapper) {
-            background-color: var(--icon-inverted-hv);
-        }
+    :global(.menu:has(.menu-icon.open)) {
```
julianjelfs, line 753: I think this would be safer as this: 

`:global(.message-bubble .menu:has(.menu-icon.open))`

as I fear that, because it's global we might accidentally hit something we don't intend to hit as it is.

frontend/app/src/components/home/ChatMessageMenu.svelte
```diff
@@ -260,13 +286,67 @@
...
+                    createdUser.displayName,
+                    kind,
+                )
+                .then((success) => {
+                    if (success && kind === "add") {
+                        client.trackEvent("reacted_to_message");
+
+                        // Note: Manually selected reactions do not increment
+                        // their fav counter by default, so we do it manually.
+                        emojiDb.incrementFavoriteEmojiCount(reaction);
```
julianjelfs, line 313: To make this reactive I think we would _also_ need to re-compute `loadedReactions`. We probably _will_ because as it stands that expression will only be evaluated when this component mounts.

julianjelfs (approved):

PR #7253, opened 2025-01-20
Support updating bot principal but not name

megrogan:


backend/canisters/user_index/impl/src/updates/update_bot.rs
```diff
@@ -89,16 +87,6 @@ fn update_bot_impl(args: Args, state: &mut RuntimeState) -> Response {
 }
 
 fn validate(args: &Args, state: &RuntimeState) -> Result<(), Response> {
-    if let Some(name) = args.name.as_ref() {
```
hpeebles, line 92: We should check that the principal is valid, eg. it can't be the anonymous principal.
I think that's probably all we need to check, maybe minimum length 10?

hpeebles (approved):

PR #7266, opened 2025-01-21
Expanded functionality of the message hover menu

ivan-jukic:
Added call-to-action buttones after emoji reactions. Depending on the message users may have different options available.

Emoji dialog button was moved to the hover menu, with the same functionality, and changed icon.

Icons for reply-in-thread, remind-me, and cancel-remind-me have been changed.

image

### DIFF
--- a/frontend/app/src/components/home/ChatMessage.svelte
+++ b/frontend/app/src/components/home/ChatMessage.svelte
@@ -38,7 +38,6 @@
     import { now } from "../../stores/time";
     import { createEventDispatcher, getContext, onDestroy, onMount, tick } from "svelte";
     import { dclickReply } from "../../stores/settings";
-    import EmoticonLolOutline from "svelte-material-icons/EmoticonLolOutline.svelte";
     import Close from "svelte-material-icons/Close.svelte";
     import ForwardIcon from "svelte-material-icons/Share.svelte";
     import UnresolvedReply from "./UnresolvedReply.svelte";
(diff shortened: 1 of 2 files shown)


frontend/app/src/components/home/ChatMessage.svelte
```diff
@@ -696,16 +699,6 @@
                             on:remindMe={remindMe} />
                     {/if}
                 </div>
-
```
julianjelfs, line 699: This still needs to be present for mobile otherwise we have no way to launch the emoji picker for reactions. 

Unless we have something inside the menu on mobile to show the favourite emojis + something to launch the picker.

frontend/app/src/components/home/ChatMessageMenu.svelte
```diff
@@ -272,11 +278,50 @@
                 </div>
             </HoverIcon>
         {/each}
+        {#if canReact && !failed}
+            <HoverIcon compact={true} onclick={() => showEmojiPicker()} title={$_("pickEmoji")}>
```
julianjelfs, line 282: Instead of `compact={true}` you can just use the shorthand `compact`

A few more similar cases below.

frontend/app/src/components/home/ChatMessageMenu.svelte
```diff
@@ -272,11 +278,50 @@
...
+            </HoverIcon>
+        {/if}
+        {#if confirmed && supportsReply && !failed}
+            {#if !inThread && canStartThread}
+                <HoverIcon
+                    compact={true}
+                    onclick={() => initiateThread()}
+                    title={$_("thread.menu")}>
+                    <div class="quick-reaction">
+                        <ChatPlusOutline size={$iconSize} color={"var(--menu-txt)"} />
```
julianjelfs, line 295: I feel like these extra icons are a little bit too big. They seem noticeably bigger than the favourite emojis and seem like they would look better with a bit more space. I think (personally) that something like 1.2em looks good.

PR #7276, opened 2025-01-22
Verified groups / communities

julianjelfs:
Add proposal types for verifying and revoking verification of groups and communities. The user making the proposal must provide evidence of officialdom. The electorate then weighs the evidence and votes. If verified the community / group gets the blue tick. When making the proposal it is also possible to choose a preferred name (in case someone else is squatting).


frontend/app/src/components/home/Home.svelte
```diff
@@ -225,6 +225,7 @@
             ? selectedMultiUserChat.subtype?.governanceCanisterId
             : undefined;
     $: nervousSystem = client.tryGetNervousSystem(governanceCanisterId);
+    // $: nervousSystem = client.tryGetNervousSystem("rrkah-fqaaa-aaaaa-aaaaq-cai");
```
hpeebles, line 228: This can go

megrogan (approved):

PR #7279, opened 2025-01-22
Return bots removed in `bot_updates` response

hpeebles:


megrogan (dismissed):

backend/canisters/user_index/api/src/queries/bot_updates.rs
```diff
@@ -21,6 +21,7 @@ pub enum Response {
 pub struct SuccessResult {
     pub added_or_updated: Vec<BotSchema>,
     pub deleted: Vec<UserId>,
+    pub removed: Vec<UserId>,
```
megrogan, line 24: As this isn't live yet let's just delete the "deleted" field

PR #7280, opened 2025-01-22
Fix min group / channel name length

hpeebles:


megrogan:
Please don't merge yet

PR #7289, opened 2025-01-23
Extract bot userId from events

hpeebles:


ivan-jukic (dismissed):

frontend/openchat-client/src/utils/chat.ts
```diff
@@ -184,8 +184,10 @@ export function activeUserIdFromEvent(event: ChatEvent): string | undefined {
         case "group_chat_created":
             return event.created_by;
         case "members_added":
+        case "bot_added":
             return event.addedBy;
         case "members_removed":
+        case "bot_removed":
             return event.removedBy;
```
julianjelfs, line 191: This will actually not ultimately work because we will not be able to look up the removed bot. The bot removed event will need to include the bot name when it was removed. But we can deal with that when that change is made.

julianjelfs (approved):

PR #7313, opened 2025-01-27
Allow sign-in provider unlinking

ivan-jukic:
Any sign-in providers that the user has used to sign in with OC can now be unlinked.

---
TODO:
- [x] integration tests
- [x] translations


backend/canisters/identity/impl/src/updates/remove_identity_link.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[update]
+#[trace]
+fn remove_identity_link(args: Args) -> Response {
+    mutate_state(|state| remove_identity_link_impl(args, state))
+}
+
+fn remove_identity_link_impl(args: Args, state: &mut RuntimeState) -> Response {
+    let caller = state.env.caller();
+    let auth_principal = state.data.user_principals.get_auth_principal(&caller);
```
hpeebles, line 14: I think you can skip fetching this this because everything you need should be returned by `get_by_auth_principal`

backend/canisters/identity/impl/src/updates/remove_identity_link.rs
```diff
@@ -0,0 +1,38 @@
...
+    let user_principal = state
+        .data
+        .user_principals
+        .get_by_auth_principal(&caller)
+        .filter(|u| u.user_id.is_some());
+
+    let principals = auth_principal.zip(user_principal);
+
+    if let Some((auth, user)) = principals {
+        if user.principal == args.linked_principal {
```
hpeebles, line 24: Each user has a single "user principal" but can have multiple "authentication principals".
The call into the Identity canister will be made using whichever auth principal they signed in with.
So we should return `CannotUnlinkActivePrincipal` if `caller == args.linked_principal`.

backend/canisters/identity/impl/src/model/user_principals.rs
```diff
@@ -94,6 +94,25 @@ impl UserPrincipals {
         }
     }
 
+    pub fn unlink_auth_principal(&mut self, linked_principal: Principal, user_principal_index: u32) -> bool {
```
hpeebles, line 97: An authentication principal can only be linked to a single user so we don't actually need to pass in the `user_principal_index`, we can instead grab it from the result of `get_by_auth_principal`.
So maybe this method just becomes `remove_auth_principal`? Then the fact that it becomes unlinked from the user is simply a result of it having been removed.

ivan-jukic: Renamed fn and moved most of the functionality into `remove_auth_principal`.

backend/canisters/identity/impl/src/model/user_principals.rs
```diff
@@ -94,6 +95,26 @@ impl UserPrincipals {
         }
     }
 
+    pub fn remove_auth_principal(&mut self, caller: Principal, linked_principal: Principal) -> RemovePrincipalResponse {
```
ivan-jukic, line 98: I think that the return type here is a bit weird; I'm not sure what your expectations/opinions are on this.

The way I see it, ideally this function would return a result containing an error type - if we think that responses should contextually stay separated from the functionality implementation - , that could be converted into a response.

frontend/app/src/i18n/en.json
```diff
@@ -829,7 +829,11 @@
             "loginApprover": "Sorry we were unable to sign into your current OpenChat authentication provider",
             "loginInitiator": "Sorry we were unable to sign into your chosen provider",
             "pollingError": "There was an unexpected error waiting for email signin",
-            "principalMismatch": "It seems like you did not re-authenticate with the same provider that you are currently signed in with"
+            "principalMismatch": "It seems like you did not re-authenticate with the same provider that you are currently signed in with",
+            "unlinkActivePrincipal": "It seems like you tried to unlink a provider you're currently logged in with, which is not allowed.",
```
hpeebles, line 833: I think "You cannot unlink the provider you are currently logged in with." is better.

hpeebles (dismissed):

hpeebles (dismissed):

frontend/app/src/components/home/profile/LinkedAuthAccounts.svelte
```diff
@@ -57,6 +71,19 @@
                     </div>
                 </TooltipWrapper>
             </div>
+        {:else}
+            <div class="unlink">
+                <TooltipWrapper position="top" align="end">
+                    <Button slot="target" cls="btn" hollow on:click={() => (unlinking = account)}>
```
julianjelfs, line 77: I think this would be more consistent with the existing UI if we simply used the `HoverIcon` component. e.g. 

```
<HoverIcon slot="target" onclick={() => (unlinking = account)}>
    <LinkOff size={$iconSize} color={"var(--icon-txt)"} />
</HoverIcon>
```

frontend/app/src/components/home/profile/LinkedAuthAccounts.svelte
```diff
@@ -71,8 +98,28 @@
         display: flex;
         align-items: center;
 
-        .current {
+        .current,
+        .unlink {
             margin-inline-start: auto;
         }
+
+        .unlink :global(.btn) {
```
julianjelfs, line 106: using `HoverIcon` would remove the need for all this special case css too.

frontend/app/src/components/home/profile/UnlinkAccounts.svelte
```diff
@@ -0,0 +1,133 @@
+<script lang="ts">
+    import LinkVariantOff from "svelte-material-icons/LinkVariantOff.svelte";
```
julianjelfs, line 2: Since this is a new component we should probably use Svelte 5 syntax since we'll have to convert everything in the end.

frontend/app/src/components/home/profile/UnlinkAccounts.svelte
```diff
@@ -0,0 +1,133 @@
...
+    {#if step == "unlink"}
+        <AlertBox>
+            <p class="info">
+                <Translatable
+                    resourceKey={i18nKey("identity.linkedAccounts.unlinkAdvice", {
+                        provider: account.provider,
+                    })} />
+            </p>
+        </AlertBox>
+        <p class="unlink-note">
```
julianjelfs, line 71: This looks a bit odd to me. Is there a reason why it is not _inside_ the `AlertBox`? If it is a question of having line breaks, you could use a `Markdown` component within the `AlertBox`

frontend/app/src/components/home/profile/UnlinkAccounts.svelte
```diff
@@ -0,0 +1,133 @@
...
+        @include font(bold, normal, fs-130, 29);
+        margin-bottom: $sp4;
+        display: flex;
+        align-items: center;
+        gap: $sp3;
+    }
+
+    .content {
+        .unlink-note {
+            padding-left: 3rem;
```
julianjelfs, line 122: We're using a mixture of rems and the $sp variables from the mixins file. I would prefer if we used the $sp variables as much as we can so that our spacings move in consistent increments.

frontend/app/src/components/home/profile/UnlinkAccounts.svelte
```diff
@@ -0,0 +1,133 @@
...
+            margin-bottom: 2rem;
+            margin-top: -0.5rem;
+            font-size: 14px;
```
julianjelfs, line 125: Ditto font sizes. We have a font-size mixin that allows us to use a consistent set of font size increments (in the hope that overall things will look more consistent).

frontend/app/src/components/home/profile/UnlinkAccounts.svelte
ivan-jukic: Didn't feel like the message fits within the alert box, it was more like a footnote, which is why I took it out of the alert box.

frontend/app/src/components/home/profile/UnlinkAccounts.svelte
ivan-jukic: I've made changes which I think should do the suggested, but please check.

julianjelfs (approved):

PR #7318, opened 2025-01-28
Restructure to handle autonomous bots

megrogan:


backend/canisters/community/impl/src/jobs/import_groups.rs
```diff
@@ -221,7 +221,8 @@ pub(crate) fn finalize_group_import(group_id: ChatId) {
             state.data.channels.add(Channel {
                 id: channel_id,
                 chat,
-                date_imported: None, // This is only set once everything is complete
+                date_imported: None,              // This is only set once everything is complete
+                bot_api_keys: Default::default(), // TODO: Should we import this?
```
hpeebles, line 225: The key will stop working because the scope will be wrong, so maybe we just say that the key gets reset when you import.

hpeebles (approved):

PR #7328, opened 2025-01-30
Complete send message promises early if message found in events response

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -3773,6 +3764,41 @@ export class OpenChat extends EventTarget {
                 },
             });
         });
+
+        // `searchPromise` is resolved either when the update call which initially sent the message completes, or if the
+        // message is found when reading new events
+        return searchPromise.then((resp) => {
+            if (resp.kind === "success" || resp.kind === "transfer_success") {
+                if (ledger !== undefined) {
+                    this.refreshAccountBalance(ledger);
```
ivan-jukic, line 3773: Just curious, in which cases would we get res.kind "success" and in which "transfer_success"? Not knowing the full context, it might appear  that a call to `refreshAccountBalance` is only required for transfer success...?

hpeebles: `transfer_success` happens in the cases where the message contains a crypto transfer, which is exactly the cases where `ledger` is defined.
So I could have done the check on `if (resp.kind === "transfer_success")` or on `if (ledger !== undefined)` since both are equivalent.
But the TS compiler doesn't know that... so I need to check `ledger` is defined otherwise the compiler will complain.

frontend/openchat-client/src/openchat.ts
```diff
@@ -3074,30 +3075,17 @@ export class OpenChat extends EventTarget {
...
 
         await this.#handleEventsResponse(serverChat, eventsResponse);
-        // We may have loaded messages which are more recent than what the chat summary thinks is the latest message,
-        // if so, we update the chat summary to show the correct latest message.
-        const latestMessage = findLast(eventsResponse.events, (e) => e.event.kind === "message");
-        const newLatestMessage =
-            latestMessage !== undefined && latestMessage.index > serverChat.latestEventIndex;
-
-        if (newLatestMessage) {
-            updateSummaryWithConfirmedMessage(
```
hpeebles, line 3088: This now takes place within `handleEventsResponse`, which makes more sense because it should happen whenever we load an event window or load new events

frontend/openchat-client/src/openchat.ts
ivan-jukic: Okay, gotcha, thanks!

frontend/openchat-shared/src/domain/chat/chat.ts
```diff
@@ -709,7 +709,6 @@ export type LocalChatSummaryUpdates = {
     updated?:
         | {
               kind?: undefined;
-              latestMessage?: EventWrapper<Message>;
```
hpeebles, line 712: We merge unconfirmed messages into the chat summary so this was never actually needed. Also, we were only setting it in one place which has now been updated to instead use `updateSummaryWithConfirmedMessage`, because it is called within `addServerEventsToStores` at which point the message is confirmed.

julianjelfs (approved):

PR #7332, opened 2025-01-30
Integrating API Keys

julianjelfs:
This is basically everything that you saw demo'd today. Should be safe to go. 

There is a bunch of stuff in the bots sdk repo too, but that's a bit of a mess at the moment so I want to spend a bit more time tidying it up.


hpeebles (approved):

hpeebles:
There are a few console logs that should most likely be removed

PR #7334, opened 2025-01-31
tone down the tick a bit and remove it from the left nav avatar

julianjelfs:
Following user feedback, the feeling was that the tick in the left nav was a bit in your face.


ivan-jukic (approved):
LGTM!

hpeebles (approved):

PR #7336, opened 2025-01-31
Temporarily remove `api_keys` fields until backend is released

hpeebles:


ivan-jukic (approved):
Optionally, I'd just leave a TODO  explaining when to uncomment that code.

hpeebles:
> Optionally, I'd just leave a TODO explaining when to uncomment that code.

Yeah true, but I think we're fine in this case because I'd imagine we release the website then immediately revert this PR, we can then release the backend changes next week and can continue dev'ing bots locally without having to use a forked version.

PR #7338, opened 2025-01-31
Wire up basic tracing

hpeebles:


hpeebles:
There may be some type wizardry we can do to avoid having to pass in the method name when using candid.
We have the `Actor` instance which is `Record<string, ActorMethod>`, and maybe we pass in the actor and the method name and somehow we tell the type system that the method name must be one of the keys in actor, and then the args must be the correct args for the `ActorMethod`.
I have no idea how to do that though...

julianjelfs (approved):

PR #7358, opened 2025-02-05
Store chat list view and land user on same view after swiping back

hpeebles:


frontend/app/src/components/home/ChatList.svelte
```diff
@@ -59,8 +59,8 @@
     let searchTerm: string = "";
     let searchResultsAvailable: boolean = false;
     let chatsScrollTop: number = 0;
-    let previousScope: ChatListScope | undefined = undefined;
-    let previousView: "chats" | "threads" = "chats";
+    let previousScope: ChatListScope | undefined = $chatListScope;
```
hpeebles, line 62: Rather than initialising these fields to the defaults we initialise them to the current store values. That way `onScopeChanged` and `onViewChanged` only trigger when there are changes after the component has rendered

julianjelfs (approved):

PR #7361, opened 2025-02-05
Use Vite for dev builds - now with HMR!

ivan-jukic:
This PR introduces Vite as our solution for running dev builds. It provides an "out-of-the-box" hot module replacing, which should help us build those new UI features much quicker.

There is quite a few changes in this PR, but the gist is that they're quite relevant for making us faster.

There is also a number of "fixes", that align the codebase with the latest syntax, and some improvements like the ability to alias imports, which might be easier to reconcile than relative paths.

---

[Screencast From 2025-02-03 19-49-06.webm](https://github.com/user-attachments/assets/b67435d2-bb49-4734-adb3-26088f2fffbb)

Video shows random changes and how they are immediately reflected in the browser when using Vite+HMR This would take about 3+ minutes to compile and preview with our current setup.


ivan-jukic:
Just a heads up that this PR should be ready for review/merge. One bit I'm not quite familiar with is our deploy to webtest/production process, and would just like to mention that we've prefixed names of the env variables used by the frontend with `OC_` due to Vite requirements - and we should make sure that any prod relevant vars are also updated.

julianjelfs (approved):
That was easy!

PR #7388, opened 2025-02-10
Remove nullish fields before validating bot responses

hpeebles:


frontend/openchat-agent/src/utils/nullish.ts
```diff
@@ -1,5 +1,5 @@
 export function deepRemoveNullishFields(value: unknown, depth = 0): unknown {
-    if (depth >= 100) throw new Error("Failed to remove null fields after 100 loops");
+    if (depth >= 100) throw new Error("Failed to remove nullish fields after 100 loops");
```
julianjelfs, line 2: I guess there is a _tiny_ risk here that something goes wrong downstream of this code because we are mutating the input by deleting fields rather than returning a copy with those fields deleted.

julianjelfs (approved):

PR #7398, opened 2025-02-12
Remove auth principal from user record after unlinking

hpeebles:


ivan-jukic (dismissed):
👍🏻

backend/canisters/identity/impl/src/model/user_principals.rs
```diff
@@ -42,6 +42,16 @@ struct AuthPrincipalInternal {
 }
 
 impl UserPrincipals {
+    pub fn remove_dangling_auth_principal_links(&mut self) {
+        for (index, user_principal) in self.user_principals.iter_mut().enumerate() {
+            user_principal.auth_principals.retain(|principal| {
```
ivan-jukic, line 50: This is a bit mind bending :), user_principals vs auth_principals, I wouldn't mind a comment just saying what we're doing here. Seems like we're removing any references to `user_principal.auth_principal`, if they've been removed from `self.auth_principals`, because unlinking will modify `auth_principals`, right?

hpeebles: Yeah good point. Have added a comment explaining the reasoning behind all of this.

ivan-jukic (approved):

PR #7411, opened 2025-02-13
Check for principal mismatch earlier to avoid making query

hpeebles:


backend/canisters/identity/impl/src/lib.rs
```diff
@@ -91,7 +91,7 @@ impl RuntimeState {
         }
 
         let (auth_principal, originating_canister) = if let Some(webauthn_key) = args.webauthn_key.as_ref() {
-            self.assert_key_not_generated_by_this_canister(&args.public_key);
+            self.assert_key_not_generated_by_this_canister(&webauthn_key.public_key);
```
hpeebles, line 94: This was checking the wrong public key

megrogan (approved):

PR #7423, opened 2025-02-14
make sure that user is there immediately after register

julianjelfs:


frontend/openchat-client/src/openchat.ts
```diff
@@ -936,6 +938,38 @@ export class OpenChat extends EventTarget {
         return this.#liveState.isDiamond ? DIAMOND_MAX_SIZES : FREE_MAX_SIZES;
     }
 
+    buildIdenticonUrl(userId: string): string {
```
hpeebles, line 941: Is this actually used?

julianjelfs: Uhm no - not sure what I was doing there.

megrogan (approved):

PR #7439, opened 2025-02-19
Support `sync_api_key` cmd + min required cmd role

megrogan:


julianjelfs (approved):
looks like roughly what I'm expecting!

PR #7451, opened 2025-02-20
Do not render previews if no LinkInfo is provided

ivan-jukic:
In case of the image links relayed from discord, fetching their metadata returns incomplete data, with only `title = ""`. This in turn indicates to the component that preview should be rendered since at least one thing is there since `"" !== undefined`.

This PR adds a check to make sure that something is there for the `LinkInfo`, and if it is, then render the previews fully or partially.

Before

After

### DIFF
--- a/frontend/app/src/components/home/GenericPreview.svelte
+++ b/frontend/app/src/components/home/GenericPreview.svelte
@@ -18,7 +18,7 @@
     export let rendered = false;
 
     let previewWrapper: HTMLElement;
-    let previewPromise: Promise<LinkInfo | null> | undefined = undefined;
+    let previewPromise: Promise<LinkInfo | undefined> | undefined = undefined;
 
     $: {
         if (intersecting && !$eventListScrolling && !rendered && !$offlineStore) {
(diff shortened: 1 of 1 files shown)


hpeebles (dismissed):

frontend/app/src/components/home/GenericPreview.svelte
```diff
@@ -18,49 +18,45 @@
...
         }
     }
 
-    async function loadPreview(url: string): Promise<LinkInfo> {
+    async function loadPreview(url: string): Promise<LinkInfo | null> {
         const response = await fetch(
             `${import.meta.env.OC_PREVIEW_PROXY_URL}/preview?url=${encodeURIComponent(url)}`,
         );
+
+        const check_if_meta_empty = (meta: Omit<LinkInfo, "url">) =>
```
julianjelfs, line 41: weird casing - been doing to much rust!

frontend/app/src/components/home/GenericPreview.svelte
```diff
@@ -18,49 +18,45 @@
...
+                if (preview && intersecting && !$eventListScrolling) {
+                    rendered = true;
+                    dispatch("rendered", url);
                 }
             });
         }
     }
 
-    async function loadPreview(url: string): Promise<LinkInfo> {
+    async function loadPreview(url: string): Promise<LinkInfo | null> {
```
julianjelfs, line 36: can we use `undefined` wherever we have control? I _know_ it is fussy but relying on falsiness has burned me _so_ many times that I just prefer to _never_ do it.

frontend/app/src/components/home/GenericPreview.svelte
ivan-jukic: yup

frontend/app/src/components/home/GenericPreview.svelte
ivan-jukic: sure

frontend/app/src/components/home/GenericPreview.svelte
hpeebles: > been doing to much rust!

No such thing!

julianjelfs (approved):

PR #7473, opened 2025-02-25
Avoid updating `globalStateStore` to update the latest message in a chat

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -2875,11 +2874,6 @@ export class OpenChat extends EventTarget {
 
         this.#addServerEventsToStores(chatId, resp.events, threadRootMessageIndex, []);
 
-        for (const event of resp.events) {
```
hpeebles, line 2878: As of a month or so ago this is now handled within `addServerEventsToStores`

megrogan (approved):

PR #7479, opened 2025-02-26
Register bot with test location

julianjelfs:


megrogan (approved):

frontend/app/src/i18n/en.json
```diff
@@ -253,7 +253,9 @@
             "principalRules": "This will be the canisterID for an onchain bot",
             "required": "Required parameter",
             "requiredDesc": "Is this a required parameter for the command",
-            "title": "Build a bot",
+            "testContext": "Test context",
+            "testContextInfo": "Select a test context to register this bot with",
```
hpeebles, line 257: Maybe "Select a Group/Community in which you'll be able to test this bot"?

hpeebles: Or something similar to that... as it stands I think it isn't clear what exactly it is you're selecting

julianjelfs: Meh - I think I prefer mine. Also it's not just Group/Community. Also direct chats.

hpeebles: Fair enough

hpeebles: Actually... this doesn't actually register the bot there. It just allows you to register it there.

hpeebles: Ah also you wouldn't be selecting a direct chat here, only the owner can test via direct chat

hpeebles: Regardless, I'm fine with leaving it as is

hpeebles (approved):

frontend/app/src/i18n/en.json
julianjelfs: But actually you're right - I shouldn't be including direct chats here. Will fix that.

hpeebles (dismissed):

hpeebles (approved):

PR #7489, opened 2025-02-27
On `user::c2c_install_bot` create a direct chat

megrogan:


backend/canisters/user/impl/src/updates/c2c_install_bot.rs
```diff
@@ -27,5 +29,13 @@ fn c2c_install_bot_impl(args: Args, state: &mut RuntimeState) -> Response {
         return AlreadyAdded;
     }
 
+    // If there isn't already a direct chat with the bot, create one now
+    if state.data.direct_chats.get(&args.bot_id.into()).is_some() {
```
hpeebles, line 33: `is_none`

PR #7518, opened 2025-03-04
Only show search options when search box focused

hpeebles:


frontend/app/src/components/bots/InstallationLocationSelector.svelte
```diff
@@ -81,6 +82,14 @@
             }
         }
     }
+
+    function onFocus() {
+        focused = true;
+    }
+
+    function onBlur() {
+        window.setTimeout(() => focused = false, 100);
```
hpeebles, line 91: This is a bit hacky, but without it the search results are hidden before they are selected and so it becomes impossible to select anything. Open to nicer ways of handling this!

megrogan (approved):

frontend/app/src/components/bots/InstallationLocationSelector.svelte
ivan-jukic: Ah yes, there's a way to do this, you basically need to stop blur event propagation when user selects an item, which then keeps the select open long enough for the selection to happen, and then you mutate `focused` as a part of the selection, or call blur method on the input field, whichever works.

I've done it a few times, don't remember all the details, there might be some other nuances with this, for example I think that the input and select dropdown might need the same parent at which level you actually stop the event propagation.

Also, I've not checked the Select component code, I'm basically assuming that this is the problem here.

PR #7546, opened 2025-03-07
allow pow-3.org to integrate oc xframe

yuquanwang:


hpeebles:
You also need to update this file - https://github.com/open-chat-labs/open-chat/blob/master/frontend/app/src/stores/xframe.ts

hpeebles (approved):

PR #7577, opened 2025-03-11
Added new DateTime component

ivan-jukic:
Replaced `svelte-datetime-picker` with `svelty-picker` component due to better support for picking time.

Time picker is equivalent in design and functionality to the material design version of it, and should work well for desktop and mobile users.

### DIFF
--- a/frontend/app/package.json
+++ b/frontend/app/package.json
@@ -55,6 +55,7 @@
         "sass": "^1.65.1",
         "stream-browserify": "3.0.0",
         "svelte-preprocess": "^6.0.3",
+        "svelty-picker": "^6.1.5",
         "tslib": "^2.6.1",
         "typescript": "=5.4.2",
         "url": "^0.11.1",
--- a/frontend/app/src/components/DateInput.svelte
+++ b/frontend/app/src/components/DateInput.svelte
@@ -1,17 +1,25 @@
 <script module lang="ts">
     export interface Props {
         align?: string;
-        closeOnSelection?: boolean;
+        clearBtn?: boolean;
+        clearToggle?: boolean;
         disabled?: boolean;
+        disableDatesFn?: (date: Date) => boolean;
+        endDate?: string | Date;
         format?: string;
         futureOnly?: boolean;
+        inputId?: string;
+        inputClasses?: string;
+        minuteIncrement?: number;
+        mode?: "date" | "time" | "datetime";
+        pickerOnly?: boolean;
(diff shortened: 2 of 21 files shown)


frontend/app/package.json
```diff
@@ -55,6 +55,7 @@
         "sass": "^1.65.1",
         "stream-browserify": "3.0.0",
         "svelte-preprocess": "^6.0.3",
+        "svelty-picker": "^6.1.5",
```
megrogan, line 58: Should the previous date picker be removed?

frontend/app/src/components/DateInput.svelte
```diff
@@ -20,57 +28,83 @@
...
     }: Props = $props();
 
     // If we're only allowed to select future dates, then minimum date and
     // time selectable is current!
-    let minDate: Date | undefined = futureOnly ? new Date() : undefined;
-
-    // Allow selection of up to 10 years in the future.
-    let maxDate = new Date();
-    maxDate.setFullYear(maxDate.getFullYear() + 10);
+    startDate = futureOnly ? new Date() : undefined;
```
megrogan, line 59: I don't think we should have this complexity. I think we should just allow any date/time to be picked but then show a validation error next to the text input if it is in the future

frontend/app/package.json
ivan-jukic: Yes, and I did that, but I might have returned it when I was resolving a conflict with the current master. Just updated! It was in the frontend/packages.json, no need for the new one to be there, since only the app is using it.

hpeebles (dismissed):

frontend/app/src/components/DateInput.svelte
```diff
@@ -1,76 +1,117 @@
 <script module lang="ts">
     export interface Props {
         align?: string;
-        closeOnSelection?: boolean;
+        clearBtn?: boolean;
```
julianjelfs, line 4: Do we actually _need_ to expose all of these props from the component. Might be better to only expose the things that need to be exposed when we have multiple use cases that need to be configured making it easier to see what is and isn't being used.

ivan-jukic: No we don't. I've already reduced this set from all the possible options. What do you think might be worth keeping/removing?

ivan-jukic: I've reduced the props size!

julianjelfs (approved):

PR #7607, opened 2025-03-17
Support bot `direct message` scenario

megrogan:


backend/canisters/user/impl/src/model/direct_chats.rs
```diff
@@ -26,18 +26,16 @@ impl DirectChats {
         self.direct_chats.get_mut(chat_id)
     }
 
-    pub fn create(
+    pub fn get_or_create<F: FnOnce() -> u128>(
         &mut self,
         their_user_id: UserId,
         their_user_type: UserType,
-        anonymized_id: u128,
+        anonymized_id: LazyCell<u128, F>,
```
hpeebles, line 33: Just pass in the function rather than the function wrapped in a `LazyCell`

backend/canisters/user/impl/src/updates/send_message.rs
```diff
@@ -10,10 +10,13 @@ use canister_api_macros::update;
 use canister_tracing_macros::trace;
 use chat_events::EditMessageArgs;
 use chat_events::EditMessageResult;
+use chat_events::TextContentInternal;
 use chat_events::{MessageContentInternal, PushMessageArgs, Reader, ReplyContextInternal, ValidateNewMessageContentResult};
 use constants::{MEMO_MESSAGE, OPENCHAT_BOT_USER_ID};
+use event_store_producer::NullRuntime;
 use event_store_producer_cdk_runtime::CdkRuntime;
 use rand::Rng;
+use std::ops::Not;
```
hpeebles, line 19: This is probably not meant to be here

megrogan: It is used on line 361

backend/canisters/user/impl/src/updates/send_message.rs
```diff
@@ -298,6 +301,44 @@ fn c2c_bot_send_message(args: c2c_bot_send_message::Args) -> c2c_bot_send_messag
...
+            if let (Some(text), Some(message_id)) = (
+                command.args.first().and_then(|a| a.value.as_string().map(String::from)),
+                user_message_id,
+            ) {
+                let chat = state
+                    .data
+                    .direct_chats
+                    .get_or_create(bot_id, UserType::BotV2, || state.env.rng().gen(), now);
+
+                chat.push_message::<NullRuntime>(
```
hpeebles, line 318: We should probably record the event rather than passing in `event_client` as `None`.

hpeebles: Or did you choose `None` because we already record the bot's reply?

megrogan: Debatable

megrogan: We wuldn't be recording this as an event if the same command was summoned with a slash.

megrogan: > Or did you choose `None` because we already record the bot's reply?

Yes

backend/canisters/user/impl/src/updates/send_message.rs
hpeebles: Ahh right it enables the `.not()` method.

hpeebles (approved):

PR #7611, opened 2025-03-18
Update BTC balance via User canister and send message once updated

hpeebles:


backend/canisters/user/impl/src/updates/update_btc_balance.rs
```diff
@@ -0,0 +1,72 @@
...
+use tracing::error;
+use user_canister::update_btc_balance::{Response::*, *};
+
+#[update(msgpack = true)]
+#[trace]
+async fn update_btc_balance(_args: Args) -> Response {
+    run_regular_jobs();
+
+    let test_mode = read_state(|state| state.data.test_mode);
+    let ckbtc_minter_canister_id = if test_mode { CKBTC_MINTER_CANISTER_ID } else { TESTNET_CKBTC_MINTER_CANISTER_ID };
```
ivan-jukic, line 16: Is this correct, if test mode true then CK btc minter canister id, otherwise if not test use testnet ckbtc minter??

hpeebles: Doh! Good catch!

backend/canisters/user/impl/src/updates/update_btc_balance.rs
```diff
@@ -0,0 +1,72 @@
...
+                    UtxoStatus::Minted(m) => total_minted += m.minted_amount,
+                    error => errors.push(format!("{error:?}")),
+                }
+            }
+
+            mutate_state(|state| {
+                if total_minted > 0 {
+                    let formatted = format_crypto_amount(total_minted as u128, 8);
+                    crate::openchat_bot::send_text_message(
+                        format!(
```
ivan-jukic, line 38: Can we use an i18n key here, then insert the message on the frontend?

backend/canisters/user/impl/src/updates/update_btc_balance.rs
```diff
@@ -0,0 +1,72 @@
...
+Your account has been credited with {formatted} BTC."
+                        ),
+                        Vec::new(),
+                        false,
+                        state,
+                    );
+                }
+                for error in errors {
+                    crate::openchat_bot::send_text_message(
+                        format!(
```
ivan-jukic, line 49: Same question really.

backend/canisters/user/impl/src/updates/update_btc_balance.rs
```diff
@@ -0,0 +1,72 @@
...
+    let test_mode = read_state(|state| state.data.test_mode);
+    let ckbtc_minter_canister_id = if test_mode { CKBTC_MINTER_CANISTER_ID } else { TESTNET_CKBTC_MINTER_CANISTER_ID };
+
+    match ckbtc_minter_canister_c2c_client::update_balance(
+        ckbtc_minter_canister_id,
+        &ckbtc_minter_canister::update_balance::Args::default(),
+    )
+    .await
+    {
+        Ok(Ok(utxos)) => {
```
ivan-jukic, line 24: A result of a result? Perhaps it needs to be flattened upstream?

ivan-jukic (approved):

PR #7615, opened 2025-03-18
support direct message mode

julianjelfs:
And a few other small fixes.


frontend/openchat-shared/src/domain/bots.ts
```diff
@@ -770,3 +779,24 @@ export function botActionScopeFromInstallLocation(
             };
     }
 }
+
+export function directMessageCommandInstance(
+    bot: ExternalBot,
+    text: string,
+): BotCommandInstance | undefined {
+    const command = bot.definition.commands.find((c) => {
+        return c.directMessages && c.params[0]?.kind === "string";
```
megrogan, line 788: As it stands this would apply even if the command has other required params which doesn't seem right

PR #7617, opened 2025-03-18
Implement ability to generate BTC deposit address

hpeebles:


frontend/app/src/components/home/profile/ReceiveCrypto.svelte
```diff
@@ -44,7 +46,11 @@
             on:error={onBalanceRefreshError} />
     </span>
     <form class="body" slot="body">
-        <AccountInfo qrSize={"larger"} centered {ledger} user={$user} />
+        {#if symbol === BTC_SYMBOL}
```
hpeebles, line 49: Currently, the symbol will never be BTC, because we're still using ckBTC.
So this change is completely safe to get in.
I'll make a subsequent change that switches ckBTC to BTC and then we can test everything end to end on prod test.

megrogan (approved):

PR #7630, opened 2025-03-19
Bump CDK to 0.18.0-alpha.2

hpeebles:


hpeebles:
The tests are failing because the PocketIC version we are using doesn't include some of the latest system functions that the new version of the CDK is now attempting to make use of.

megrogan (approved):

PR #7631, opened 2025-03-20
add workflow to trigger github bot

julianjelfs:


.github/workflows/pr_created.yaml
```diff
@@ -0,0 +1,36 @@
+name: PR Created
+
+on:
+  pull_request:
+    types: [opened, reopened]
```
megrogan, line 5: Will this trigger for sdraft PRs? Ideally draft PRs would be ignored and this would only trigger when the PR is made non-draft.

julianjelfs: this will not include drafts as it stands. 

`[opened, reopened, ready_for_review]` would include them

ivan-jukic (approved):

PR #7635, opened 2025-03-20
Nightly build OC docker image and push

ivan-jukic:
Only builds a new image if there were new commits to master in the last 24 hours.

### DIFF
--- a/.github/workflows/docker_release.yaml
+++ b/.github/workflows/docker_release.yaml
@@ -30,6 +30,9 @@ jobs:
     if: needs.check-merges.outputs.merged == 'true'
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
       - name: Log in to Docker Hub
         uses: docker/login-action@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/docker_release.yaml
```diff
@@ -0,0 +1,45 @@
...
+          else
+            echo "merged=false" >> "$GITHUB_OUTPUT"
+          fi
+
+  build-and-push:
+    needs: check-merges
+    if: needs.check-merges.outputs.merged == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
```
julianjelfs, line 33: Does this need to be done again?

ivan-jukic: If it runs on a schedule, would it have the checkout from before? Do we want to make sure we get the latest checked out?

julianjelfs: But doesn't it _always_ run check-merge which itself checks out the latest?

julianjelfs (approved):

.github/workflows/docker_release.yaml
julianjelfs: Sounds plausible. Might at well be sure.

megrogan (approved):

PR #7639, opened 2025-03-21
Avoid playing incoming video call ringtone if already on another call

hpeebles:


hpeebles:
> I'm not sure we don't want the existing behaviour

With this version the incoming call pop up will still appear, it just won't ring, which I think makes sense.
A new call shouldn't start ringing over your current one.

megrogan (approved):

PR #7647, opened 2025-03-24
Update canisters post release

hpeebles:


backend/canisters/user/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [unreleased]
 
+## [[2.0.1663-user](https://github.com/open-chat-labs/open-chat/releases/tag/v2.0.1663-user)] - 2025-03-21
```
megrogan, line 9: Should the user changelog be changing yet seeing as we've only upgraded a fraction of the canisters?

hpeebles: I think it still makes sense to say what went into that release. I purposely didn't remove the code that we could have removed had we upgraded all of them.

megrogan (approved):

PR #7666, opened 2025-03-25
Fix userId that test ICP gets sent to

hpeebles:


scripts/deploy-local.sh
```diff
@@ -82,7 +82,7 @@ echo "Canisters created"
     $NNS_INDEX_CANISTER_ID \
     true \
 
-./scripts/get-test-icp.sh "dccg7-xmaaa-aaaaa-qaamq-cai" $IDENTITY
+./scripts/get-test-icp.sh "w7lou-c7777-77774-qaamq-cai" $IDENTITY
 ./scripts/deploy-test-ledger.sh $IDENTITY
-./scripts/mint-test-tokens.sh "dccg7-xmaaa-aaaaa-qaamq-cai" $IDENTITY
+./scripts/mint-test-tokens.sh "w7lou-c7777-77774-qaamq-cai" $IDENTITY
```
megrogan, line 87: Seems rather fragile

megrogan (approved):

scripts/deploy-local.sh
hpeebles: Yeah it definitely is. Not sure how we can make it better though.

PR #7674, opened 2025-03-26
require the bot rotation challenge for suspect users

julianjelfs:


frontend/app/src/components/RotationChallenge.svelte
```diff
@@ -0,0 +1,191 @@
...
+    interface Props {
+        onClose: () => void;
+        onResult: (e: MouseEvent, success: boolean) => void;
+        imagePath?: string;
+    }
+
+    let { onClose, onResult, imagePath = "/assets/robot.svg" }: Props = $props();
+
+    let leftImage: HTMLImageElement;
+    let targetRotation: number = $state(0);
```
hpeebles, line 22: Should this just be `const targetRotation = normaliseAngle(Math.random() * 2 * Math.PI)`?
Because it should never change, right?

frontend/app/src/components/RotationChallenge.svelte
```diff
@@ -0,0 +1,191 @@
...
+    let userRotation: number = $state(0);
+    let leftImageTransform = $derived(`rotate(${userRotation}rad)`);
+    let rightImageTransform = $derived(`rotate(${targetRotation}rad)`);
```
hpeebles, line 25: Same here, this can just be `const ...`

hpeebles (approved):

hpeebles (approved):

PR #7685, opened 2025-03-27
add a simple bit of pub sub - how does it feel?

julianjelfs:
I think this could (if used sparingly) remove quite a bit of annoying wiring.

### DIFF
--- a/frontend/app/src/components/home/video/ActiveCallSummary.svelte
+++ b/frontend/app/src/components/home/video/ActiveCallSummary.svelte
@@ -19,7 +19,7 @@
         hasPresence,
     } from "../../../stores/video";
     import page from "page";
-    import { createEventDispatcher, getContext } from "svelte";
+    import { getContext } from "svelte";
     import Avatar from "../../Avatar.svelte";
     import FancyLoader from "../../icons/FancyLoader.svelte";
     import Microphone from "svelte-material-icons/Microphone.svelte";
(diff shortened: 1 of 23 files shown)


frontend/app/src/components/home/video/ActiveCallSummary.svelte
```diff
@@ -115,7 +115,7 @@
     }
 
     function hangup() {
-        dispatch("hangup");
+        publish("hangup", undefined);
```
ivan-jukic, line 118: Perhaps payload is optional?

julianjelfs: Yes! I was just wrestling with how to make this optional _only_ when it should be. And I have just got that to work I think.

frontend/app/src/utils/pubsub.ts
```diff
@@ -0,0 +1,45 @@
...
+            if (subscriptions[name]!.length === 0) {
+                delete subscriptions[name];
+            }
+        }
+    };
+}
+
+export function publish<K extends NoPayloadEvents>(name: K): void;
+export function publish<K extends keyof Events>(name: K, payload: Events[K]): void;
+export function publish<K extends keyof Events>(name: K, payload?: Events[K]): void {
```
ivan-jukic, line 43: Huh, I would've imagined that adding `?` would suffice 😮

julianjelfs: That would be enough to make it unconditionally optional but that would allow us to make mistakes. I want it to be required when it's required and not allowed when it's not required. 

e.g. now this is now _not_ allowed: 

`publish("startVideoCall")`

but this _is_: 

`publish("hangup")`

ivan-jukic (approved):

PR #7689, opened 2025-03-28
Avoid retrying queries if we run into a Typebox validation error

hpeebles:


frontend/openchat-agent/src/services/canisterAgent/msgpack.ts
```diff
@@ -189,16 +188,7 @@ export abstract class MsgpackCanisterAgent extends CanisterAgent {
         validator: Resp,
     ): Out {
         const response = deserializeFromMsgPack(new Uint8Array(responseBytes));
-        try {
```
hpeebles, line 192: I moved this logging into `typebox.ts` so it covers requests and responses

julianjelfs (approved):

PR #7700, opened 2025-03-29
Review $effects

julianjelfs:
Specifically the Markdown.svelte component is probably causing some significant thrashing on prod at the moment. 

And refactor to hide and show main panels rather than re-render.


frontend/app/src/components/home/ChatList.svelte
```diff
@@ -164,6 +167,7 @@
     function onViewChanged() {
         previousView = $chatListView;
         const scrollTop = previousView === "chats" ? chatsScrollTop ?? 0 : 0;
+        console.log("Scroll stuff: ", chatsScrollTop);
```
hpeebles, line 170: Do you want this to be kept in here?

julianjelfs: Nope - gone

hpeebles (approved):

PR #7713, opened 2025-04-01
Streak insurance

julianjelfs:
image

image


frontend/app/src/i18n/en.json
```diff
@@ -1674,6 +1676,22 @@
     "stopRecording": "Stop recording",
     "storage": "Storage",
     "storageUsed": "{used}GB used of {limit}GB",
+    "streakInsurance": {
+        "addDay": "Add a day",
+        "atRisk": "Your streak is at risk",
+        "bought": "days bought",
```
hpeebles, line 1682: Maybe "days insured"? Especially if we want to ever give out free days such as when they hit 100 or 365 or whenever.

PR #7721, opened 2025-04-02
fix reactivity of reactions

julianjelfs:


frontend/openchat-client/src/utils/chat.ts
```diff
@@ -1627,7 +1627,7 @@ function mergeLocalUpdates(
     }
 
     if (localUpdates?.reactions !== undefined) {
-        let reactions = [...message.reactions];
+        let reactions = message.reactions.map((r) => ({ ...r }));
```
julianjelfs, line 1630: this might not be 100% the _best_ solution but it works. I'm going to continue to look into whether we should be using `SvelteSet` and whether there is a nice generic way to do that but I'd rather get this out first just to address the problem.

hpeebles (dismissed):

hpeebles (approved):

PR #7739, opened 2025-04-04
Avoid making duplicate calls to refresh account balances

hpeebles:


frontend/app/src/components/home/BalanceWithRefresh.svelte
```diff
@@ -29,12 +29,6 @@
             ? client.formatTokens(value, tokenDetails.decimals)
             : convertValue(conversion, tokenDetails);
 
-    $: {
-        if (ledger) {
-            refresh();
```
hpeebles, line 34: I'm really not sure why this causes refresh to run multiple times given that the component's key is `ledger`... but it does.
With the call to `refresh` outside of this reactive statement the refresh function only ever triggers once.

frontend/app/src/components/home/BalanceWithRefresh.svelte
```diff
@@ -70,6 +64,8 @@
                 return t.ethBalance?.toFixed(6) ?? "???";
         }
     }
+
+    refresh();
```
julianjelfs, line 68: We _probably_ use this component in scenarios where the user can select the token they are using. I think that's where we would want the refresh call to react to the change in the value of ledger. I'm a bit worried that this change might break that.

julianjelfs (approved):

PR #7764, opened 2025-04-09
Fix serializing empty MessagePack responses

hpeebles:


backend/libraries/canister_api_macros/src/lib.rs
```diff
@@ -72,24 +72,26 @@ fn canister_api_method(method_type: MethodType, attr: TokenStream, item: TokenSt
...
+        };
+
         let serializer_name = format!("{msgpack_name}_serializer");
         let serializer_ident = Ident::new(&serializer_name, Span::call_site());
 
         let deserializer_name = format!("{msgpack_name}_deserializer");
         let deserializer_ident = Ident::new(&deserializer_name, Span::call_site());
 
-        let encode_with = if empty_return {
-            quote! {}
```
hpeebles, line 82: By not specifying `encode_with`, it falls back to the default which is to use candid. And candid encodes unit types with a prefix that states that the encoded value is a unit, whereas MessagePack and JSON return empty bytes.

ivan-jukic (approved):
LGTM!

PR #7775, opened 2025-04-10
Stop sending requests to uninstalled ledgers and remove from frontend

hpeebles:


backend/canisters/registry/api/src/lib.rs
```diff
@@ -32,6 +32,8 @@ pub struct TokenDetails {
     pub enabled: bool,
     pub last_updated: TimestampMillis,
     pub payments: Vec<Payment>,
+    #[ts(skip)]
+    pub uninstalled: bool,
```
hpeebles, line 36: Tokens which are uninstalled won't be sent to the frontend, so the frontend can ignore this field, hence the `skip` attribute.

julianjelfs (dismissed):

julianjelfs (approved):

PR #7809, opened 2025-04-17
simplify scrolling to message on current chat

julianjelfs:
Chipping away a bit more at the routeChange function in the Home component. I think I see a path to removing all to wiring around scrolling to a specific message from the home component. This is step one.

### DIFF
--- a/frontend/app/src/components/home/CurrentChatMessages.svelte
+++ b/frontend/app/src/components/home/CurrentChatMessages.svelte
@@ -1,40 +1,42 @@
 <script lang="ts">
-    import { getContext } from "svelte";
-    import ChatEvent from "./ChatEvent.svelte";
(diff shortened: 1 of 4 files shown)


frontend/app/src/components/home/CurrentChatMessages.svelte
```diff
@@ -266,6 +268,19 @@
             groupInner(filteredProposals),
         ),
     );
+
+    // if the messageIndex has changed but the chatId has not, scroll to the specified message
+    let previousChatId: ChatIdentifier | undefined = undefined;
+    $effect(() => {
+        if (
```
hpeebles, line 275: So is Svelte clever enough to only trigger this when `pathState.messageIndex` or `app.selectedChatId` change?
As opposed to when `pathState` or `app` are updated?

PR #7825, opened 2025-04-22
Group webhook endpoints

megrogan:


backend/canisters/group/impl/src/updates/delete_webhook.rs
```diff
@@ -0,0 +1,34 @@
...
+fn delete_webhook(args: Args) -> Response {
+    run_regular_jobs();
+
+    mutate_state(|state| delete_webhook_impl(args, state)).into()
+}
+
+fn delete_webhook_impl(args: Args, state: &mut RuntimeState) -> OCResult {
+    state.data.verify_not_frozen()?;
+
+    let member = state
```
hpeebles, line 19: You should use `state.get_calling_member(..)`

hpeebles (approved):

PR #7826, opened 2025-04-22
Proxy fails to load images on localhost

av1ctor:
When returning an image inside a markdown text message on **localhost**, the preview proxy will fail and the image plus the text message won't be rendered.


julianjelfs:
> When returning an image inside a markdown text message on **localhost**, the preview proxy will fail and the image plus the text message won't be rendered.

Could you give an example of a message that causes a problem with the current code (including what you expect to happen and what actually happens). I'm just want to make sure that I understand the problem this is seeking to solve.

av1ctor:
Hi, @julianjelfs, sure.

I'm developing a chat bot in Rust using the Open Chat SDK.

If I include a link to a local host image in a text message markdown response like ```NFT 1234 minted!\n  ```, the proxy will fail because it's a localhost address and not even the "NFT 1234 minted!" text message will be rendered, making it impossible to test the bot.

I reported it at https://oc.app/community/fpl34-aiaaa-aaaac-aiboa-cai/channel/657697207/113

julianjelfs:
Sorry for the delay looking at this - got a few issues to fix before I can have a proper look but it's definitely on my list!

julianjelfs:
closing this as it doesn't seem to be an issue any more. Please re-open if this is not the case.

PR #7863, opened 2025-04-29
Webhooks in frontend

megrogan:


frontend/app/src/utils/user.ts
```diff
@@ -18,3 +19,34 @@ export function buildDisplayName(
 export function trimLeadingAtSymbol(term: string): string {
     return term.length > 0 && term[0] === "@" ? term.substring(1) : term;
 }
+
+export function findSender(
+    senderId: string,
+    users: UserLookup,
+    webhooks: WebhookDetails[],
```
julianjelfs, line 26: Given that the value of this param is _always_ app.selectedChat.webhooks, you could just reference that directly. There is not actually any need (from a reactivity point of view) to pass it in. Not really sure how I feel about that since a large part of me still thinks - yeah but pure functions are just better.

frontend/openchat-client/src/state/chat_details/server.svelte.ts
```diff
@@ -26,6 +27,7 @@ export class ChatDetailsServerState {
     #rules = $state<VersionedRules>(emptyRules());
     #bots = $state<ReadonlyMap<string, ExternalBotPermissions>>(new Map());
     #apiKeys = $state<ReadonlyMap<string, PublicApiKeyDetails>>(new Map());
+    #webhooks = $state<WebhookDetails[]>([]);
```
julianjelfs, line 30: Given that these are keyed we might want to consider storing them in a map so we can easily look them up. Doesn't make them any harder to iterate really.

frontend/app/src/utils/user.ts
megrogan: I'll leave it as is

julianjelfs (approved):

PR #7900, opened 2025-05-04
Fix case where new achievement pop-up would be shown repeatedly

hpeebles:
Something like this could work... completely untested so far though


julianjelfs:
I would quite like to understand why this is a problem in the first place though. It doesn't seem to be happening for me.

PR #7916, opened 2025-05-07
Update canisters post release

hpeebles:


megrogan:
We should remove bot_context from Message and MessageInternal

megrogan (approved):

PR #7923, opened 2025-05-07
Group commands by bot when there is no filter

megrogan:


julianjelfs (commented):
It seems weird to _only_ group by bot when there is no prefix. What's the rationale for that? Aren't they going to suddenly jump around rather than just filter down?

PR #7952, opened 2025-05-12
Fix transfer amount when submitting proposal

hpeebles:
This fixes the transfer fees specified when submitting a proposal.
They should be as follows:

1 fee for the approval + 1 fee for the transfer to the ProposalsBot + 1 fee for the transfer to the neuron if required.
So 3 fees in total.

### DIFF
--- a/frontend/app/src/components/home/proposal/MakeProposalModal.svelte
+++ b/frontend/app/src/components/home/proposal/MakeProposalModal.svelte
@@ -150,7 +150,7 @@
         busy = true;
 
         if (
-            !(await approvePayment(PROPOSALS_BOT_CANISTER, proposalCost + BigInt(2) * transferFee))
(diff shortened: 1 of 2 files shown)


frontend/app/src/components/home/proposal/MakeProposalModal.svelte
```diff
@@ -150,7 +150,7 @@
         busy = true;
 
         if (
-            !(await approvePayment(PROPOSALS_BOT_CANISTER, proposalCost + BigInt(2) * transferFee))
+            !(await approvePayment(PROPOSALS_BOT_CANISTER, proposalCost + transferFee))
```
hpeebles, line 153: We only need to approve the proposal cost + a single fee, but then a fee gets charged on top of that for the approval and another fee gets charged again for the transfer.
So 3 fees are needed in total.

megrogan (approved):

PR #7962, opened 2025-05-14
365 celebration

julianjelfs:
https://github.com/user-attachments/assets/e86ed964-c65b-4d21-9fbb-cca637d9270d


frontend/app/src/components/Avatar.svelte
```diff
@@ -1,15 +1,15 @@
...
-    import { getContext } from "svelte";
     import { now } from "../stores/time";
-    import Verified from "./icons/Verified.svelte";
-    import { trackedEffect } from "@src/utils/effects.svelte";
+    import LighteningBolt from "./home/nav/LighteningBolt.svelte";
 
     const client = getContext<OpenChat>("client");
 
-    type Specialisation = "verified" | "bot" | "none";
+    type Specialisation = "bot" | "none" | "super_user";
```
hpeebles, line 12: Maybe rather than "super_user" we say "max_streak"?
I think that will result in less confusion when people see this value.

frontend/app/src/components/Avatar.svelte
```diff
@@ -19,8 +19,8 @@
         blocked?: boolean;
         statusBorder?: string;
         selected?: boolean;
-        verified?: boolean;
         bot?: boolean;
+        superUser?: boolean;
```
hpeebles, line 23: This should be `maxStreak`

PR #7965, opened 2025-05-14
Fix cases where requests were being made to canisters while offline

hpeebles:


frontend/openchat-client/src/state/unread/markRead.svelte.ts
```diff
@@ -148,6 +152,7 @@ export class MessageReadTracker {
 
     stop(): void {
         console.log("stopping the mark read poller");
+        this.#stopped = true;
         if (this.#timeout !== undefined) {
             window.clearTimeout(this.#timeout);
```
hpeebles, line 157: This was clearing the timeout, but if the async function to flush data to the backend was already in progress, then this would have no effect and a new timeout would be created upon completion of that async task.

julianjelfs (approved):

PR #8028, opened 2025-05-22
Maintain subscription after calling `value`

hpeebles:


frontend/openchat-client/src/utils/stores.ts
```diff
@@ -206,10 +221,8 @@ class _Derived<S extends Stores, T> {
     }
 
     get value(): T {
-        const unsub = this.#started ? undefined : this.subscribe(NOOP);
-        const value = this.#innerStore.value;
-        unsub?.();
-        return value;
+        this.#valueSubscriber ??= this.subscribe(NOOP);
```
julianjelfs, line 221: Don't we want to only create a subscriber if there are no subscribers of any kind rather than if there is no value subscriber. Seems like in a lot of cases there would already be subscribers so there would be no need to create this value subscriber.

hpeebles: Doing it this way just makes it consistent, otherwise you may call `value` on an already started store, then the store's subscriber unsubscribes, then next time you call `value` it'll have to run startup again.
Also, adding a subscriber is extremely cheap if the start up process is already complete.

PR #8038, opened 2025-05-23
Separate webhook userIds from regular userIds

hpeebles:


julianjelfs (commented):
I think it would be better to defer this and add it on top of my branch as it's using things that don't exist any more.

julianjelfs (approved):

PR #8047, opened 2025-05-27
Deduplicate undo functions of local updates

hpeebles:


frontend/openchat-client/src/state/localUpdates.ts
```diff
@@ -726,11 +729,10 @@ export class GlobalLocalState {
 
     updateChatRulesAccepted(id: ChatIdentifier, rulesAccepted: boolean): UndoLocalUpdate {
         return this.#modifyChatSummaryUpdates(id, (upd) => {
-            const prev = upd.rulesAccepted;
             upd.rulesAccepted = rulesAccepted;
             return (upd) => ({
                 ...upd,
```
julianjelfs, line 734: come to think of it, I don't think there is any reason for this spread. Given that the store itself has the `notEq` equality checker, we _could_ just do: 

```
return (upd) => {
    upd.rulesAccepted = undefined;
    return upd;
}
```

Not a big deal but might as well avoid the allocation.

julianjelfs (approved):

PR #8068, opened 2025-05-29
include application/json header

julianjelfs:


backend/notification_pusher/core/src/bot_notifications/pusher.rs
```diff
@@ -55,7 +56,13 @@ impl Pusher {
     async fn push_notification(&self, payload: Vec<u8>, endpoint: String) -> Result<(), String> {
         let mut url = Url::parse(&endpoint).map_err(|e| e.to_string())?;
         url = url.join("notify").map_err(|e| e.to_string())?;
-        self.http_client.post(url).body(payload).send().await.unwrap();
+        self.http_client
+            .post(url)
+            .header(CONTENT_TYPE, "application/json")
```
hpeebles, line 61: The payload isn't always JSON, bots can choose to get JSON or Candid, with JSON being the default.
So we should probably pass in the content type header value to cater for both cases.

PR #8082, opened 2025-06-02
Check for existence, assign & map FCM tokens to users

ivan-jukic:
This PR includes a query endpoint to check if a FCM token exists in the store, and another update endpoint to add a token if it's not already there.

We also have a FCM token store that's optimised to reduce memory footprint, and allow better performance when looking up tokens for user.


backend/canisters/notifications_index/api/src/queries/fcm_token_exists.rs
```diff
@@ -0,0 +1,11 @@
+use candid::CandidType;
+use serde::{Deserialize, Serialize};
+use ts_export::ts_export;
+
+#[ts_export(notifications_index, fcm_token_exists)]
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub struct Args {
+    pub fcm_token: String,
```
hpeebles: This can be of type `FcmToken`

backend/canisters/notifications_index/impl/src/model/fcm_token_store.rs
```diff
@@ -0,0 +1,133 @@
...
+            Ok(())
+        } else {
+            Err("Token is not associated with current user".to_string())
+        }
+    }
+
+    /// Get all FCM tokens associated with specific users! It is very likely
+    /// we'll have to push notifications to multiple users at once, so this
+    /// method allows us to retrieve all tokens for a set of users.
+    pub fn get_tokens_for_users(&self, user_ids: Vec<&UserId>) -> Vec<&FcmToken> {
```
hpeebles: Probably better to make this `get_tokens_for_user`, since that is a fast lookup by using `self.fcm_user_tokens.range((user_id, String::default())..).take_while(|(u, _)| u == user_id)`

ivan-jukic: Ah, gotcha, leveraging the actual btree

backend/canisters/notifications_index/api/src/queries/fcm_token_exists.rs
ivan-jukic: FcmToken would need to implement TS trait?

backend/canisters/notifications_index/impl/src/model/fcm_token_store.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub fn check_token_exists(&self, token: &FcmToken) -> bool {
+        self.fcm_user_tokens.iter().any(|(_, t)| t == token)
+    }
+
+    /// Add a new FCM token, and assign it to a user.
+    ///
+    /// If the token already exists, it will not be added again - this also
+    /// ensures that the user does not have duplicate tokens, or that the same
+    /// token is not assigned to multiple users.
+    pub fn add_token(&mut self, user_id: UserId, token: FcmToken) -> Result<(), String> {
```
hpeebles: The method names should probably be `exists` (or maybe `contains`?), `add`, `remove`, and maybe `get_all` / `get_for_user`.
The type name is `FcmTokenStore` so having `token` in each method name seems redundant.

ivan-jukic: Contains sounds fine in the context of the store.

hpeebles (approved):

PR #8091, opened 2025-06-03
Read notification canisters from NotificationsIndex

hpeebles:


backend/notification_pusher/cli/src/main.rs
```diff
@@ -15,12 +14,9 @@ async fn main() -> Result<(), Error> {
 
     info!("Initializing notification pusher");
 
-    let args: Vec<String> = std::env::args().collect();
-    let index = args.get(1).map(|a| a.parse::<u64>().unwrap()).unwrap_or_default();
     let vapid_private_pem = read_env_var("VAPID_PRIVATE_PEM")?;
     let index_canister_id = Principal::from_text(read_env_var("NOTIFICATIONS_INDEX_CANISTER_ID")?)?;
-    let notifications_canister_id = Principal::from_text(read_env_var("NOTIFICATIONS_CANISTER_ID")?)?;
-    let index_store = DummyStore::new(HashMap::from([(notifications_canister_id, index)]));
```
hpeebles: You used to be able to say "start from index N", but this really isn't necessary, so to keep things simple I removed that

backend/notification_pusher/cli/src/main.rs
```diff
@@ -13,16 +11,7 @@ async fn main() -> Result<(), Error> {
...
-    let index = args.get(1).map(|a| a.parse::<u64>().unwrap()).unwrap_or_default();
-    let config = Config::init_with_store(move |envc| {
-        if let Some(ncid) = envc.notifications_canister_id {
-            Ok(DummyStore::new(HashMap::from([(Principal::from_text(ncid)?, index)])))
-        } else {
-            Err("Notifications canister ID is not provided".into())
-        }
-    })
-    .await?;
+    let config = Config::init_with_store(DummyStore::default()).await?;
```
ivan-jukic, line 14: Okay, that's cool, makes it even simpler!

PR #8103, opened 2025-06-05
Avoid cloning every chat when updating `allChatsStore`

hpeebles:
Rather than using `structuredClone` on every chat, this PR introduces the `Immutable` class, which uses the "copy on write" pattern to only clone chats as required.

### DIFF
--- a/frontend/openchat-shared/src/utils/immutable.ts
+++ b/frontend/openchat-shared/src/utils/immutable.ts
@@ -2,8 +2,8 @@ export class Immutable<T> {
     #original: Readonly<T>;
     #modified: T | undefined = undefined;
 
-    constructor(value: T, devMode = false) {
-        this.#original = devMode ? Object.freeze(value) : value;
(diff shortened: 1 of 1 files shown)


frontend/openchat-shared/src/utils/immutable.ts
```diff
@@ -0,0 +1,17 @@
+export class Immutable<T> {
```
julianjelfs, line 1: This is a bit weird. Seems like it's purpose is not really to make something immutable but rather to clone it if necessary. Not sure of the purpose of `Object.freeze`. It's only a shallow freeze and it means that if we return the original we are returning a (partially) frozen version of it. Probably not usually harmful - but why?

hpeebles: That `Object.freeze` can go, the return type is already `Readonly` so the compiler will do the same checks already.

PR #8105, opened 2025-06-05
Only skip updating chat details store if chatId matches

hpeebles:
Prior to this, we would detect that there have been no new updates and so skip setting the store value.
But when you change chat we need to update the store regardless of if there are new updates.

### DIFF
--- a/frontend/openchat-client/src/openchat.ts
+++ b/frontend/openchat-client/src/openchat.ts
@@ -3368,11 +3368,10 @@ export class OpenChat {
         switch (serverChat.kind) {
             case "group_chat":
             case "channel":
-                const lastUpdated = serverChat.lastUpdated;
                 const resp: GroupChatDetailsResponse = await this.#sendRequest({
                     kind: "getGroupDetails",
                     chatId: serverChat.id,
-                    chatLastUpdated: lastUpdated,
+                    chatLastUpdated: serverChat.lastUpdated,
                 }).catch(CommonResponses.failure);
                 if ("members" in resp) {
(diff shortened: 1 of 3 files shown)


frontend/openchat-client/src/openchat.ts
```diff
@@ -3374,21 +3374,35 @@ export class OpenChat {
...
+                if ("members" in resp) {
+                    if (!chatIdentifiersEqual(serverChat.id, selectedChatIdStore.value)) {
+                        console.warn(
+                            "Attempting to set chat details on the wrong chat - probably a stale response",
+                            serverChat.id,
+                            selectedChatIdStore.value,
+                        );
+                        return;
+                    }
+                    if (selectedServerChatStore.value?.chatId === serverChat.id && resp.timestamp <= lastUpdated) {
```
julianjelfs: I don't get why we have this check on chat Id when we have already done the check above

julianjelfs: In fact I don't think I get this change at all

julianjelfs: No wait - I _do_ get it. But it should be using `chatIdentifiersEqual` should it not?

hpeebles: The check above is on `selectedChatIdStore`, but then the store we update is the `selectedServerChatStore`, and this is the only place it is updated, so when you change chat we always need to update it, but after that, we only update it if there have actually been updates.

julianjelfs: Right but should still be using `chatIdentifiersEqual`

julianjelfs (approved):

PR #8114, opened 2025-06-06
Switch roles from being passed round as strings to instead use numbers

hpeebles:


frontend/app/src/components/home/groupdetails/Member.svelte
```diff
@@ -62,10 +70,10 @@
             canBlockUser,
     );
 
-    let ownerText = $derived($_("owner"));
-    let adminText = $derived($_("admin"));
-    let moderatorText = $derived($_("moderator"));
-    let memberText = $derived($_("member"));
+    let ownerText = $_(roleAsText(ROLE_OWNER));
```
julianjelfs: I _think_ these will not change if e.g. the user changes their selected language if this is not in a $derived. Not the end of the world but ...

megrogan (approved):

frontend/app/src/components/home/groupdetails/Member.svelte
hpeebles: Ah right yeah I'll revert that

PR #8127, opened 2025-06-09
Only return updated chats from the worker when loading updates

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -6463,38 +6447,36 @@ export class OpenChat {
     }
 
     #updateReadUpToStore(chatSummaries: ChatSummary[]): void {
-        withPausedStores(() => {
```
julianjelfs: Have we lost the pausing here? The reason that's there is because we are doing `messagesRead.syncWithServer` inside a loop which will cause multiple triggers.

frontend/openchat-client/src/openchat.ts
```diff
@@ -6119,33 +6118,34 @@ export class OpenChat {
...
         if (!anonUserStore.value) {
             userIds.add(currentUserIdStore.value);
         }
+
         await this.getMissingUsers(userIds);
 
         withPausedStores(() => {
-            if (chatsResponse.state.blockedUsers !== undefined) {
-                userStore.setBlockedUsers(chatsResponse.state.blockedUsers);
+            this.#updateReadUpToStore(chatsAddedUpdated);
```
hpeebles, line 6138: I moved the function call here, where we are already paused. Let me just check other usages.

hpeebles: This is the only place `updateReadUpToStore` is called, and it is now within the main `withPausedStores` function that we use here.

julianjelfs: OK - but - nesting is supported and then you won't forget when you move it back!

hpeebles: Yeah fair enough, have reverted that change

julianjelfs (approved):

PR #8130, opened 2025-06-09
Only return updated fields from the worker when loading updates

hpeebles:


julianjelfs (commented):
Seems like a reasonable idea but it's not very ... nice. 

Feels like there some generic `Thingy<T>` that can be concocted to make this a bit less gross?

hpeebles:
> Seems like a reasonable idea but it's not very ... nice.
> 
> Feels like there some generic `Thingy<T>` that can be concocted to make this a bit less gross?

Yeah I briefly tried this yesterday but didn't come up with anything I was happy with, I'll give it another try now that everything else is in place.

PR #8160, opened 2025-06-11
Add `initiated_by` to bot event notifications

megrogan:


backend/libraries/chat_events/src/chat_event_internal.rs
```diff
@@ -213,6 +213,42 @@ impl ChatEventInternal {
             ChatEventInternal::Empty => None,
         }
     }
+
+    pub fn initiated_by(&self) -> Option<UserId> {
+        match self {
```
hpeebles, line 218: Maybe combine all of the None cases into a single branch

megrogan: I think it's ok as is - the variants are in the same order as they are defined which has some merit

PR #8176, opened 2025-06-15
Avoid superfluous events computation after changing selected chat

hpeebles:
After selecting a new chat the route changes, this causes the `selectedChatId` to update, which then triggers the `eventsStore` to be recalculated but the `serverEventsStore` is still holding events for the previous chat.
The `serverEventsStore` then gets updated and the `eventsStore` is calculated again.

As of this PR, we clear the server events store immediately after the selected chat changes, before the events store is recalculated.

### DIFF
--- a/frontend/openchat-client/src/openchat.ts
+++ b/frontend/openchat-client/src/openchat.ts
@@ -3821,7 +3821,7 @@ export class OpenChat {
             return;
         }
         serverThreadEventsStore.update((state) => {
-            const existing = messageContextsEqual(state?.threadId, id) ? state.events : [];
+            const existing = state && messageContextsEqual(state.threadId, id) ? state.events : [];
             return {
                 threadId: id,
                 events: fn(existing),
@@ -3842,7 +3842,7 @@ export class OpenChat {
             return;
         }
         serverEventsStore.update((state) => {
-            const existing = chatIdentifiersEqual(state?.chatId, chatId) ? state.events : [];
+            const existing = state && chatIdentifiersEqual(state.chatId, chatId) ? state.events : [];
             return {
                 chatId,
                 events: fn(existing),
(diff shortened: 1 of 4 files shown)


frontend/openchat-client/src/openchat.ts
```diff
@@ -3811,7 +3811,13 @@ export class OpenChat {
             );
             return;
         }
-        serverThreadEventsStore.update(fn);
+        serverThreadEventsStore.update((state) => {
+            const existing = state?.threadId === id ? state.events : [];
```
julianjelfs: incorrect comparison klaxon. Use `messageContextsEqual`

frontend/openchat-client/src/openchat.ts
```diff
@@ -3826,7 +3832,13 @@ export class OpenChat {
             );
             return;
         }
-        serverEventsStore.update(fn);
+        serverEventsStore.update((state) => {
+            const existing = state?.chatId === chatId ? state.events : [];
```
julianjelfs: `chatIdentifiersEqual`

hpeebles: Gah!

frontend/openchat-client/src/state/app/stores.ts
```diff
@@ -543,6 +538,22 @@ export const userGroupSummariesStore = derived(communitiesStore, (communities) =
     }, new Map<number, UserGroupSummary>());
 });
 
+export const serverEventsStore = writable<EventWrapper<ChatEvent>[]>([], undefined, eqIfEmpty);
```
hpeebles, line 541: I updated some of the equalityCheck functions, other than that only the ordering of these stores changed, so that the stores related to the `selectedChatIdStore` can all be cleared as the first subscription on that store.

julianjelfs (approved):

PR #8187, opened 2025-06-17
Fix proposal tallies not updating after voting

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -5321,10 +5321,7 @@ export class OpenChat {
             proposalId,
             isNns,
         }).then((resp) => {
-            proposalTalliesStore.update((map) => {
```
hpeebles, line 5333: `eventsStore` wasn't depending on `proposalTalliesStore`, so vote tallies weren't updated after `proposalTalliesStore` was updated.
I've now moved the proposal tally updates into `localUpdates` which `eventsStore` is already dependent on.

PR #8189, opened 2025-06-17
Implement disappearing messages in direct chats

hpeebles:


megrogan (approved):

megrogan:
That was more straightforwards than I was expecting - wish we'd done it a year ago!

PR #8191, opened 2025-06-18
Ensure we exclude webhook userIds when checking for missing users

hpeebles:


frontend/openchat-shared/src/utils/chat.ts
```diff
@@ -44,10 +44,9 @@ export function userIdsFromEvents(events: EventWrapper<ChatEvent>[]): Partitione
         }
         switch (e.event.kind) {
             case "message":
+                userIds.add(e.event.sender);
```
hpeebles, line 47: All users now get added to `userIds` and the webhooks are excluded later.
This is because is some scenarios it is impossible to detect if a user is a webhook or not, eg. if a webhook message is replied to.

PR #8194, opened 2025-06-18
Add `bot_community_events` endpoint

megrogan:


backend/canisters/local_user_index/impl/src/queries/bot_community_events.rs
```diff
@@ -0,0 +1,34 @@
...
+use canister_api_macros::{query, update};
+use canister_tracing_macros::trace;
+use local_user_index_canister::bot_community_events::*;
+use oc_error_codes::OCErrorCode;
+use types::BotInitiator;
+
+#[update(candid = true, msgpack = true)]
+#[trace]
+async fn bot_community_events_c2c(args: Args) -> Response {
+    bot_community_events(args).await
```
hpeebles: You should split the main logic out and have both call that, otherwise it goes through the logging twice

hpeebles (approved):

PR #8219, opened 2025-06-23
Checking existence and assigning FCM tokens to users

ivan-jukic:
Added/fixed functions to check if an FCM token is added for a user. If not, then it is added. Added functions svelte<-> native to check for a FCM token locally on a device.

### DIFF
--- a/backend/canisters/notifications_index/api/src/updates/add_fcm_token.rs
+++ b/backend/canisters/notifications_index/api/src/updates/add_fcm_token.rs
@@ -10,6 +10,4 @@ pub struct Args {
     pub fcm_token: FcmToken,
 }
 
-#[ts_export(notifications_index, add_fcm_token)]
-#[derive(CandidType, Serialize, Deserialize, Debug)]
-pub struct Response(pub UnitResult);
+pub type Response = UnitResult;
--- a/backend/canisters/notifications_index/api/src/updates/add_fcm_token.rs
+++ b/backend/canisters/notifications_index/api/src/updates/add_fcm_token.rs
@@ -10,4 +10,6 @@ pub struct Args {
     pub fcm_token: FcmToken,
 }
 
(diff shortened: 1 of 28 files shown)


backend/canisters/notifications_index/impl/src/updates/mod.rs
```diff
@@ -5,3 +5,65 @@ mod remove_subscription;
...
+use types::{CanisterId, UserId};
+use user_index_canister::c2c_lookup_user::Response;
+use user_index_canister_c2c_client::c2c_lookup_user;
+
+pub(crate) enum LookupResult {
+    Found(UserId),
+    NotFound((Principal, CanisterId)),
+}
+
+pub(crate) enum LookupError {
```
hpeebles: You can use `OCResult` instead

backend/canisters/notifications_index/api/src/updates/add_fcm_token.rs
```diff
@@ -10,4 +10,6 @@ pub struct Args {
     pub fcm_token: FcmToken,
 }
 
-pub type Response = UnitResult;
+#[ts_export(notifications_index, add_fcm_token)]
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub struct Response(pub UnitResult);
```
hpeebles: This wrapper type isn't needed

frontend/openchat-agent/src/services/openchatAgent.ts
```diff
@@ -2718,6 +2721,17 @@ export class OpenChatAgent extends EventTarget {
         return this._notificationClient.removeSubscription(subscription);
     }
 
+    fcmTokenExists(fcm_token: string): Promise<boolean> {
```
hpeebles: You're still in Rust casing mode

PR #8227, opened 2025-06-25
Fix bot messages incorrectly appearing as ephemeral messages

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -8804,13 +8804,36 @@ export class OpenChat {
...
                                     removePlaceholder?.();
+                                    publish("ephemeralMessage", {
+                                        message: resp.message,
+                                        scope,
+                                        botId: bot.id,
+                                        commandName: bot.command.name,
+                                    });
+                                } else if (scope.kind === "chat_scope") {
+                                    const chat = allChatsStore.value.get(scope.chatId);
+                                    if (chat !== undefined) {
```
julianjelfs, line 8815: This looks like it is half recreating what's being done in savePlaceholderBotMessage rather than just using it? There seems to be stuff not being done now.

hpeebles: But that is posting the placeholder message, whereas the change I made is to post the response from the bot, which is unconfirmed at first and will later be confirmed once it gets confirmed in the canister.

julianjelfs: That's exactly what the placeholder message is. A message which will later get replaced by what comes back from the server.

hpeebles: The placeholder message is displayed before sending the command to the bot, the bot then replies and also sends the same message to the canister.
By dropping an unconfirmed message in the chat we avoid waiting for the bot's response message to be confirmed in the canister.

julianjelfs: Yes and that's exactly what sendBotPlaceholder already does.

julianjelfs: It's meant to be called more than once and it covers both scenarios.

hpeebles: Ahhhh I see what you mean. You mean we can simply call it again. I thought you meant it already was doing everything we needed.

julianjelfs: Yes that's the intention. Not sure if / when that got messed up. Probably the ephemeral message change.

PR #8234, opened 2025-06-25
Introduce and backdate a ChannelCreated event

megrogan:


megrogan:
This hasn't been tested yet so don't merge

backend/canisters/community/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -41,4 +44,60 @@ fn post_upgrade(args: Args) {
...
+
+        // Check that all events have an event index corresponding to their index in the vector
+        for (index, event) in community_events.iter().enumerate() {
+            if event.index != EventIndex::from(index as u32) {
+                panic!("Event index mismatch: expected {}, got {}", index, event.index);
+            }
+        }
+
+        // Insert ChannelCreated events into CommunityEvents
+        for channel in state.data.channels.iter() {
```
hpeebles, line 61: I think it would be much simpler to just append the new events then sort the whole list by timestamp.

hpeebles: What you've done here is probably correct, but its hard to validate just by eyeballing it.

megrogan: I want to keep the order the same as before when events have the same timestamp

hpeebles: Then order by timestamp and then event index

hpeebles: In fact just use a stable sort and it will preserve the order of equal elements.

PR #8249, opened 2025-06-27
Fix ChannelCreated timestamp for imported groups

megrogan:


megrogan:
I accidentally tagged the PR commit which is why pre-release tests ran...

hpeebles (approved):

PR #8252, opened 2025-06-27
Update the cache primer to include proposal chats

hpeebles:


frontend/openchat-agent/src/utils/cachePrimer.ts
```diff
@@ -207,9 +206,9 @@ export class CachePrimer {
     }
 
     private shouldEnqueueChat(chat: ChatSummary, lastUpdated: bigint | undefined): boolean {
-        if (chat.membership.archived || isProposalsChat(chat)) return false;
-
-        return lastUpdated === undefined || chat.lastUpdated > lastUpdated;
+        return chat.membership.archived
```
megrogan: A single logic statement would be better

hpeebles: `return !chat.membership.archived && (lastUpdated === undefined || chat.lastUpdated > lastUpdated);`

megrogan (approved):

PR #8257, opened 2025-06-30
Hide member joined events in public channels

megrogan:


frontend/openchat-client/src/utils/chat.ts
```diff
@@ -563,7 +569,11 @@ export function flattenTimeline(grouped: EventWrapper<ChatEvent>[][][]): Timelin
     return timeline;
 }
 
-export function isEventKindHidden(kind: ChatEvent["kind"]): boolean {
+export function isEventKindHidden(kind: ChatEvent["kind"], isPublicChannel: boolean): boolean {
+    if (isPublicChannel && kind === "member_joined") {
+        return true;
+    }
+
```
hpeebles: I think this should be moved into the switch statement

hpeebles (approved):

PR #8261, opened 2025-06-30
Restructured native notification data on the backend

ivan-jukic:
If notifications do not have "notification" data set, only attached data in a HashMap structure, the Firebase service will not auto handle the recieved notification if the app is in the background or not running.

This PR introduces changes to the FCM data on the backend to denote specific parts of the info that we require. We've also modified the native app code to allow for handling of notifications.

As an added bonus, we also have a preparation for a user avatar to be shown in the notification, for now it's only a placeholder.

image

### DIFF
--- a/backend/canisters/community/impl/src/updates/add_members_to_channel.rs
+++ b/backend/canisters/community/impl/src/updates/add_members_to_channel.rs
@@ -4,6 +4,7 @@ use crate::{
 use canister_api_macros::update;
 use canister_tracing_macros::trace;
 use community_canister::add_members_to_channel::{Response::*, *};
+use futures::channel;
 use oc_error_codes::OCErrorCode;
 use types::{AddedToChannelNotification, ChannelId, CommunityId, FcmData, OCResult, UserId, UserNotificationPayload, UserType};
 
@@ -111,12 +112,10 @@ fn commit(
 
     // TODO i18n
     let fcm_body = format!("You have been added to channel {}", channel_name.clone());
-    let fcm_data = FcmData::builder()
-        .with_body(fcm_body)
-        .with_chat_id(community_id.to_string())
-        .with_alt_sender_name(&added_by_display_name, &added_by_name)
-        .with_sender_avatar_id(community_avatar_id)
-        .build();
+    let fcm_data = FcmData::for_community_chat(community_id, channel_id)
+        .set_body(fcm_body)
+        .set_sender_name_with_alt(&added_by_display_name, &added_by_name)
+        .set_avatar_id(community_avatar_id);
 
     let notification = UserNotificationPayload::AddedToChannel(AddedToChannelNotification {
         community_id: state.env.canister_id().into(),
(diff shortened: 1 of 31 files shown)


backend/canisters/community/impl/src/updates/send_message.rs
```diff
@@ -252,8 +252,10 @@ fn process_send_message_result(
 
         // TODO i18n
         let fcm_data = FcmData::builder()
-            .with_alt_title(&sender_display_name, &sender_username)
             .with_alt_body(&message_text, &message_type)
+            .with_chat_id(community_id.to_string())
```
hpeebles: You're setting the chatId to be the communityId, this probably isn't quite right.
You'll most likely need the channelId too.

ivan-jukic: Yes, semantics here is a bit weird 🤔 this value should be either community or group id. If this value is not set, but we have a sender id, then it's a direct chat. I thought that the chat id would encapsulate both community and group chats. Does this make sense? Perhaps I need to add a comment about this? Or, maybe we need to have that distinction between community and a group chat, and functions with_community_id and with_group_id?

ivan-jukic: Ah yes, distinction will be required for routing.

hpeebles: Yeah I think separate it into group_id and community_id

hpeebles: You'll likely need channel_id too

hpeebles: So the function could be `with_channel_id` and it takes the communityId + the channelId

backend/canisters/community/impl/src/updates/add_members_to_channel.rs
```diff
@@ -4,8 +4,9 @@ use crate::{
 use canister_api_macros::update;
 use canister_tracing_macros::trace;
 use community_canister::add_members_to_channel::{Response::*, *};
+use futures::channel;
```
hpeebles: I'm pretty sure this isn't used

backend/libraries/types/src/fcm_data.rs
```diff
@@ -1,90 +1,152 @@
...
-    image: Option<String>,
-    data: HashMap<String, String>,
-}
+    pub fn for_direct_chat(direct_chat_id: ChatId) -> Self {
+        Self::default(Chat::Direct(direct_chat_id))
+    }
 
-impl FcmDataBuilder {
-    fn new() -> Self {
+    pub fn for_group_chat(group_chat_id: ChatId) -> Self {
```
hpeebles: I think for_group, for_channel and for_direct_chat (as it currently is) would be better names

backend/libraries/types/src/fcm_data.rs
```diff
@@ -1,90 +1,152 @@
...
 
-#[derive(Default)]
-pub struct FcmDataBuilder {
-    title: String,
-    body: String,
-    image: Option<String>,
-    data: HashMap<String, String>,
-}
+    pub fn for_direct_chat(direct_chat_id: ChatId) -> Self {
+        Self::default(Chat::Direct(direct_chat_id))
```
hpeebles: If you also set the sender_id then later you can avoid matching on the type when writing the sender_id to the map

hpeebles:
Needs the changelog entries

PR #8284, opened 2025-07-02
Don't trigger build+test when only markdown files change

kitsiosk:
### Change Summary
Currently, the `Backend linting and unit tests` CI workflow will run even when a push changes **only markdown files** (e.g., README.md). 

With these changes, a push that changes *only* markdown files will not trigger the workflow, thus speeding up the project's CI and saving compute resources 🌱 (see below for quantity).

### Example
Here is an example of the behaviour described above: the commit [`67caf1c9`](https://github.com/open-chat-labs/open-chat/commit/67caf1c991bbc3f06b65141b4ff56532cac95bae) changed these files: `backend/canisters/user_index/CHANGELOG.md` and, when pushed, triggered [this](https://github.com/open-chat-labs/open-chat/actions/runs/9371553889/) workflow run, which ran for ~30 CPU minutes. With the proposed changes, these 30 CPU minutes would have been saved, clearing the queue for other workflows and speeding up the CI of the project, while also saving resources in general 🌱. 

Note that this is a single example out of (a minimum of) 48 examples over the last few months; a lower estimate for the accumulated gain is **22 CPU hours**, but the actual number could be much **higher** because we could only look at a subset of the data.

### Context
Hi,

We are a team of [researchers](https://www.ifi.uzh.ch/en/zest.html) from University of Zurich and we are currently working on energy optimizations in GitHub Actions workflows.

Kindly let us know (here or in the email below) if you would like to **ignore other file types** or apply the filter to **other workflows*, or if you have any question whatsoever.

Best regards,  
[Konstantinos Kitsios](https://www.ifi.uzh.ch/en/zest/team/konstantinos_kitsios.html)  
konstantinos.kitsios@uzh.ch

### DIFF
--- a/.github/workflows/backend.yaml
+++ b/.github/workflows/backend.yaml
@@ -2,6 +2,8 @@ on:
   pull_request:
     branches:
       - master
+    paths-ignore:
+      - '**/*.md'
 
 name: Backend linting and unit tests
 


hpeebles:
Hey! Thanks for the PR!
But we've actually tried doing this before, and we then run into the issue that those checks are "required" checks.
So if we skip them the PR can't be merged.
I haven't looked into this in probably ~2 years so maybe there is a solution now though.
Basically we'd want GitHub to mark the PR as ok if the required checks either complete successfully or were skipped.
An alternative could be to change the checks to handle this manually, so when certain conditions are met the checks still run, but they simply skip doing the actual processing.

hpeebles:
It seems like it is still yet to be resolved - https://github.com/orgs/community/discussions/44490

kitsiosk:
Thanks for your prompt reply!

> we then run into the issue that those checks are "required" checks.

Right, I didn't think of it. Let me have a look and get back to you shortly

kitsiosk:
Hi again,

You were right, this seems like a weird behavior from the GitHub side. In the issue you mentioned, a GitHub developer says they will work on it in the next 12 months (posted September '24). Let's see! Till then, feel free to close this one if you wish.

Thanks a lot for your feedback.

PR #8286, opened 2025-07-03
Revert local network changes in dfx.json

megrogan:


dfx.json
```diff
@@ -192,9 +192,10 @@
   },
   "networks": {
     "local": {
-      "bind": "0.0.0.0:8080",
-      "proxy": {
-        "domain": "0.0.0.0"
+      "bind": "127.0.0.1:8080",
+      "type": "ephemeral",
+      "replica": {
+        "subnet_type": "system"
```
hpeebles, line 197: Do we need this?

hpeebles: We don't need `"type": "ephemeral"`, that is the default for local

hpeebles: Yeah I've just been looking at the DFX code, let's keep the subnet type so that the NNS canisters work as expected

megrogan: Sure I can remove it - I just reverted to what we had before

PR #8287, opened 2025-07-03
Add an optional ChannelId to CommunityMemberJoined

megrogan:


backend/canisters/community/impl/src/model/events.rs
```diff
@@ -285,3 +285,23 @@ impl From<CommunityEventInternal> for CommunityEvent {
         }
     }
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct CommunityMemberJoinedInternal {
+    #[serde(rename = "u", alias = "user_id")]
```
hpeebles, line 291: Are the aliases needed? Because this hasn't been released yet, right?

PR #8298, opened 2025-07-04
Native notifications are displayed and react to current app context

ivan-jukic:
These changes allow the native android notifications to either be shown or omitted depending on the current app context; and to show the correct context when tapped on (if not already).

For example, this means that if the user is currently within a DM context, any notification received for their current chat will not be shown, but as soon as they change their app's context, notifications for that chat will start appearing. Tapping on the notifications will put the user back into that DM context.

Since our code is split between UI Svelte and native Kotlin, these have to communicate for this to work:
- Svelte code reports to native if it's ready to receive events
- If svelte code has not reported ready, events are stored in a queue
- Once UI is ready any queued events are flushed

We are also tracking if the app is in the foreground or background
- If the app is in the background or closed, we immediately display any notification via native code
- If the app is in the foreground, notification is sent to UI, UI checks if the notification should be shown or not; if yes, UI sends it back to native to show the notification. This way we have one place in native code where notifications are constructed.


ivan-jukic:
Note to self - changelogs.

Changes in this PR will require release of user, group, community and local_user_index canisters + notifications_pusher.

backend/canisters/community/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Remove duplication by removing `MessageContent::message_type` function ([#8293](https://github.com/open-chat-labs/open-chat/pull/8293))
 - Deprecate `winners` field on prize messages ([#8302](https://github.com/open-chat-labs/open-chat/pull/8302))
+- Re-enabled fcm_data; needs to be released with group, user and local_user_index canisters ([8298](https://github.com/open-chat-labs/open-chat/pull/8298))
```
hpeebles: "needs to be released with group, user and local_user_index canisters", canister upgrades are always sequential, so I think you should clarify the order dependencies

backend/canisters/group/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use MessagePack to serialize upgrade args ([#8269](https://github.com/open-chat-labs/open-chat/pull/8269))
 - Remove duplication by removing `MessageContent::message_type` function ([#8293](https://github.com/open-chat-labs/open-chat/pull/8293))
 - Deprecate `winners` field on prize messages ([#8302](https://github.com/open-chat-labs/open-chat/pull/8302))
+- Re-enabled fcm_data; needs to be released with community, user and local_user_index canisters ([8298](https://github.com/open-chat-labs/open-chat/pull/8298))
```
hpeebles: Same here

backend/canisters/local_user_index/CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Deprecate `winners` field on prize messages ([#8302](https://github.com/open-chat-labs/open-chat/pull/8302))
 
+### Changed
+- Re-enabled fcm_data; needs to be released with community, group and user canisters ([8298](https://github.com/open-chat-labs/open-chat/pull/8298))
```
hpeebles: and here

backend/libraries/types/src/notifications.rs
```diff
@@ -29,7 +29,7 @@ pub struct UserNotification {
     pub notification_bytes: ByteBuf,
 
     // Values relevant for the FCM notifications
-    #[serde(skip_deserializing, rename = "f")]
+    #[serde(default, rename = "f")]
```
hpeebles: `serde(default)` already applies to `Option` automatically

ivan-jukic: Nice, thanks for that nugget!

backend/canisters/community/CHANGELOG.md
ivan-jukic: I'll just remove since the order doesn't matter.

hpeebles (approved):

PR #8305, opened 2025-07-07
Add shorter aliases to `MessageContent` fields

hpeebles:


julianjelfs:
I'm worried that the ts bot sdk will have to parse it which is fucking horrible.

hpeebles:
For end to end encryption the message content will be serialized and the bytes will be encrypted, then those bytes will need to be deserialized again forever more. Hence why I thought we'd reduce the sizes now, otherwise we never can.

PR #8308, opened 2025-07-08
Use sign in when linking accounts with passkeys

ivan-jukic:
This fix should allow users to select their other passkey that they may use when logging into OpenChat, and link it with their current identity.

### DIFF
--- a/frontend/openchat-client/src/openchat.ts
+++ b/frontend/openchat-client/src/openchat.ts
@@ -9348,7 +9348,7 @@ export class OpenChat {
             chatId,
         });
 
-        if (chatId === selectedChatIdStore.value) {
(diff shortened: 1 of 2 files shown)


frontend/app/src/components/home/profile/LinkAccounts.svelte
```diff
@@ -133,7 +133,7 @@
 
         if (provider === AuthProvider.PASSKEY) {
             if (substep.kind === "initiator") {
-                const [identity, delegation, webAuthnKey] = await client.signUpWithWebAuthn(false);
+                const [identity, delegation, webAuthnKey] = await client.signInWithWebAuthn(false);
```
hpeebles, line 136: Hmm we probably want to ask users if they already have a passkey or not, because if they want to add a new passkey to their account then this won't work

PR #8315, opened 2025-07-09
Send user a message each time a day of streak insurance is used up

hpeebles:


backend/canisters/user/impl/src/lib.rs
```diff
@@ -209,6 +210,15 @@ impl RuntimeState {
             LocalUserIndexEvent::NotifyStreakInsuranceClaim(claim),
         ];
         self.push_local_user_index_canister_events(events, self.env.now());
+        openchat_bot::send_text_message(
+            format!(
+                "One day of streak insurance was just used up in order to protect your streak from being lost.\
```
megrogan: remove "in order"

backend/canisters/user/impl/src/lib.rs
```diff
@@ -209,6 +210,15 @@ impl RuntimeState {
...
+Your streak is now {new_streak} days."
```
megrogan: ... and you have {days_left} days of streak insurance left.

megrogan (approved):

PR #8319, opened 2025-07-10
Added script to build Android app release version & tweaks

ivan-jukic:
Added a script which builds a release version of the Android app. The build process uses a debug key/certificate to sign the app, since it matches the one set in the app's assetlinks.json. This allows verification agains the app's domain.

Actual app release will be signed with a dedicated key/certificate.

Included a few tweaks
- Resize viewport on keyboard show/hide
- Add status bar offset to the top (needs a bit of improvement)
- Disabled pinch-to-zoom for the native app

### DIFF
--- a/frontend/turbo.json
+++ b/frontend/turbo.json
@@ -77,6 +77,14 @@
             "persistent": true,
             "dependsOn": ["^wait"]
         },
+        "android:release": {
+            "outputs": ["lib/**", "build/**"],
+            "dependsOn": [
+                "^android:release",
+                "openchat-service-worker#build",
+                "openchat-worker#build"
+            ]
+        },
         "wait": {
             "cache": false
         },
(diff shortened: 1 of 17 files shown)


frontend/turbo.json
```diff
@@ -77,6 +77,14 @@
             "persistent": true,
             "dependsOn": ["^wait"]
         },
+        "android:release": {
+            "outputs": ["lib/**", "build/**"],
+            "dependsOn": [
+                "^android:release",
```
hpeebles, line 83: Could this just depend on `^build`?
Then you wouldn't need to add the new steps to each package.json other than the app one, and we could do the same with `build:docker` too.

ivan-jukic: Not sure / I don't think so, it's set up this way so that in the end of the build chain we have a call to a specific app npm script. Maybe there's a way to depend on build for all packages, and then specify a specific npm script for the app, but Turbo is quite quirky, and this seems to work for now.

I hope to revisit this at some point, but as a separate PR.

PR #8320, opened 2025-07-10
When a bot invites a user also send an invitation

megrogan:


hpeebles:
You need to pass in the username so that the notifications show the correct text

megrogan:
> You need to pass in the username so that the notifications show the correct text

As discussed this hasn't changed - please review again thanks!

PR #8321, opened 2025-07-10
Fix temporarily not appearing as channel member after creating community

hpeebles:


frontend/openchat-client/src/state/localUpdates.ts
```diff
@@ -442,7 +442,11 @@ export class GlobalLocalState {
 
     addCommunity(val: CommunitySummary) {
         this.removeCommunityPreview(val.id);
-        return addToWritableLocalMap(val.id, val, this.communities);
+        addToWritableLocalMap(val.id, val, this.communities);
+        this.chats.update((map) => {
```
hpeebles, line 446: This is the only actual change, the channels were not being added to the chat list so when the new community was added to the list, the channels appeared as "Other channels" since according to the state the user was not in the channels.

julianjelfs (approved):

PR #8322, opened 2025-07-10
Fix member count on new groups with invited users

hpeebles:


frontend/app/src/components/home/createOrUpdateGroup/CreateOrUpdateGroup.svelte
```diff
@@ -424,7 +426,7 @@
                     <div class="members">
                         <ChooseMembers
                             userLookup={searchUsers}
-                            bind:members={candidateGroup.members}
```
hpeebles, line 427: This was causing the candidate group to have multiple members, even though these users are only invited, they are not already members. So I've introduced a new variable to hold the invited users to avoid incorrectly modifying the candidate group members.

frontend/openchat-client/src/utils/chat.ts
```diff
@@ -869,7 +869,7 @@ export function groupChatFromCandidate(
         minVisibleEventIndex: 0,
         minVisibleMessageIndex: 0,
         lastUpdated: BigInt(0),
-        memberCount: candidate.members.length + 1, // +1 to include us
+        memberCount: 1,
```
hpeebles, line 872: For a new group this is always 1

PR #8356, opened 2025-07-15
Generalise Escrow so it can be used by 3rd parties

megrogan:
Integration tests to follow...done


backend/canisters/escrow/impl/src/model/swaps.rs
```diff
@@ -69,12 +72,15 @@ pub struct Swap {
...
+    pub fn new(id: u32, caller: Principal, args: escrow_canister::create_swap::Args, now: TimestampMillis) -> Swap {
+        let offered_by = args.token0_principal.unwrap_or(caller);
+
         Swap {
             id,
             location: args.location,
             created_at: now,
-            created_by: caller,
+            offered_by,
+            accept_by: args.token1_principal,
```
hpeebles: accept_by sounds a bit weird. Maybe restricted_to?

backend/canisters/escrow/impl/src/queries/deposit_account.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{RuntimeState, read_state};
+use canister_api_macros::query;
+use canister_tracing_macros::trace;
+use escrow_canister::deposit_account::{Response::*, *};
+use icrc_ledger_types::icrc1::account::Account;
+
+#[query(candid = true, msgpack = true)]
+#[trace]
+fn deposit_accounts(args: Args) -> Response {
```
hpeebles: deposit_account

backend/canisters/escrow/impl/src/queries/deposit_account.rs
```diff
@@ -0,0 +1,28 @@
...
+    read_state(|state| deposit_accounts_impl(args, state))
+}
+
+fn deposit_accounts_impl(args: Args, state: &RuntimeState) -> Response {
+    if let Some(swap) = state.data.swaps.get(args.swap_id) {
+        let principal = args.principal.unwrap_or(state.env.caller());
+
+        if swap.offered_by != principal || swap.accept_by.is_some_and(|p| p != principal) {
```
hpeebles: && rather than ||

backend/canisters/escrow/impl/src/updates/notify_deposit.rs
```diff
@@ -82,42 +83,45 @@ fn prepare(args: &Args, state: &mut RuntimeState) -> Result<PrepareResult, Respo
...
-                if accepted_by == user_id {
+                if accepted_by == principal {
                     Err(Success(SuccessResult {
                         complete: swap.token0_received,
                     }))
                 } else {
                     Err(SwapAlreadyAccepted)
                 }
-            } else {
+            } else if swap.accept_by == Some(principal) || swap.accept_by.is_none() {
```
hpeebles: is_none_or

backend/canisters/escrow/impl/src/queries/deposit_account.rs
hpeebles: I think this should be something like lookup_swap and return the full swap details. Otherwise there's no way to send someone a swap and for them to validate the swap is what you say it is.
Then as part of that it can return the deposit account.

backend/canisters/escrow/impl/src/queries/lookup_swap.rs
```diff
@@ -0,0 +1,41 @@
...
+fn lookup_swap_impl(args: Args, state: &RuntimeState) -> Response {
+    let Some(swap) = state.data.swaps.get(args.swap_id) else {
+        return SwapNotFound;
+    };
+
+    let accepting_principal = args.accepting_principal.unwrap_or(state.env.caller());
+
+    let escrow_canister_id = state.env.canister_id();
+
+    if swap.restricted_to.is_some_and(|p| p != accepting_principal) {
```
hpeebles, line 21: This should work if you are the offered_by principal too

megrogan: Anyone can call this endpoint but if the accepter is restricted then it must match the accepting_principal otherwise the token1_deposit_address would be wrong. The offerer will know the accepter's principal in this case so they can call this endpoint just fine. So I don't think the PR needs changing.

PR #8358, opened 2025-07-16
Allow account linking via a code

ivan-jukic:
Adds a feature to create account linking codes. These will be used with native apps to link existing accounts to the passkeys created on these devices.


backend/canisters/identity/impl/src/lib.rs
```diff
@@ -142,6 +143,14 @@ impl RuntimeState {
         }
     }
 
+    pub fn get_user_id_by_caller(&self) -> Option<UserId> {
+        let caller = self.env.caller();
+        self.data
+            .user_principals
+            .get_by_auth_principal(&caller)
```
ivan-jukic, line 155: Does this look right 🤔 ?

backend/integration_tests/src/account_linking_tests.rs
```diff
@@ -0,0 +1,122 @@
...
+        GetAccountLinkingCodeResponse::Success(code) => code,
+        response => panic!("'get_account_linking_code' error: {response:?}"),
+    };
+
+    assert_eq!(fetched_linking_code, created_linking_code);
+
+    //
+    // ---- Initialise new identity to link ----
+    //
+    let (new_principal, new_pub_key) = random_delegated_principal(canister_ids.sign_in_with_email);
```
ivan-jukic, line 58: I'm slightly confused about the requirements for the canister id here. It's later validated on the backend, and it must be from expected set of canister ids. Could anyone shed a bit of light on this?

I've not used Webauthn key here, would that yield a different result if I had used a different canister id here?

scripts/run-integration-tests.sh
```diff
@@ -49,6 +49,6 @@ function cleanup() {
 
 trap cleanup EXIT
 
-cargo test --package integration_tests $TESTNAME -- --test-threads $TEST_THREADS || exit 1
+cargo test --package integration_tests $TESTNAME -- --nocapture --test-threads $TEST_THREADS || exit 1
```
ivan-jukic, line 52: Could we keep this 🤔 ? I find it useful sometimes if I want to println! something in the test... Is there a different way you'd recommend?

scripts/run-integration-tests.sh
```diff
@@ -49,6 +49,6 @@ function cleanup() {
```
ivan-jukic, line 49: Note to self, add changelog.

backend/canisters/identity/impl/src/lib.rs
```diff
@@ -184,6 +194,8 @@ struct Data {
     rng_seed: [u8; 32],
     challenges: Challenges,
     test_mode: bool,
+    #[serde(default)]
+    account_linking_codes: HashMap<String, AccountLinkingCode>,
```
hpeebles: It's probably worth putting this in its own struct which can expose the various required helper functions, and we'll likely want additional fields in the future too such as last_updated, etc.

backend/integration_tests/src/account_linking_tests.rs
hpeebles: With delegated identities, the identity is generated from a principal (in this case a canisterId) and a subaccount.
So for example if you sign in via the SignInByEmail canister, then the principal will be the SignInByEmail canisterId and the subaccount will be different for each user.

hpeebles: The Identity canister verifies that the originating canister is one of the whitelisted ones.
And for some of them it enforces that a captcha is required.

backend/integration_tests/src/account_linking_tests.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    //
+    // ---- Verify that the linking code can be used to link accounts ----
+    //
+
+    // In a real scenario, to avoid running an update call, we first verify
+    // that the linking code is valid with a query. Only after that we proceed
+    // to link the accounts using an update call.
+    let VerifyAccountLinkingCodeResponse(verify_success) = client::identity::verify_account_linking_code(
```
hpeebles: This is potentially a security issue, since queries don't modify state, someone could send queries attempting to find valid account linking codes and we'd have no way of detecting it.
I think it would be better to skip this query call and only perform the check via an update, so we can detect every attempt and throttle if necessary.

backend/integration_tests/src/account_linking_tests.rs
```diff
@@ -0,0 +1,122 @@
...
+    //
+    let get_linking_code_response = client::identity::get_account_linking_code(
+        env,
+        user_auth.auth_principal,
+        canister_ids.identity,
+        &identity_canister::get_account_linking_code::Args {},
+    );
+
+    assert_eq!(get_linking_code_response, GetAccountLinkingCodeResponse::NotFound);
+}
```
hpeebles, line 87: Maybe add a final check that the account has been linked successfully by querying for the account using the new principal.

backend/canisters/identity/impl/src/queries/get_account_linking_code.rs
```diff
@@ -0,0 +1,33 @@
...
+    } else {
+        NotFound
+    }
+}
+
+fn find_account_linking_code_for_user_id(state: &RuntimeState, user_id: UserId) -> Option<AccountLinkingCode> {
+    state
+        .data
+        .account_linking_codes
+        .iter()
```
hpeebles: `iter()` -> `values()`

backend/canisters/identity/impl/src/updates/create_account_linking_code.rs
```diff
@@ -0,0 +1,53 @@
...
+const MAX_CODE_GEN_ATTEMPTS: usize = 20;
+
+#[update(msgpack = true, candid = true)]
+#[trace]
+fn create_account_linking_code(_args: Args) -> Response {
+    mutate_state(create_account_linking_code_impl)
+}
+
+fn create_account_linking_code_impl(state: &mut RuntimeState) -> Response {
+    let caller_user_id = state.get_user_id_by_caller();
```
hpeebles:
```suggestion
    let Some(user_id) = state.get_user_id_by_caller() else {
        return UserNotFound;
    };
```

backend/canisters/identity/impl/src/updates/create_account_linking_code.rs
```diff
@@ -0,0 +1,53 @@
...
+    let now = state.env.now();
+    let rng = state.env.rng();
+
+    if let Some(user_id) = caller_user_id {
+        // Clean up expired codes - keeps the memory footprint smaller in
+        // exchange for a bit of extra CPU time.
+        state
```
hpeebles: I think this logic should all be moved into an `AccountLinkingCodes` struct.

backend/canisters/identity/impl/src/updates/link_with_account_linking_code.rs
```diff
@@ -0,0 +1,104 @@
...
+//
+// TODO Consider ways to make this functionality more secure: set limit on the
+// number of attempts to link; lock-out after too many failed attempts; log
+// and audit failed attempts; consider reducing validity time for linking codes.
+fn link_with_account_linking_code_impl(args: Args, state: &mut RuntimeState) -> OCResult {
+    let now = state.env.now();
+
+    // Basically checks if the code provided by the user match any of the codes
+    // that are saved.
+    let Some(linking_code) = state.data.account_linking_codes.get(&args.code) else {
```
hpeebles, line 32: Maybe `account_linking_codes.remove(..)`

backend/integration_tests/src/account_linking_tests.rs
ivan-jukic: Yes, that's true, but it's highly unlikely that a brute force attack would work. Codes last very short, there's 62^6 possible permutations for a 6 character linking code, and they're deleted as soon as they're used. The only way a malicious user could use one of the codes is if they actually see it, and use it before the originating user.

Could we throttle requests on the Cloudflare level?

backend/canisters/identity/impl/src/lib.rs
hpeebles: Maybe `get_caller_user_id`?

backend/canisters/identity/impl/src/queries/get_account_linking_code.rs
```diff
@@ -0,0 +1,33 @@
...
+#[query(msgpack = true, candid = true)]
+fn get_account_linking_code(_args: Args) -> Response {
+    read_state(get_account_linking_code_impl)
+}
+
+// Fetches existing account linking code for the caller's user ID.
+fn get_account_linking_code_impl(state: &RuntimeState) -> Response {
+    let now = state.env.now();
+
+    if let Some(user_id) = state.get_user_id_by_caller() {
```
hpeebles: You could remove the duplicate `NotFound` cases by returning in the `Success` case and then having `NotFound` as the fallback on the last line.

backend/canisters/identity/impl/src/queries/verify_account_linking_code.rs
```diff
@@ -0,0 +1,24 @@
...
+// TODO Make this a bit more secure: rate limit number of request.
+fn verify_account_linking_code_impl(args: Args, state: &RuntimeState) -> Response {
+    let now = state.env.now();
+
+    if let Some(linking_code) = state.data.account_linking_codes.get(&args.code) {
+        // Check if the account linking code is valid!
+        Response(linking_code.is_valid(now) && linking_code.value == args.code)
+    } else {
+        Response(false)
+    }
```
hpeebles:
```suggestion
    let is_valid = state.data.account_linking_codes
        .get(&args.code)
        .map_or(false, |lc| lc.is_valid(now) && lc.value == args.code);
    Response(is_valid)
```

backend/integration_tests/src/account_linking_tests.rs
hpeebles: But realistically everyone is going to enter the correct code, so they'd generate the code with one update call, then enter it with another.
These calls don't go via CloudFlare, they go directly to the IC API boundary nodes which then route them to the canister, so there's no way we can log them.
So it would in theory be possible for someone to just hammer query calls looking for valid requests and they could leave it running non-stop and we'd have no way to know (the subnet metrics would should a big spike in queries, but we wouldn't be able to know what those query calls were).

backend/libraries/types/src/account_linking_code.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        // Map bytes to characters
+        let code: String = bytes[..ALC_LENGTH]
+            .iter()
+            .map(|&b| {
+                let idx = (b as usize) % ALC_CHARSET.len();
+                ALC_CHARSET[idx] as char
+            })
+            .collect();
```
hpeebles: Untested so likely to fail building, but something like this could be simpler -

```suggestion
        let code: String = (0..ALC_LENGTH)
            .map(|_| {
                let idx = (rng.next_u32() as usize) % ALC_CHARSET.len();
                ALC_CHARSET[idx] as char
            })
            .collect();
```

backend/canisters/identity/impl/src/queries/get_account_linking_code.rs
ivan-jukic: I've removed this file, and moved the functionality to create linking code update handler.

backend/libraries/types/src/account_linking_code.rs
ivan-jukic: Thanks for the suggestion 🙏🏻 though I'm afraid the simplicity of each may be up to the interpreter 😅  I don't see them as much different, and the current code works for sure, so I'd rather not tinker with it for now.

backend/canisters/identity/impl/src/queries/verify_account_linking_code.rs
ivan-jukic: Also, file removed.

backend/integration_tests/src/account_linking_tests.rs
ivan-jukic: Okay, thanks for clarification. I've removed this method, won't be in the next push.

This is a bit odd though that we basically have no control over the queries. I suppose Dfinity will/should do something about it.

When it comes to someone hammering a query, they can hammer any query endpoint if their intention is DDoS-ing  us, and I really doubt it any hacker would try to guess a random linking code for a random account, it's not worth it. Besides we can make the code 8 chars long, which puts the possible permutations into quadrillions for 50ish different characters, plus we can have a confirmation "yes it's me / or it's not me" dialog on the web app when linking, asking the user to confirm that it's indeed them, and make this basically bulletproof.

backend/integration_tests/src/account_linking_tests.rs
ivan-jukic: Added! Please give it a check.

backend/libraries/types/src/account_linking_code.rs
```diff
@@ -0,0 +1,103 @@
...
+            user_id,
+        }
+    }
+
+    pub fn is_valid(&self, now: u64) -> bool {
+        now < self.expires_at
+    }
+
+    pub fn is_valid_for_more_than_a_minute(&self, now: u64) -> bool {
+        (self.expires_at - now) > (60 * 1000)
```
hpeebles: This could underflow

backend/libraries/types/src/account_linking_code.rs
```diff
@@ -0,0 +1,103 @@
...
+use ts_export::ts_export;
+
+// For [a-zA-Z0-9] characters this gives us 62^6 = 56,800,235,584 possible combinations
+const ALC_LENGTH: usize = 6;
+const ALC_DURATION: u64 = 5 * 60 * 1000; // 5 minutes in milliseconds
+const ALC_CHARSET: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[ts_export]
+pub struct AccountLinkingCode {
```
hpeebles: I think this should be moved to "identity_canister/src/model" as it won't be used by any other canisters

ivan-jukic: You mean identity/impl/src/model? How would you resolve cyclical dep then with identity/api?

backend/libraries/types/src/account_linking_code.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// Check if there's no codes
+    pub fn is_empty(&self) -> bool {
+        self.codes.is_empty()
+    }
+
+    // Generates a random 6 character string.
+    fn generate_code(rng: &mut StdRng) -> String {
+        let bytes: [u8; 32] = rng.r#gen();
```
hpeebles: Rather than generating 32 bytes, just generate `ALC_LENGTH` bytes

backend/canisters/identity/impl/src/updates/create_account_linking_code.rs
```diff
@@ -0,0 +1,37 @@
...
+    // Clean up expired codes, keeps the memory footprint smaller in
+    // exchange for a bit of extra CPU time.
+    state.data.account_linking_codes.prune_expired(now);
+
+    // Check if we have a valid code available, and return it if there is one!
+    // In this situation a valid code would be the one which is still valid for
+    // more than 60 seconds; otherwise user may have too little time to enter
+    // the code on their other device.
+    let existing_linking_code = state.data.account_linking_codes.get_by_user_id(&user_id);
+    if let Some(code) = existing_linking_code {
```
hpeebles, line 30: I think the code returned should always be valid for 5 minutes (or whatever duration we choose), so that we can always say to the user that the code is valid for 5 minutes.
If we return pre-existing codes then the timing is no longer consistent, so I think it would be better to simply always generate a new code.

ivan-jukic: The idea was to show a timer within the UI with the code expiry.

backend/libraries/types/src/account_linking_code.rs
hpeebles: Ah right I see, so `AccountLinkingCode` could move to identity/api and `AccountLinkingCodes` (which contains much more logic and dependencies) could move to identity/impl/src/model.

backend/canisters/identity/impl/src/updates/create_account_linking_code.rs
hpeebles: Ah ok yeah that makes sense

backend/libraries/types/src/account_linking_code.rs
ivan-jukic: Gotcha, changes pushed.

backend/canisters/identity/api/src/model/account_linking_code.rs
```diff
@@ -0,0 +1,32 @@
...
+use ts_export::ts_export;
+use types::UserId;
+
+const ALC_DURATION: u64 = 5 * 60 * 1000; // 5 minutes in milliseconds
+
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+#[ts_export]
+pub struct AccountLinkingCode {
+    pub value: String,
+    pub expires_at: u64, // timestamp in milliseconds
```
hpeebles: You should use the `TimestampMillis` type to convey that the value is in milliseconds

backend/canisters/identity/api/src/model/account_linking_code.rs
```diff
@@ -0,0 +1,32 @@
...
+            expires_at: now + ALC_DURATION,
+            user_id,
+        }
+    }
+
+    pub fn is_valid(&self, now: u64) -> bool {
+        now < self.expires_at
+    }
+
+    pub fn is_valid_for_more_than_a_minute(&self, now: u64) -> bool {
```
hpeebles: I think this is redundant, you can just use `is_valid(now + MINUTE_IN_MS)`

backend/canisters/identity/impl/src/model/account_linking_codes.rs
```diff
@@ -0,0 +1,76 @@
...
+const ALC_CHARSET: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+#[derive(CandidType, Serialize, Deserialize, Default, Debug)]
+pub struct AccountLinkingCodes {
+    codes: HashMap<String, AccountLinkingCode>,
+}
+
+impl AccountLinkingCodes {
+    /// Creates a new code, saves it, and returns a clone back!
+    pub fn get_new_linking_code(&mut self, user_id: UserId, rng: &mut StdRng, now: u64) -> AccountLinkingCode {
```
hpeebles: `TimestampMillis`

backend/canisters/identity/impl/src/model/account_linking_codes.rs
```diff
@@ -0,0 +1,76 @@
...
+    // pub fn is_empty(&self) -> bool {
+    //     self.codes.is_empty()
+    // }
+
+    // Generates a random 6 character string.
+    fn generate_code(rng: &mut StdRng) -> String {
+        let bytes: [u8; ALC_LENGTH] = rng.r#gen();
+
+        // Map bytes to characters
+        bytes[..ALC_LENGTH]
```
hpeebles: You no longer need the range

backend/integration_tests/src/client/identity.rs
```diff
@@ -15,6 +14,8 @@ generate_msgpack_update_call!(initiate_identity_link);
 generate_msgpack_update_call!(initiate_identity_link_via_qr_code);
 generate_msgpack_update_call!(prepare_delegation);
 generate_msgpack_update_call!(remove_identity_link);
+generate_msgpack_update_call!(create_account_linking_code);
```
hpeebles: Alphabetical

backend/canisters/identity/impl/src/model/account_linking_codes.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /// Get number of saved codes
+    pub fn len(&self) -> usize {
+        self.codes.len()
+    }
+
+    // Generates a random 6 character string.
+    fn generate_code(rng: &mut StdRng) -> String {
+        let bytes: [u8; ALC_LENGTH] = rng.r#gen();
```
hpeebles: I think each value should be generated from an input larger than 1 byte, so that each char is more uniformly distributed.
There are 62 options, so 62 * 4 = 248, which means the first 8 chars are each chosen 5/256 times, whereas the remaining chars are chosen 4/256 times.
If we used u32 rather than u8 then the output would be much closer to uniformly random.
Another option would be to generate numbers between 0 and 61.

hpeebles: Ah you can do this using `rng.gen_range(..)`

ivan-jukic: Yeah, took a deeper look into the rng and found it, seemed actually nicer, and a bit shorter code.

PR #8376, opened 2025-07-22
Limit number of recipients of video call WebRTC messages

hpeebles:


julianjelfs (commented):
It would be nice to fallback to notifying the call participants if the call is _not_ for the selected chat as it's probably better than doing nothing.

hpeebles:
From my quick glance it seems as though we only store the call participants in the `ActiveCallParticipants.svelte` component.
Let's talk through how we can make use of that participants list when ending calls tomorrow.

hpeebles:
Maybe we move the active call participants into a store in the client.
Then when you leave a call all the logic can be handled by the client.
As in, the client can detect if the user was the last person in the call and if so, end the call.

hpeebles:
Also, I've just realised that we only end the call if you are the last participant, so "notifying the call participants" would mean notifying no-one (unless we notify all the participants who took part in the call).

julianjelfs:
> Maybe we move the active call participants into a store in the client. Then when you leave a call all the logic can be handled by the client. As in, the client can detect if the user was the last person in the call and if so, end the call.

But ... this is already the case. The `activeVideoCall` is a store and it contains the `DailyCall` and when we hangup it _does_ check if we are the last person in the call.

julianjelfs (approved):

PR #8380, opened 2025-07-23
Support range headers when accessing files

hpeebles:


backend/canisters/storage_bucket/impl/src/queries/http_request.rs
```diff
@@ -69,7 +69,7 @@ fn start_streaming_file(file_id: FileId, headers: &[(String, String)], state: &R
             };
 
             return HttpResponse {
-                status_code: 200,
+                status_code: if partial { 206 } else { 200 },
```
julianjelfs: As far as I understand it a 206 Partial Response is not enough. We also need to supply a Content-Range response header in this case e.g. 

Content-Range: bytes 0-123456/123457

julianjelfs (commented):
Also I think the browser will _always_ have an initial Range header set to `bytes=0-` then it is up to the server how it partially fulfils this request. To get the most benefit we need to actually return the video in chunks and add the correct Content-Range response header.

backend/canisters/storage_bucket/impl/src/queries/http_request.rs
```diff
@@ -44,38 +46,91 @@ fn http_request_streaming_callback(token: Token) -> StreamingCallbackHttpRespons
...
-                    callback: CallbackFunc::new(canister_id, "http_request_streaming_callback".to_string()),
-                    token: build_token(file_id, 1),
-                })
+        if let Some(file_bytes) = state.data.files.blob_bytes(&file.hash) {
+            let file_bytes_len = file_bytes.len();
+            let mut response_headers = vec![
+                HeaderField("Content-Type".to_string(), file.mime_type.clone()),
+                HeaderField("Cache-Control".to_string(), CACHE_HEADER_VALUE.to_string()),
+                HeaderField("X-Cacheable-Resource".to_string(), "true".to_string()),
+                HeaderField("Access-Control-Allow-Origin".to_string(), "*".to_string()),
```
ivan-jukic, line 57: Should this be a concern?

hpeebles: Hmm not sure, but that has been in place for years, so let's not let it block this PR

PR #8387, opened 2025-07-24
Show accout linking code in UI

ivan-jukic:
Allows account linking codes to be fetched and shown in the UI. Feature enabled only for dev and test.

image

image

image


frontend/app/src/components/home/profile/AccountLinkingCode.svelte
```diff
@@ -0,0 +1,220 @@
...
+
+    const client = getContext<OpenChat>("client");
+
+    type ModalStatus = "closed" | "initialLoad" | "open";
+    type CodeStatus = "notRequested" | "success" | "error";
+
+    let modalStatus = $state<ModalStatus>("closed");
+    let codeStatus = $state<CodeStatus>("notRequested")
+    let loadingCode = $state(false);
+    let remaining = $state(undefined);
```
julianjelfs: `$state()` is the same as `$state(undefined)`

frontend/app/src/components/home/profile/AccountLinkingCode.svelte
```diff
@@ -0,0 +1,220 @@
...
+                remainingInterval = setInterval(calculateRemaining, 1000);
+            }).catch((e) => {
+                codeStatus = "error";
+            }).finally(() => {
+                modalStatus = "open";
+                loadingCode = false;
+            });
+    }
+
+    function calculateRemaining() {
```
julianjelfs: I think if you have a look in something like the PrizeContent component. It has a similar type of countdown and makes use of some stores and OC client formatting functions which I think you can probably re-use

frontend/app/src/components/home/profile/AccountLinkingCode.svelte
```diff
@@ -0,0 +1,220 @@
...
+    <p class="para">
+        <Translatable resourceKey={i18nKey("accountLinkingCode.settingsMenu.disclaimer")} />
+    </p>
+    <Button onClick={showModal} loading={"initialLoad" === modalStatus}>
+        <Translatable resourceKey={i18nKey("accountLinkingCode.settingsMenu.cta")} />
+    </Button>
+</div>
+
+<style lang="scss">
+    :global(.modal-content) {
```
julianjelfs: Don't think that we can do this because it will affect _all_ instances of `.modal-content` - you will need to scope it to avoid that.

frontend/app/src/i18n/en.json
```diff
@@ -108,6 +108,25 @@
...
+            "title": "Link with new device",
+            "disclaimer": "If you would like to link this account to a device with the OpenChat native app installed, click on the button below to get your Account Linking Code! Use the code on a trusted device only.",
+            "cta": "Show account linking code"
+        },
+        "webModal": {
+            "title": "Account Linking Code",
+            "subtitle": "Use the code below to link your current account with a new trusted device. The code is case sensitive.",
+            "expired": "Your code has expired!",
+            "willExpireIn": "The code will expire in {remaining}",
+            "close": "Thanks, close the modal",
```
julianjelfs: Do people understand the word "modal"? I would have thought it's ok to just say "Close".

frontend/app/src/i18n/en.json
```diff
@@ -108,6 +108,25 @@
     "accountCopiedToClipboard": "Account address copied to clipboard!",
     "accounts": "Accounts",
     "accountSuspended": "Account suspended",
+    "accountLinkingCode": {
```
julianjelfs, line 109: You'll need to run `npm run sync_lang` from `/frontend/app` to generate the other languages.

julianjelfs (changes requested):

frontend/app/src/i18n/en.json
ivan-jukic: Of course, if we agree on the current copy, I'm happy to run it.

frontend/app/src/components/home/profile/AccountLinkingCode.svelte
ivan-jukic: There seems to be some weirdness when I refresh the page, and open the code modal, the initial remaining time will always be about 4 seconds higher than it should be, and in the next tick it will skip those 4 seconds to the actual time, e.g. initial render shows 1:19 remaining, and in the next second it's 1:15, and then it starts counting down correctly.

julianjelfs: I've pulled this locally to have a look and I'm not seeing any such behaviour. Seems to be working exactly as expected for me 🤷

frontend/app/src/components/home/profile/AccountLinkingCode.svelte
```diff
@@ -139,42 +128,41 @@
...
     .subtitle {
         @include font(book, normal, fs-100);
         padding-top: $sp4;
         padding-bottom: $sp4;
         color: var(--txt-light);
     }
 
-    .header, .footer, .error {
+    .header, .footer, .code-content {
+        width: 28rem;
```
julianjelfs: this might be asking for trouble on mobile where it should probably be unrestricted?

frontend/app/src/components/home/profile/AccountLinkingCode.svelte
ivan-jukic: https://github.com/user-attachments/assets/0489a987-dff8-4a40-b191-7a42f3f6b811

You can see it on 15+ sec in the video, seconds jump from 47 to 42 in the countdown. This is after a page reload.

ivan-jukic: Also what I've seen locally; the code itself lasts 3 min, but the countdown would start at 03:03, and then suddenly jump to 02:58.

ivan-jukic: I've pushed an update to the time store that seems to fix the issue I'm seeing.

julianjelfs (approved):

PR #8389, opened 2025-07-24
Use full names for bot event notification fields

megrogan:


megrogan:
> Can we get these types into the typebox.ts file as well? Would make life easier because then I can use the typebox validators rather than hand rolling the parsing.

Ok done

PR #8406, opened 2025-07-29
Allow platform operators to reinstate missed daily claims

hpeebles:


backend/canisters/user/impl/src/lib.rs
```diff
@@ -353,6 +353,36 @@ Your streak is now {new_streak} days and you have {days_remaining_text} of strea
...
+            .streak
+            .reinstate_missed_daily_claims(days_to_reinstate, daily_claims, now);
+
+        let count = new_events.len();
+        for event in new_events {
+            self.data.chit_events.push(event);
+        }
+        let new_streak = self.data.streak.days(now);
+
+        let first_line = if count == 1 {
```
ivan-jukic, line 372: This seems a bit "weird", but it's not a big deal. Perhaps at some point we could add some utils/extras to deal with plurals and such.

hpeebles: Definitely open to suggestions for making it better but this seems like the simplest and easiest to understand way for now.

hpeebles: I could make the if check only control adding the "s" after claim and toggling between "s" and "ve", but that seemed like overkill.

ivan-jukic (approved):

backend/canisters/user/impl/src/lib.rs
ivan-jukic: Really no need. Next time if something similar pops up.

PR #8411, opened 2025-07-30
Split linking with account linking code into two steps

ivan-jukic:
The first step is to verify the code, and fetch the user's username, which will then be used to label the passkey. Second step is to finalise the linking.

This also streamlines the process a bit, code is verified in the first step, and added to the verified codes map and attached to the temp caller principal. In the second step we still use the temp caller principal to fetch the verified code, and finalise the linking.


backend/canisters/identity/impl/src/updates/finalise_account_linking_with_code.rs
```diff
@@ -71,15 +62,9 @@ fn link_with_account_linking_code_impl(args: Args, state: &mut RuntimeState) ->
 
     // Check that the target user principal has not reached the maximum number of linked identities
     if user_principal.auth_principals.len() >= MAX_LINKED_IDENTITIES {
-        // return LinkedIdentitiesLimitReached(MAX_LINKED_IDENTITIES as u32);
         return Err(OCErrorCode::UserLimitReached.into());
```
hpeebles: UserLimitReached isn't quite right, we may need a new error code.

backend/integration_tests/src/account_linking_tests.rs
```diff
@@ -3,7 +3,8 @@ use crate::env::ENV;
 use crate::utils::now_millis;
 use crate::{TestEnv, client};
 use identity_canister::create_account_linking_code::Response as CreateAccountLinkingCodeResponse;
-use identity_canister::link_with_account_linking_code::Response as LinkWithAccountLinkingCodeResponse;
+use identity_canister::finalise_account_linking_with_code::Response as FinaliseAccountLinkingWithCodeResponse;
+use identity_canister::verify_account_linking_code::Response as VerifyAccountLinkingCodeResponse;
 use std::ops::Deref;
 use testing::rng::random_delegated_principal;
```
hpeebles, line 9: On line 17 you should use `random_principal` to mimic the random temp key that each user will generate.

backend/integration_tests/src/account_linking_tests.rs
```diff
@@ -58,22 +59,42 @@ fn test_account_linking_create_link_code() {
     let (new_principal, new_pub_key) = random_delegated_principal(canister_ids.sign_in_with_email);
```
hpeebles: I think use `random_principal` to generate the temp key and use `random_internet_identity_principal` to generate the new key that we'll link the account to.

backend/integration_tests/src/account_linking_tests.rs
```diff
@@ -58,22 +59,42 @@ fn test_account_linking_create_link_code() {
...
         canister_ids.identity,
-        &identity_canister::link_with_account_linking_code::Args {
+        &identity_canister::verify_account_linking_code::Args {
             code: created_linking_code.value.clone(),
+        },
+    );
+
+    match verify_response {
+        VerifyAccountLinkingCodeResponse::Success(username) => {
+            assert_eq!(username, "72awj".to_string());
```
hpeebles: You can grab the username from `user.username`

PR #8422, opened 2025-07-31
Total chit earned access gate

megrogan:
Upgrade order:

LocalUserIndex
UserIndex
GroupIndex
Community
Group
Website (incl associated FE changes)


backend/canisters/user_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -28,4 +29,29 @@ fn post_upgrade(args: Args) {
...
+            .users
+            .iter()
+            .filter_map(|user| {
+                let total_chit_earned = user.total_chit_earned();
+                if total_chit_earned != 0 { Some((user.user_id, total_chit_earned)) } else { None }
+            })
+            .collect();
+
+        for (user_id, total_chit_earned) in users_to_sync {
+            state.push_event_to_all_local_user_indexes(
```
hpeebles: This will attempt to push the messages immediately, which you can't do in post upgrade.
I think just copy the code here and remove the bit that triggers the push.
Then the job will just start as normal at the end of the upgrade.

backend/libraries/group_community_common/src/user_cache.rs
```diff
@@ -17,15 +17,24 @@ pub struct CachedUser {
     pub diamond_membership_expires_at: Option<TimestampMillis>,
     #[serde(rename = "u", alias = "is_unique_person", default, skip_serializing_if = "is_default")]
     pub is_unique_person: bool,
+    #[serde(rename = "c", alias = "chit_earned", default, skip_serializing_if = "is_default")]
```
hpeebles: No need for this alias, and we can remove the other ones, they were just needed when the shorter names were first introduced.

hpeebles (approved):

PR #8425, opened 2025-08-01
Account linking with code in app

ivan-jukic:
Added the UI code to allow account linking from within an android app.


frontend/app/src/i18n/en.json
```diff
@@ -1109,12 +1109,20 @@
     "native": {
         "auth": {
             "createAccount": "Create new account",
-            "error": "Authentication failed. Please try again.",
+            "errors": {
+                "alreadyRegistered": "Auth method has already been linked. Try signing in!",
+                "codeInvalid": "Your account linking code seems to be invalid!",
+                "cretePasskeyFail": "Passkey creation failed, please try again.",
```
hpeebles: create*

frontend/app/src/i18n/en.json
```diff
@@ -1109,12 +1109,20 @@
     "native": {
         "auth": {
             "createAccount": "Create new account",
-            "error": "Authentication failed. Please try again.",
+            "errors": {
+                "alreadyRegistered": "Auth method has already been linked. Try signing in!",
```
hpeebles, line 1116: I think some of this text can be improved but happy to approve as is and we can fine tune the text later

frontend/openchat-agent/src/services/identity/mappers.ts
```diff
@@ -227,7 +225,43 @@ export function apiWebAuthnKey(key: WebAuthnKeyFull): IdentityWebAuthnKey {
...
+            number,
+            number,
+            number,
+            number,
+            number,
         ],
     };
 }
+
+export function mapAccountLinkingErrorCode(code: number): AccountLinkingErrorCode {
```
hpeebles: You should add these to the `ErrorCode` enum

frontend/openchat-agent/src/services/identity/identity.client.ts
```diff
@@ -233,6 +239,42 @@ export class IdentityClient extends MsgpackCanisterAgent {
         );
     }
 
+    verifyAccountLinkingCode(code: string): Promise<VerifyAccountLinkingCodeResponse> {
+        return this.executeMsgpackUpdate(
+            "verify_account_linking_code",
+            { code },
+            (resp) => {
```
hpeebles: You can use `unitResult` then handle the various error codes in the svelte component using the `ErrorCode` enum

frontend/app/src/i18n/en.json
ivan-jukic: Really up to you, this is just the first draft, users won't see this until the app is out.

frontend/openchat-agent/src/services/identity/identity.client.ts
ivan-jukic: TS actually does not like this, since the return is Success object, and the shape of my data is different.

frontend/app/src/i18n/en.json
hpeebles: Ah no I was meaning "create" in the key name, rather than "crete". The value itself was fine.

frontend/openchat-agent/src/services/identity/identity.client.ts
hpeebles: Right I see you're returning data too, that means you should be able to use `mapResult`

frontend/openchat-shared/src/domain/identity.ts
```diff
@@ -150,19 +151,6 @@ export type FinaliseAccountLinkingResponse = { kind: "success" } | AccountLinkin
 
 export type AccountLinkingError = {
```
hpeebles: Is this still used? I think you can delete it. It is exactly the same as `OCError`.

frontend/openchat-agent/src/services/identity/identity.client.ts
ivan-jukic: Ok, but why? I don't really need to map the data to anything. `mapResult` just takes another map function arg.

hpeebles: Just means you only need to map the success case, because the error case is now the same for all functions (that use the new format)

ivan-jukic: I suppose my question is, why map anything?

I'm not changing the shape of my data at all. It's actually just handled in the identityAgent.ts to create delegation, which is then used within the worker.ts. Top level code never really handles any of this, just gets back success info.

hpeebles: Yeah if we could consolidate our types between backend and frontend that would be a huge benefit.
But for now the backend returns enums where the variant is the field name, whereas the frontend uses a record type where the `kind` field specifies the variant.

hpeebles: We originally used candid on the backend which used a quite weird structure for some types and we had no control over that so we never attempted to consolidate the types back then. Now that we control the serialization on both ends its probably worth revisiting this (not right now though) to see if we can remove the mapping code.

hpeebles: Something like `(resp) => mapResult(resp, (username) => ({ kind: "success", username })),`

hpeebles: Maybe for new endpoints we should attempt to make the backend return exactly the format the frontend uses.

ivan-jukic: I suppose that would make more sense, let's try and do that if possible at all. It would remove this translation layer. Anyway, updated according to your suggestion.

hpeebles (approved):

PR #8431, opened 2025-08-04
Invert the map of blocked users

hpeebles:


backend/canisters/user_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -17,9 +19,35 @@ fn post_upgrade(args: Args) {
...
     let reader = get_reader(&memory);
 
-    let (data, errors, logs, traces): (Data, Vec<LogEntry>, Vec<LogEntry>, Vec<LogEntry>) =
+    let (mut data, errors, logs, traces): (Data, Vec<LogEntry>, Vec<LogEntry>, Vec<LogEntry>) =
         msgpack::deserialize(reader).unwrap();
 
+    let blocked_users = data.blocked_users.collect_all();
+    let blocked_users_len = data.blocked_users.len();
+
+    let mut blocked_user_pairs = Vec::new();
```
ivan-jukic, line 28: I'm being nitpicky, but perhaps it just makes a bit more sense to move this var down just before the next for loop where it's used.

backend/canisters/user_index/impl/src/lifecycle/post_upgrade.rs
```diff
@@ -17,9 +19,35 @@ fn post_upgrade(args: Args) {
     let memory = get_upgrades_memory();
     let reader = get_reader(&memory);
 
-    let (data, errors, logs, traces): (Data, Vec<LogEntry>, Vec<LogEntry>, Vec<LogEntry>) =
+    let (mut data, errors, logs, traces): (Data, Vec<LogEntry>, Vec<LogEntry>, Vec<LogEntry>) =
         msgpack::deserialize(reader).unwrap();
 
+    let blocked_users = data.blocked_users.collect_all();
```
ivan-jukic, line 25: Could we improve naming here a bit, or in the for loops? We're using `blocked_users` var name here, and then again as a local var within the for loops.

PR #8437, opened 2025-08-05
Initialise `total_chit_earned` from the monthly CHIT amounts

hpeebles:


backend/canisters/user_index/impl/src/model/user.rs
```diff
@@ -17,29 +17,29 @@ pub struct User {
     pub user_id: UserId,
     #[serde(rename = "un")]
     pub username: String,
-    #[serde(rename = "dn", default, skip_serializing_if = "Option::is_none")]
```
hpeebles, line 20: `default` isn't needed on `Option`s

julianjelfs (approved):

PR #8439, opened 2025-08-05
Implement ability to pay for premium items using CHIT

hpeebles:


backend/canisters/user_index/impl/src/updates/pay_for_premium_item.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    let now = state.env.now();
+    user.pay_for_premium_item(args.item_id, cost, now);
+    state
+        .data
+        .premium_items
+        .log_purchase(user.user_id, args.item_id, false, cost, now);
+
+    Ok(())
```
julianjelfs: Might be convenient to return the new balance?

PR #8444, opened 2025-08-06
Send strongly typed notifications to LocalUserIndex

hpeebles:


backend/canisters/community/impl/src/lib.rs
```diff
@@ -128,14 +126,13 @@ impl RuntimeState {
...
-        fcm_data: FcmData,
     ) {
         if !recipients.is_empty() {
             let notification = Notification::User(UserNotification {
                 sender,
                 recipients,
-                notification_bytes: ByteBuf::from(serialize_then_unwrap(notification)),
-                fcm_data: Some(fcm_data),
+                notification: Some(notification),
+                notification_bytes: None,
```
ivan-jukic: I guess we have to keep this for backwards compatibility? Perhaps might be worth leaving a comment if that's the case?

backend/canisters/group/impl/src/lib.rs
```diff
@@ -129,14 +128,13 @@ impl RuntimeState {
...
-        fcm_data: FcmData,
     ) {
         if !recipients.is_empty() {
             let notification = Notification::User(UserNotification {
                 sender,
                 recipients,
-                notification_bytes: ByteBuf::from(serialize_then_unwrap(notification)),
-                fcm_data: Some(fcm_data),
+                notification: Some(notification),
+                notification_bytes: None,
```
ivan-jukic: Same here?

backend/canisters/user/impl/src/lib.rs
```diff
@@ -116,21 +114,15 @@ impl RuntimeState {
...
         self.data.local_user_index_event_sync_queue.push(IdempotentEnvelope {
             created_at: self.env.now(),
             idempotency_id: self.env.rng().next_u64(),
             value: local_user_index_canister::UserEvent::Notification(Box::new(Notification::User(UserNotification {
                 sender,
                 recipients: vec![recipient],
-                notification_bytes: ByteBuf::from(serialize_then_unwrap(notification)),
-                fcm_data: Some(fcm_data),
+                notification: Some(notification),
+                notification_bytes: None,
```
ivan-jukic: ...

backend/libraries/types/src/fcm_data.rs
```diff
@@ -26,7 +26,7 @@ impl FcmData {
     fn default(chat_id: Chat) -> Self {
         Self {
             chat_id,
-            thread_id: None,
+            thread_root_message_index: None,
```
ivan-jukic, line 29: Ah, thanks for the correction here! I remember having some doubts about this one, but amidst everything forgot to follow up on it.

backend/libraries/types/src/notifications.rs
```diff
@@ -221,6 +219,38 @@ pub enum UserNotificationPayload {
...
+                .set_optional_image(n.image_url)
+                .set_avatar_id(n.group_avatar_id),
+            UserNotificationPayload::ChannelMessage(n) => FcmData::for_channel(n.community_id, n.channel_id)
+                .set_sender_id(n.sender)
+                .set_sender_name_with_alt(n.sender_display_name, n.sender_name)
+                .set_optional_thread(n.thread_root_message_index)
+                .set_body_with_alt(n.message_text, n.message_type)
+                .set_optional_image(n.image_url)
+                .set_avatar_id(n.channel_avatar_id.or(n.community_avatar_id)),
+            _ => FcmData::for_group(Principal::anonymous().into()),
```
ivan-jukic: I feel like there should be more here, no?

backend/canisters/community/impl/src/lib.rs
hpeebles: I tried to handle deserializing from the old and new types but serde untagged doesn't work with u128 values.
So for this changeover both the old and new fields are defined.
Once everything is updated we'll remove the old field and set the new one to required rather than optional.

backend/libraries/types/src/notifications.rs
hpeebles: Yup, just sorting those now, but wanted to get these ones in to show the general idea

backend/canisters/community/impl/src/lib.rs
ivan-jukic: Sounds good. I'd personally just add a comment here (and in couple other places) to not lose it out of sight. But, if you prefer keeping it all in your head, feel free to resolve the comment 😅

PR #8493, opened 2025-08-15
Misc fixes mostly around new CHIT features

julianjelfs:
* make sure instagram previews don't overflow
* stream & cache public profiles so that they load better in the front end
* use the intersection observer to avoid unnecessary loading of custom emojis
* fix the dirty checking on the user profile (specifically bio)
* tweak the chit badge sizing
* slightly improve the parsing of notification messages that contain custom emojis

### DIFF
--- a/backend/canisters/local_user_index/impl/src/lib.rs
+++ b/backend/canisters/local_user_index/impl/src/lib.rs
@@ -595,12 +595,13 @@ impl Data {
                     .collect();
 
                 if !filtered_recipients.is_empty() {
-                    self.notifications.add(NotificationEnvelope::User(UserNotificationEnvelope {
-                        recipients: filtered_recipients,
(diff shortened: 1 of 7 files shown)


frontend/openchat-shared/src/domain/chit.ts
```diff
@@ -198,5 +198,5 @@ export enum PremiumItem {
 export const premiumPrices: Record<PremiumItem, number> = {
     [PremiumItem.BotEmojis]: 10_000,
     [PremiumItem.PopularEmojis]: 50_000,
-    [PremiumItem.CustomProfileBackground]: 50_000,
+    [PremiumItem.CustomProfileBackground]: 1_000,
```
megrogan: That seems cheap

julianjelfs: Ooh well spotted - that was me testing!

PR #8496, opened 2025-08-15
Support muting of @everyone notifications

megrogan:


backend/libraries/group_chat_core/src/members.rs
```diff
@@ -35,6 +35,7 @@ pub struct GroupMembers {
     moderators: BTreeSet<UserId>,
     bots: BTreeMap<UserId, UserType>,
     notifications_unmuted: BTreeSet<UserId>,
+    at_everyone_muted: BTreeSet<UserId>,
```
hpeebles, line 38: `serde(default)`

hpeebles: And let's get rid of the alias on line 30, we haven't needed that for ages.

backend/libraries/group_chat_core/src/members.rs
```diff
@@ -803,6 +850,12 @@ pub struct GroupMemberStableStorage {
         skip_serializing_if = "is_default_notifications_muted"
     )]
     notifications_muted: Timestamped<bool>,
+    #[serde(
+        rename = "em",
+        default = "default_at_everyone_muted",
+        skip_serializing_if = "is_default_at_everyone_muted"
```
hpeebles: You should be able to use "is_default"

backend/libraries/group_chat_core/src/members.rs
megrogan: Oops

hpeebles (approved):

PR #8498, opened 2025-08-15
improve the behaviour when exiting preview

julianjelfs:


frontend/app/src/components/home/PreviewFooter.svelte
```diff
@@ -47,13 +49,18 @@
...
         if (previewingCommunity && $selectedCommunitySummaryStore) {
             client.removeCommunity($selectedCommunitySummaryStore.id);
+            page(routeForScope(client.getDefaultScope()));
         } else {
             if (!chat.public) {
                 client.declineInvitation(chat.id);
             }
-            client.removeChat(chat.id);
+            client.removePreviewedChat(chat.id);
+            if ($mobileWidth) {
```
megrogan: Don't need an else if

frontend/openchat-client/src/openchat.ts
```diff
@@ -8068,8 +8068,8 @@ export class OpenChat {
         return false;
     }
 
-    selectDefaultChat(): boolean {
-        if (!get(mobileWidth)) {
+    selectDefaultChat(desktopOnly: boolean = true): boolean {
```
megrogan, line 8071: So this is always true?

frontend/app/src/components/home/PreviewFooter.svelte
julianjelfs: I don't understand why you think that

frontend/openchat-client/src/openchat.ts
julianjelfs: uhm no. Sometimes it's false.

PR #8503, opened 2025-08-18
Consolidate `AccountInfo` components

hpeebles:
The BitcoinAccountInfo component was exactly the same as AccountInfo but with a few additional elements.
So I've merged the 2 together and wrapped the additional elements in `if` conditions.


frontend/app/src/components/home/AccountInfo.svelte
```diff
@@ -26,19 +34,96 @@
...
+        }
+    });
+
+    const btcAddressPromise = new Lazy(() => client.getBtcAddress());
+
+    let error = $state();
+    $effect(() => {
+        if (account === undefined) {
+            if (selectedNetwork === BTC_SYMBOL) {
+                btcAddressPromise.get().catch((e) => error = e);
```
julianjelfs: This seems a bit odd - getting the btc address and then doing ... nothing with it. Feels fishy.

hpeebles: When the promise completes it'll set the `btcAddress` store which `account` is derived from (if the BTC network is still selected).

hpeebles: On line 56

julianjelfs: Yes - I understand. But it's hard to like. Feels quite messy but I'm not sure how to make it less so.

hpeebles: I could make `account` derive from a local bitcoin store which is set after the promise completes, but that just feels worse than having the single source of truth

julianjelfs: It's not the source of truth so much as the control flow which is about as confusing as could be. We have a thing called `btcAddressPromise` which is _not_ a promise. Then we call `get` on it but we do nothing with the response. It's quite weird. Feels like we could just call `client.getBtcAddress` when required and put the result in a `$state` and the whole thing would be a lot less confusing.

hpeebles: Yeah makes sense, have just updated it, see what you think of the new version

frontend/app/src/components/home/AccountInfo.svelte
```diff
@@ -26,19 +33,90 @@
...
+    });
+
+    // We use `Lazy` to ensure this is only called once
+    const btcAddressPromise = new Lazy(() => client.getBtcAddress());
+
+    let error = $state();
+    $effect(() => {
+        if (account === undefined) {
+            if (selectedNetwork === BTC_SYMBOL) {
+                btcAddressPromise.get().then((addr) => btcAddress = addr).catch((e) => error = e);
```
julianjelfs: Isn't it fine to just call `client.getBtcAddress` directly here and not use `Lazy` at all? because internally `client.getBtcAddress` already deals with not loading the address more than once.

hpeebles: It's quite a slow call as it goes User canister -> ckBTC canister (on another subnet).
So if you switch tokens back and forth it would call it twice.

julianjelfs: All the more reason to deal with it inside the OC client?

julianjelfs (approved):

PR #8504, opened 2025-08-18
Handle custom emojis with long names

hpeebles:


backend/libraries/types/src/reactions.rs
```diff
@@ -15,8 +13,8 @@ impl Reaction {
 
     pub fn is_valid(&self) -> bool {
         let len = self.0.len();
-
-        (1..=MAX_REACTION_LENGTH_BYTES).contains(&len)
+        let max_length = if self.0.starts_with("@CustomEmoji(") { 100 } else { 40 };
```
julianjelfs: or `@CE`

hpeebles: I've just switched it to CE, since the website can go out before the backend changes

julianjelfs: OK yes that works since this is only checked when adding the reaction

PR #8509, opened 2025-08-18
Allow muting of @everyone notifications

megrogan:
The website must be upgraded with [this commit](https://github.com/open-chat-labs/open-chat/commit/1a41e08afb427e8684bfc2d808149a24e5506216) before this PR is merged - then LUIs/groups/communities need to be upgraded before the website is upgraded with the changes in this PR.


julianjelfs (commented):
There's something a bit unpleasant about the types here isn't there. We are _always_ doing exactly one thing (mute, unmute, mute everyone, unmute everyone) but all of the interfaces imply that we can do 0, 1 or 2 things. 

I wonder if we could hide that unpleasantness until we arrive at the bottom layer.

megrogan:
> There's something a bit unpleasant about the types here isn't there. We are _always_ doing exactly one thing (mute, unmute, mute everyone, unmute everyone) but all of the interfaces imply that we can do 0, 1 or 2 things.
> 
> I wonder if we could hide that unpleasantness until we arrive at the bottom layer.

I considered that but we also might want to do multiple things with a slightly different UI. Perhaps the clicking a single notifications menu item opens a dialog with options that can all be changed in one go.

julianjelfs:
> > There's something a bit unpleasant about the types here isn't there. We are _always_ doing exactly one thing (mute, unmute, mute everyone, unmute everyone) but all of the interfaces imply that we can do 0, 1 or 2 things.
> > I wonder if we could hide that unpleasantness until we arrive at the bottom layer.
> 
> I considered that but we also might want to do multiple things with a slightly different UI. Perhaps clicking a single notifications menu item opens a dialog with options that can all be changed in one go. It makes sense to me that the openchat-client supports this.

So what does it mean to set both things to undefined?

megrogan:
> > > There's something a bit unpleasant about the types here isn't there. We are _always_ doing exactly one thing (mute, unmute, mute everyone, unmute everyone) but all of the interfaces imply that we can do 0, 1 or 2 things.
> > > I wonder if we could hide that unpleasantness until we arrive at the bottom layer.
> > 
> > 
> > I considered that but we also might want to do multiple things with a slightly different UI. Perhaps clicking a single notifications menu item opens a dialog with options that can all be changed in one go. It makes sense to me that the openchat-client supports this.
> 
> So what does it mean to set both things to undefined?

Acknowledged, but all other combinations make sense and I don't think the alternative will be nicer.

PR #8532, opened 2025-08-26
UI component library 

julianjelfs:
This is a first stab at organising a new library of components meant for the native app re-work. 

Essentially this is a new project within the front end. It exposes components directly so there is no build step (the `app` project's build process will deal with it). 

It also contains a small vite site to serve as a test harness so that the components can be developed in isolation. This serves the purpose of a much simpler storybook. Storybook itself is a) awful and b) doesn't work with Svelte 5 (note that it _does_ claim to work - but has many problems). 

I have just filled in the button component so far just to make sure that everything hangs together. 

image

### DIFF
--- a/frontend/app/src/components/home/thread/ThreadPreview.svelte
+++ b/frontend/app/src/components/home/thread/ThreadPreview.svelte
@@ -20,7 +20,7 @@
     import page from "page";
     import { getContext, onMount } from "svelte";
     import { _ } from "svelte-i18n";
-    import EyeOffIcon from "svelte-material-icons/EyeOff.svelte";
+    import EyeOffIcon from "svelte-material-icons/EyeOffOutline.svelte";
     import { i18nKey } from "../../../i18n/i18n";
     import { pop } from "../../../utils/transition";
     import Avatar from "../../Avatar.svelte";


frontend/component-lib/src/components/Button.svelte
```diff
@@ -0,0 +1,88 @@
...
+        children,
+        icon,
+        disabled = false,
+        onClick,
+        loading = false,
+        secondary = false,
+    }: Props = $props();
+
+    let spinnerColour = secondary ? "var(--primary)" : "var(--textOnPrimary)";
+    let iconColour = secondary ? "var(--primary)" : "var(--textOnPrimary)";
```
ivan-jukic, line 23: These two var are identical, perhaps we need only one? And without `var(--<insert var name here>)`?

julianjelfs: Yes - I thought about that. I guess I concluded that although they are the same _in this theme_ they are _logically_ different so I thought I would keep them different.

ivan-jukic: Yeah, that's a fair point, unless we want to restrict it and make it consistent across different themes. I'm still unsure how different we'd like our themes to be.

On one hand making them more streamlined will make them easier to produce; while on the other hand too much restrictions may prevent us from being a bit "free" with future themes.

PR #8540, opened 2025-08-28
GH Action to deploy Notification Pusher

ivan-jukic:
A bit of automation to make Notification Pusher easier to deploy!

The Action will:
- Get the latest NP version from tags
- Check that it's not already deployed, and create deploy location
- Set up Rust and necessary target
- Build binary
- Upload binary
- Update latest location and restart service
- Rollback if update fails

### DIFF
--- a/scripts/deploy-notification-pusher-prod.sh
+++ b/scripts/deploy-notification-pusher-prod.sh
@@ -7,72 +7,82 @@ EC2_HOST="$3"
 EC2_SSH_KEY_PATH="$4"
 
 remote_cmd() {
-    ssh -i "$EC2_SSH_KEY_PATH" -o StrictHostKeyChecking=no "$EC2_USER@$EC2_HOST" "$@" || {
-        echo "❌ SSH command failed: $*"
-        exit 1
-    }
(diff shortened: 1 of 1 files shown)


hpeebles:
As far as I can tell this isn't copying over the .env file.

But copying it isn't always what we want because it may have changed.

Maybe we avoid using a config file and read secrets from AWS Secret Manager or something like that?

ivan-jukic:
> As far as I can tell this isn't copying over the .env file.
> 
> But copying it isn't always what we want because it may have changed.
> 
> Maybe we avoid using a config file and read secrets from AWS Secret Manager or something like that?

@hpeebles yes, thanks for the reminder. I'll add a cp command to just copy over the .env from the previous version. I suppose it doesn't change often, and in those few cases where it does, we could manually edit it and restart the service (the action would fail if env changes are required, and rolled back, but the new uploaded version would remain).

On the other hand, we should be using AWS Secrets Manager for this kind of things. Haven't looked up our current setup yet, but it should be relatively straightforward; add secrets to AWS SM, attach IAM role to our EC2 instance that allows it to fetch the secrets, get the secrets on deploy and dump them into a file. Best if we do this separately from this PR.

megrogan:
Personally I would prefer a shell script like we use for all our other upgrades rather than using a GH action as a special case

megrogan (approved):

PR #8566, opened 2025-09-04
Bottom nav bar

julianjelfs:
image


frontend/component-lib/src/components/bottom_bar/BottomBarItem.svelte
```diff
@@ -0,0 +1,68 @@
...
+</Container>
+
+<style lang="scss">
+    :global(.bottom_bar_icon svg) {
+        width: var(--icon-lg);
+        height: var(--icon-lg);
+    }
+
+    .selection {
+        height: 4px;
```
ivan-jukic, line 45: I imagined we'd use rems insted of px, any reasons not to? 🤔

frontend/component-lib/src/components/bottom_bar/BottomBarItem.svelte
```diff
@@ -0,0 +1,68 @@
...
+<Container
+    onClick={onSelect}
+    borderWidth={"zero"}
+    borderColour={"red"}
+    borderStyle={"dashed"}
+    crossAxisAlignment={"center"}
+    mainAxisAlignment={"spaceBetween"}
+    height={{ kind: "fill" }}
+    direction={"vertical"}>
+    <div class="selection" class:selected></div>
```
ivan-jukic: If we'd want to animate this, it would need to be a child of the bottom bar. Until that time though I have no issues if it stays here.

julianjelfs: Yeah - I decided to leave that for future me / you

frontend/component-lib/src/components/bottom_bar/BottomBarItem.svelte
julianjelfs: I figured we would use pixels in a case like this where we probably _don't_ want it to change size if the font size changes? But maybe we do.

frontend/component-lib/src/components/bottom_bar/BottomBarItem.svelte
julianjelfs: Done

PR #8568, opened 2025-09-05
Add message link previews and re-style previews

megrogan:


frontend/app/src/components/home/LinkPreviews.svelte
```diff
@@ -246,6 +332,21 @@
 
         .inner {
             flex: 1;
+            border-left: $sp2 solid var(--currentChat-msg-separator);
+            padding-left: 12px;
+
+            &.rtl {
+                border-right: $sp2 solid var(--currentChat-msg-separator);
```
julianjelfs: might be able to use `border-block-start` and `padding-block-start` here to remove the need for special rtl handling. If _not_ then the &.rtl section also need to reset the ltr rules.

PR #8577, opened 2025-09-08
chat summary component

julianjelfs:
image


frontend/component-lib/src/components/LatestMessage.svelte
```diff
@@ -0,0 +1,51 @@
...
+    let textColour = $derived<TypographyColour>(
+        latestMessage.kind === "video_call" && latestMessage.inProgress ? "accent" : "secondary",
+    );
+</script>
+
+<Container gap={"xs"} crossAxisAlignment={"center"}>
+    {#if latestMessage.kind === "video_call"}
+        <Video color={iconColour} />
+    {:else if latestMessage.kind === "video"}
+        <Movie color={"var(--text-secondary"} />
```
ivan-jukic, line 42: Without digging too much, could we use string constants here, or enum for colours?

frontend/component-lib/src/components/avatar/MultiAvatar.svelte
```diff
@@ -0,0 +1,72 @@
...
+    <div class="second">
+        <Avatar url={second} size={"sm"} />
+    </div>
+    <div class="count">
+        <CountBadge size={"large"} mode={"additive"}>+{plus}</CountBadge>
+    </div>
+</div>
+
+<style lang="scss">
+    :global(.multi_avatars .second img) {
```
ivan-jukic, line 30: Nice one! I think in general having CSS in our component lib is to be expected. We do want to avoid any css in our app code, if at all possible.

frontend/component-lib/src/components/LatestMessage.svelte
julianjelfs: In general I would say _yes_ - but it's difficult to do here because these are the svelte-material-ui icons themselves so we don't control the interface and the color prop is just a string. On the plus side though, this `LatestMessage` component itself is doing the work so the choice of colour is never actually exposed in this case.

julianjelfs: But I guess we could still have some constants that at least helps us spit out the _right_ strings. e.g. 

`<Movie color={ColourVars.textSecondary} />`

wouldn't stop you doing

`<Movie color={"yellow"} />` 

but would probably help avoid errors a _bit_.

julianjelfs: But then, it's equally possible to use `yellow` instead of `var(--text-secondary)` in the css. So - not sure...

ivan-jukic: We do want to use variables because of theming, at leat in this case.  I guess I was just curious about using something that does not require me to remember all the var names because I won't. Having something like `ColourVars` would be easier since IDE would then just suggest what options I have, and I only need to remember a single thing.

julianjelfs: Well, in this case, since we are assigning the colour in the typescript, we _could_ actually pull it from the theme itself. Something like this:

```
    let iconColour = $derived(
        latestMessage.kind === "video_call" && latestMessage.inProgress
            ? theme.colours.primary
            : theme.colours.textSecondary,
    );
```

There are pros and cons to doing that though. On the plus side it's type safe and easy. On the down side we now have to make sure that is reactive (by storing the currently selected theme in a store or something). So using this method we end up injecting the actual colour into the icon component rather than the variable name. 

I have tried to avoid doing this since it seems clearer to _either_ use the css vars everywhere _or_ pull the variable values out of the theme everywhere. 

PS - I think you _will_ remember the css var names pretty quickly. 

I think I would prefer to leave as is for this PR and think about a nice way to get hold of the vars so we don't have to remember them.

PR #8605, opened 2025-09-11
List action component 

julianjelfs:
... and some rejigging of the colours 

image

image

### DIFF
--- a/backend/canisters/local_user_index/CHANGELOG.md
+++ b/backend/canisters/local_user_index/CHANGELOG.md
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fix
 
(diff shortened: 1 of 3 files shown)


frontend/component-lib/src/components/ListAction.svelte
```diff
@@ -0,0 +1,85 @@
+<script lang="ts">
```
ivan-jukic, line 1: I haven't really wrote this yet in the guide in Figma, but the green plus icon is optional, and in figma even its icon is configurable, though we might say that it's fairly unlikely that it will change.

I'd just add a prop saying `showSmallIcon: boolean`... and enable/disable it by default? In figma it's enabled simply to have that visual info that it exists.

julianjelfs: OK - will do

PR #8609, opened 2025-09-12
bring the big buttons into line with colour changes

julianjelfs:
image

Also made a decent start on message bubbles: 

image


frontend/component-lib/src/components/Container.svelte
```diff
@@ -70,11 +74,15 @@
         borderColour = "var(--background-2)",
```
ivan-jukic, line 74: I suppose one day we'd see something like `theme.background2.toCssVar()` 😅 
Feel free to resolve.

frontend/component-lib/src/components/message/Message.svelte
```diff
@@ -0,0 +1,48 @@
...
+        messageContent,
+        threadSummary,
+        reactions,
+        time,
+        edited = false,
+        avatarUrl,
+    }: Props = $props();
+</script>
+
+<Container gap={"sm"} allowOverflow mainAxisAlignment={me ? "end" : "start"}>
```
ivan-jukic, line 28: Yes, it's simple as that :)

frontend/component-test/src/Avatars.svelte
```diff
@@ -17,6 +17,7 @@
             <Avatar name={`large ${name}`} url={`/${file}`} size={"lg"} />
             <Avatar name={`medium ${name}`} url={`/${file}`} size={"md"} />
             <Avatar name={`small ${name}`} url={`/${file}`} size={"sm"} />
+            <Avatar name={`extra small ${name}`} url={`/${file}`} size={"xs"} />
```
ivan-jukic, line 20: Yes, I suppose I've not added this contingency in the figma designs.

PR #8621, opened 2025-09-21
Bump Rust version to 1.90.0

hpeebles:


megrogan (approved):

backend/canisters/community/impl/src/model/invited_users.rs
```diff
@@ -14,13 +14,6 @@ pub struct UserInvitation {
     pub timestamp: TimestampMillis,
 }
 
-#[derive(Serialize, Deserialize)]
```
ivan-jukic, line 17: Is the removal of these types related to the upgrade?

backend/canisters/community/impl/src/model/invited_users.rs
```diff
@@ -14,13 +14,6 @@ pub struct UserInvitation {
...
-pub struct ChannelInvitation {
```
ivan-jukic, line 18: Is the removal of the types below related to the upgrade?

hpeebles: These types aren't used, so clippy must've simply got better at detecting unused types.

PR #8643, opened 2025-10-28
Fix BTC exchange rate missing

hpeebles:


frontend/openchat-agent/src/services/icpcoins/mappers.ts
```diff
@@ -29,7 +31,7 @@ export function coinsByMarketcapResponse(
 
 type OverviewJson = Static<typeof OverviewJson>;
 export const OverviewJson = Type.Object({
-    price_usd: Type.Union([Type.Number(), Type.Null()])
+    price_usd: Type.Optional(Type.Union([Type.Number(), Type.Null()]))
```
hpeebles, line 34: I had to mark it as optional because we actually remove any keys where the value is `null`

hpeebles: I left the `Type.Null` in there just to be safe in case we ever change that behaviour

PR #8673, opened 2025-11-20
Log error if failed to decode notification

hpeebles:


frontend/openchat-service-worker/src/service_worker.ts
```diff
@@ -124,7 +124,6 @@ async function handlePushNotification(event: PushEvent): Promise<void> {
 
     const webPushNotification = decodeWebPushNotification(bytes, timestamp);
     if (webPushNotification === undefined) {
-        console.debug("SW: unable to decode notification", id);
```
hpeebles, line 127: This was a duplicate since we already log the error within `decodeWebPushNotification`

PR #8683, opened 2025-11-24
Use optional `email` on `register_user` as backup auth

megrogan:


Cargo.toml
```diff
@@ -73,6 +73,8 @@ members = [
     "backend/external_canisters/ckbtc_minter/c2c_client",
     "backend/external_canisters/cmc/api",
     "backend/external_canisters/cmc/c2c_client",
+    "backend/external_canisters/sign_in_with_email/api",
```
hpeebles: These were all alphabetical

backend/canisters/local_user_index/impl/src/updates/register_user.rs
```diff
@@ -212,6 +222,7 @@ fn commit(
             principal,
             user_id,
             username: username.clone(),
+            email: email.clone(),
```
hpeebles: This clone isn't needed

PR #8693, opened 2025-11-26
Update canisters post release

hpeebles:


backend/canisters/user_index/impl/src/lib.rs
```diff
@@ -407,9 +407,7 @@ struct Data {
     pub deleted_users: Vec<DeletedUser>,
     #[serde(with = "serde_bytes")]
     pub ic_root_key: Vec<u8>,
-    // TODO: Remove this after the canister is upgraded
-    pub identity_canister_user_sync_queue: VecDeque<(Principal, Option<UserId>)>,
-    #[serde(default)]
+    #[serde(alias = "identity_canister_user_sync_queue_2")]
     pub identity_canister_user_sync_queue_2: VecDeque<UserIdentity>,
```
megrogan: And rename it back to identity_canister_user_sync_queue?

hpeebles: Doh! Yeah I was going to do that but then got pulled away for a few mins and forgot

PR #8720, opened 2025-12-15
Bump Rust version to 1.92.0

hpeebles:


hpeebles:
This increases the wasm code sections significantly, and we are already right at the limit, so we cannot merge this in as is.

PR #8723, opened 2025-12-19
Display badges according to the figma design

ivan-jukic:
Updated badges as they are in the design. While they are to be displayed next to the user names, they should be less frequent.

Added H3 typography element, and updated error colour value.

Updated Chat header, paddings and fonts.

Updated Chat List items font sizes and paddings.


frontend/app/src/components_mobile/home/CurrentChatHeader.svelte
```diff
@@ -191,13 +191,12 @@
...
             {:else if hasUserProfile}
                 <Container crossAxisAlignment={"center"} gap={"xxs"}>
-                    {chat.name}
+                    {chat.name ?? chat.username}
                     <Badges
                         uniquePerson={chat.uniquePerson}
                         diamondStatus={chat.diamondStatus}
                         streak={chat.streak}
                         chitEarned={chat.chitEarned} />
-                    <Label colour={"textSecondary"}>{chat.username}</Label>
```
julianjelfs, line 200: Not sure about this. I think it _is_ useful to see both the display name _and_ the username since display names do not have to be unique.

frontend/app/src/components_mobile/home/profile/LinkAccounts.svelte
```diff
@@ -412,20 +412,20 @@
             </Column>
         {:else if providerStep === "choose_eth_wallet"}
             <div class="eth-options">
-                {#await import("../SigninWithEth.svelte")}
+                {#await import("../SignInWithEth.svelte")}
```
julianjelfs, line 415: The casing is already correct here (well it matches the files). This change makes them _incorrect_.  If we change it we need to to also change the case in the filenames.

ivan-jukic: Uhm, well in this case it's this file below right? Seems that `SignIn...` is correct.
https://github.com/open-chat-labs/open-chat/blob/the_big_fork/frontend/app/src/components_mobile/home/SignInWithEth.svelte

I have a UI docker build, it throws error when filenames are not the same. Not sure why it works on Mac, I'd expect the same 🤔

frontend/app/src/components_mobile/home/CurrentChatHeader.svelte
ivan-jukic: We can't really have both name and username in that one line, and the second line is already reserved for the online status. I mean we can, but doesn't look/feel nice...

Perhaps we could move the username to the second line... I'd rather remove it for now, and rethink how, or where we add it. In general, it's easy to access user profile from this point, and the username will be there; and it will also display by default if name is not set.

I wonder if there's any apps out there that would have a similar UI requirement...?

julianjelfs: yeah ok

PR #8744, opened 2026-01-02
Avoid applying timeouts to worker requests + track count pending

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -9628,7 +9592,14 @@ export class OpenChat {
             console.error("Error sending postMessage to worker", err);
             throw err;
         }
-        return this.responseHandler(req, correlationId, timeout);
+        return this.responseHandler(correlationId);
+    }
+
+    monitorPendingRequests() {
+        const pendingRequests = this.#pending.size;
+        if (pendingRequests >= 0) {
```
julianjelfs: >= 0 ??  

Surely this is _always_ true?

hpeebles: Ha yeah I just realised I'd pushed the limit I'd been using for local testing

PR #8751, opened 2026-01-05
Update FE to use multi-user API for changing roles

megrogan:


frontend/openchat-agent/src/services/group/mappersV2.ts
```diff
@@ -246,3 +250,19 @@ export function apiUpdatedRules(rules: UpdatedRules): TUpdatedRules {
         new_version: rules.newVersion,
     };
 }
+
+export function changeRoleResult(value: TGroupChangeRoleResponse): ChangeRoleResponse {
```
hpeebles: This one and the community one can be merged.
The other shared mapping code lives in src/services/common/chatMappersV2.ts

PR #8797, opened 2026-01-14
Avoid loading new worker after signing in

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -7953,7 +7953,7 @@ export class OpenChat {
         };
         if (assumeIdentity) {
             this.#webAuthnKey = webAuthnKey;
-            this.#authIdentityStorage.set(sessionKey, delegationChain);
+            await this.#authIdentityStorage.set(sessionKey, delegationChain);
```
hpeebles, line 7956: These awaits are because it all happens much faster now, so we need to ensure the new identity is written to indexed db before connecting to the worker

PR #8835, opened 2026-01-26
Fix race condition where chats could fail to be initialized

hpeebles:
So now, when you sign in, we stop the chats poller, then restart it once your OC identity is loaded

### DIFF
--- a/frontend/openchat-client/src/openchat.ts
+++ b/frontend/openchat-client/src/openchat.ts
@@ -801,6 +801,8 @@ export class OpenChat {
                   identityKeyAndChain.key,
                   identityKeyAndChain.delegation,
               );
+        // Stop the chats poller until we have finished loading the new identity
+        this.#chatsPoller?.stop();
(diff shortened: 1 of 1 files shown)


frontend/openchat-client/src/openchat.ts
```diff
@@ -6441,6 +6437,12 @@ export class OpenChat {
                 this.refreshSwappableTokens();
                 window.setTimeout(() => this.#refreshBalancesInSeries(), 1000);
             }
+
+            // horribly enough - we need to slightly defer this so that all the cascade of derived stuff is complete
+            // I am hopeful that we can remove this when we aren't manually synchronising runes & stores
+            tick().then(() => {
```
hpeebles, line 6443: By moving this inside the `if (initialLoad)` check, we don't need to worry about an existing iteration being in progress after we have set `chatsInitialisedStore` to false, since that value will no longer be overwritten at the end of the iteration.

PR #8841, opened 2026-01-27
Only interact with OC identity storage from within the worker

hpeebles:


frontend/openchat-client/src/openchat.ts
```diff
@@ -941,15 +938,7 @@ export class OpenChat {
         });
     }
 
-    async #loadUser() {
-        this.#startRegistryPoller();
-
-        if (this.#ocIdentity === undefined) {
-            // short-circuit if we *know* that the user is anonymous
-            this.onCreatedUser(anonymousUser());
```
hpeebles, line 949: I extracted the anonymous user bit out of here, it is now at line 838

PR #8850, opened 2026-02-03
Implement mapping of `hideOnlineStatus` in the frontend

hpeebles:


julianjelfs (approved):
Looks fine, but I wish we had had something like: 

`status: "hidden" | "visible" | "custom"` 

instead.

megrogan (approved):

PR #8862, opened 2026-02-09
Feature/native app updates

julianjelfs:
This facilitates OTA updates of the android app's embedded PWA without upgrading the app shell. 

This will upgrade itself on _every_ update at the moment. This is appropriate for the full version, but for the store version we probably need to enhance with some sort of convention e.g. it will auto update if only the patch number is higher but not if minor or major has incremented. 

Note that this is also currently configured to point at webtest rather than prod so that we can test it more easily. 

Note also that this only creates a single bundle at the moment. Follow up task will be to create two bundles: one for the store version and one for the full version.

### DIFF
--- a/frontend/tauri-plugin-oc/src/update_manager.rs
+++ b/frontend/tauri-plugin-oc/src/update_manager.rs
@@ -8,7 +8,10 @@ use tauri::{AppHandle, Manager, Runtime};
 
 const VERSION_ENDPOINT: &str = "https://webtest.oc.app/version";
 // TODO: This needs to be the actual URL where the bundle can be downloaded
-const BUNDLE_URL_TEMPLATE: &str = "https://webtest.oc.app/downloads/bundle-{}.zip";
+#[cfg(feature = "store")]
+const BUNDLE_URL_TEMPLATE: &str = "https://webtest.oc.app/downloads/store-{}.zip";
+#[cfg(not(feature = "store"))]
+const BUNDLE_URL_TEMPLATE: &str = "https://webtest.oc.app/downloads/full-{}.zip";
 
 #[derive(Serialize, Deserialize, Debug)]
 struct ServerVersion {
--- a/.github/workflows/android_release.yaml
+++ b/.github/workflows/android_release.yaml
@@ -156,16 +156,16 @@ jobs:
               env:
                   # Ensure these are passed if needed, though the gradle file uses the debug fallback we set up
                   ANDROID_KEY_STORE_FILE: ${{ env.ANDROID_KEY_STORE_FILE }}
-                  OC_DISABLE_WALLET: "false"
+                  OC_APP_STORE: "false"
 
             - name: Build Android APK (Store)
               run: |
-                  cargo tauri android build --verbose --target aarch64 armv7
+                  cargo tauri android build --verbose --target aarch64 armv7 --features store
                   mv src-tauri/gen/android/app/build/outputs/apk/universal/release/openchat-release.apk ./openchat_${{ env.VERSION }}_store.apk
               env:
                   # Ensure these are passed if needed, though the gradle file uses the debug fallback we set up
                   ANDROID_KEY_STORE_FILE: ${{ env.ANDROID_KEY_STORE_FILE }}
-                  OC_DISABLE_WALLET: "true"
+                  OC_APP_STORE: "true"
 
             - name: Upload APKs
               uses: actions/upload-artifact@v4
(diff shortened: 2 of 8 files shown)


frontend/app/src/components_mobile/UpgradeBanner.svelte
```diff
@@ -20,26 +21,51 @@
     onDestroy(() => poller.stop());
 
     function checkVersion(): Promise<void> {
-        if (import.meta.env.OC_NODE_ENV !== "production" || $activeVideoCall !== undefined)
+        console.log("Checking version");
```
ivan-jukic, line 24: Should we keep this log? For logs that we want to keep around, should we use console.info?

ivan-jukic: I see there's a few more in this section.

frontend/tauri-plugin-oc/src/lib.rs
```diff
@@ -43,7 +58,64 @@ pub fn init<R: Runtime>() -> TauriPlugin<R> {
             commands::svelte_ready,
             commands::release_notifications,
             commands::minimize_app,
+            commands::restart_app,
+            commands::get_server_version,
+            commands::download_update,
         ])
+        .register_uri_scheme_protocol("oc", |ctx, request| {
```
ivan-jukic, line 65: I've just read up a bit about this, and seems quite powerful to handle local assets. Does this now save us some time by not going to the internet for some assets?

frontend/tauri-plugin-oc/src/update_manager.rs
```diff
@@ -0,0 +1,140 @@
...
+    version: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct CachedVersion {
+    version: String,
+}
+
+pub struct UpdateManager<R: Runtime> {
+    app_handle: AppHandle<R>,
```
ivan-jukic, line 27: I wonder if this could be Arc. I think I saw the app handle being cloned in some code before, and Arc would save us a bit of memory shenanigans if we just point to the same memory location.

frontend/tauri-plugin-oc/src/update_manager.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    pub async fn get_server_version(&self) -> Result<Version, Box<dyn std::error::Error>> {
+        let client = Client::new();
+        let resp = client.get(VERSION_ENDPOINT).send().await?;
+        let server_info: ServerVersion = resp.json().await?;
+        let server_version = Version::parse(&server_info.version)?;
+        Ok(server_version)
+    }
+
+    pub async fn check_for_updates(&self) -> Result<bool, Box<dyn std::error::Error>> {
```
ivan-jukic, line 75: I've probably missed it somewhere, what's the difference between cached and bundled version? I assumed we'd only need to compare local and server versions, but I see we're doing two comparisons.

frontend/tauri-plugin-oc/src/update_manager.rs
```diff
@@ -0,0 +1,140 @@
...
+        if !cache_dir.exists() {
+            fs::create_dir_all(&cache_dir)?;
+        }
+
+        // Extract to a temp dir first then move? Or just extract.
+        // We'll extract to the cache directory.
+        // Important: We need to handle clearing old versions or overwriting.
+
+        // For simplicity, we extract all.
+        archive.extract(&cache_dir)?;
```
ivan-jukic, line 129: Is this it? We just extract and bundle installation is done?

frontend/tauri-plugin-oc/src/update_manager.rs
julianjelfs: we have a bundled version which is what it was built with (this is our ultimate fallback). Then we have the cached version which is any version that we have previously downloaded from the canister. And then we have the server version.

frontend/tauri-plugin-oc/src/lib.rs
julianjelfs: I don't think it's any different. The assets were always built into the app and they still are. The uri scheme protocol is just there to give us a hook for _where_ to load them from depending on whether we have downloaded an update or not. I _think_.

frontend/tauri-plugin-oc/src/update_manager.rs
julianjelfs: This will leave us in a state where the bundle is saved to the right place. Then we return to the PWA and tell it whether we downloaded an update or not.  The PWA part is then responsible for triggering the restart (if necessary). When the app restarts it will load the stuff from the cache directory and will be up to date.

frontend/app/src/components_mobile/UpgradeBanner.svelte
julianjelfs: Yes I was going to hang onto these for a bit because there is more to do in here and it's useful for debugging.

PR #8939, opened 2026-04-13
Lazily import scream.mp3

hpeebles:


julianjelfs (commented):
The downside to this is that it will kind of ruin the effect because the audio will not be there when you need it.

hpeebles:
Hmm the file is only 63KB, I'm sure I sure it as ~1.3MB in the network tab but that must have been something else.
Let's just close this.

PR #9008, opened 2026-05-25
Check document is valid before serving it from cache

hpeebles:


julianjelfs (commented):
As discussed - this change will perhaps allow us to recover from the situation, but it could do more to tell us exactly what is going on. Can we change it to report the exact reason that the document is not valid (undefined response, non-200 response or truncated response)?

julianjelfs:
> As discussed - this change will perhaps allow us to recover from the situation, but it could do more to tell us exactly what is going on. Can we change it to report the exact reason that the document is not valid (undefined response, non-200 response or truncated response)?

Oh wait - I think you already did!

julianjelfs (approved):

PR #9043, opened 2026-06-05
feat: redesign proposal message bubbles

ivan-jukic:
Restyle the mobile proposal card with a new header/status layout, inline vote buttons, a details sheet, and an updated voting progress bar. Adds ProposalStatusLabel and a "how a proposal passes" explainer.

image

image

image

image


frontend/app/src/components_mobile/home/proposals/ProposalContent.svelte
```diff
@@ -145,195 +201,323 @@
...
-                </CommonButton>
-            </Row>
+{#if showNeuronInfo}
+    <Sheet onDismiss={() => (showNeuronInfo = false)}>
+        <Column gap="xl" padding={["md", "xl", "huge"]}>
+            <Title fontWeight={"bold"}>
+                <Translatable resourceKey={i18nKey("proposal.noEligibleNeurons")} />
+            </Title>
+            <Body colour="textSecondary">
+                {@html $_("proposal.noEligibleNeuronsMessage", {
```
julianjelfs: Is it not possible to avoid this use of @html by changing the resource string to use markdown instead of html?

ivan-jukic: It's in response to this comment https://github.com/open-chat-labs/open-chat/pull/9043#discussion_r3372170252

But yeah, we can't render links using target="_blank" with our Markdown component, so I've reverted it back to html.

julianjelfs: But I think we already have some code in place to ensure that external links open in a new window. This certainly works in chat messages (which are using the Markdown component). I would definitely much prefer that we didn't use @html. Seems like a really bad accident waiting to happen!

ivan-jukic: I get it, let me build and test this. oc.app links might actually supposed to be open within the app.

ivan-jukic: Okay, it's markdown again. Instead of using blank, we detect which page we're navigating to, and then open that in a new custom browser tab - no html rendering necessary. Even with html anchor and blank target attr the link was opened with the app due to deep linking, so it wasn't working anyway.

julianjelfs (approved):

PR #9063, opened 2026-07-05
feat: remove turbo, consolidate frontend + runtime platform detection

ivan-jukic:
## Summary

Overhauls the frontend dev environment to match the big-chat setup, plus a merge of latest `master`.

### Turbo removal / single-package consolidation
- Drop `turbo.json` and per-package `package.json` files; merge their dependencies into a single `frontend/package.json`.
- Resolve former workspace packages via `@shared` / `@client` / `@agent` / `@worker` path aliases (vite resolves them from TypeScript source) and rewrite imports accordingly.
- Replace turbo-orchestrated npm scripts with direct vite/rollup dev and build scripts.
- Point Tauri's `beforeDevCommand` at `npm run dev:mobile` so it resolves from `frontend/`.
- Serve web and Android dev on a single port (5001); align `devUrl` and the Rust `DEFAULT_DEV_URL` fallback.

### Runtime platform detection
- Detect the native platform at runtime instead of baking it in via the `OC_APP_TYPE` env var, so the same bundle can run as web, Android, or iOS within a single dev session.
- Add `openchat-shared/src/utils/tauri.ts` with `isTauriApp` / `isAndroidTauriApp` / `isIosTauriApp` / `runtimeAppType` helpers (inspect `window.__TAURI_INTERNALS__` + user agent).
- Drop the `appType` field from client/agent config and `OpenChat`; add `isNativeIos()` and include iOS in `isNativeApp()`.

## Notes
- Includes a merge of `master`. Conflicts were import-alias overlaps (kept the `@client` aliases + master's `stripSuspendedGate`) and the consolidated `package-lock.json`.
- ⚠️ Two dep pins from `master` were adjusted down to versions resolvable on the current registry: `dompurify` `^3.4.11` → `^3.4.8` and `vite` `^6.4.3` → `^6.4.2`. Re-apply and regenerate the lock if those are intended upstream bumps.


frontend/openchat-shared/src/utils/tauri.ts
```diff
@@ -0,0 +1,41 @@
...
+    if ("__TAURI_INTERNALS__" in w) return true;
+    return w.location?.protocol === "tauri:";
+}
+
+export function isAndroidTauriApp(): boolean {
+    return isTauriApp() && /android/i.test(navigator.userAgent);
+}
+
+export function isIosTauriApp(): boolean {
+    return isTauriApp() && /iphone|ipad|ipod/i.test(navigator.userAgent);
```
ivan-jukic: **iPad may be misdetected as `web`.** The UA regex `/iphone|ipad|ipod/` misses iPadOS 13+, where WKWebView defaults to a desktop-class user agent (`Macintosh; Intel Mac OS X`) and contains none of these tokens. In a Tauri iOS app on iPad, `isIosTauriApp()` → false, so `isNativeIos()`/`isNativeApp()`/`runtimeAppType()` all report `web` — the exact bug this file exists to prevent.

Since you've already confirmed `isTauriApp()` is true here, a more robust check is to treat "Tauri + not Android + (iOS UA **or** Mac-UA-with-touch)" as iOS, e.g. `/iphone|ipad|ipod/i.test(ua) || (/macintosh/i.test(ua) && navigator.maxTouchPoints > 1)`. iPhone works as-is; iPad is the gap.

frontend/openchat-shared/src/utils/tauri.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function runtimeAppType(): "android" | "ios" | "web" {
```
ivan-jukic: Nit: `runtimeAppType()` is exported but has no callers anywhere in the frontend (grep = only this definition). Either wire it in where the old `config.appType` string was used, or drop it to avoid a dead public util.

frontend/app/build-workers.mjs
```diff
@@ -0,0 +1,57 @@
...
+// Replaces the per-package rollup builds that Turbo used to run. Reuses the
+// same source aliases as the dev server and the app build so the worker and
+// service worker pull agent/shared from source.
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { build } from "vite";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+const pkgAliases = [
```
ivan-jukic: This `pkgAliases` block is a near-verbatim copy of `ocPackageAliases` in `app/vite.config.ts` (and the `paths` in both tsconfigs). Four hand-maintained copies of the same alias→source map will drift over time. Worth extracting the JS ones into a single shared module both `vite.config.ts` and this script import.

ivan-jukic (commented):
## Review: remove turbo, consolidate frontend + runtime platform detection

Reviewed the substantive logic; the bulk of the 975 files is mechanical (import-alias rewrites `openchat-*` → `@shared`/`@client`/`@agent`/`@worker`, `package-lock.json`, turbo removal). Verified there are **no missed bare `openchat-*` imports** in `.ts`/`.svelte` source (grep = 0) and **no remaining `import.meta.env.OC_APP_TYPE` runtime reads**, so the runtime-detection swap is complete on the consuming side. Nice cleanup overall.

Findings below — 3 inline, plus the following that don't anchor cleanly to a changed hunk:

### 🔴 Security — dep pins downgraded from master
The PR body notes `dompurify ^3.4.11 → ^3.4.8` and `vite ^6.4.3 → ^6.4.2` were adjusted down to what resolved on the registry. **dompurify is the HTML sanitizer** — a downgrade can silently reintroduce a fixed XSS/mutation-XSS bypass. Please confirm whether 3.4.9–3.4.11 carried a security fix before merging, and regenerate the lock at the intended version. This is the one item I'd treat as merge-blocking until confirmed.

### 🟡 Dead code left by the runtime-detection swap
Nothing reads `OC_APP_TYPE` at runtime anymore, so these are now dead:
- `frontend/app/rollup.config.mjs` — the `import.meta.env.OC_APP_TYPE` define.
- `frontend/app/rollup-plugin-android-bundle.mjs:104` — injects `window.OC_CONFIG={OC_APP_TYPE:"android",...}` into the Android HTML; the runtime helpers no longer consult `OC_CONFIG`.

Harmless but worth removing so a future reader doesn't assume `OC_APP_TYPE` still drives behavior. (`process.env.OC_APP_TYPE` is still legitimately used at *build* time in `vite.config.ts`/`rollup.extras.mjs` — those stay.)

### 🟡 Alias list duplicated in 4 places
The `@shared/@client/@agent/@worker` → source mapping is now hand-maintained in `frontend/tsconfig.json`, `frontend/app/tsconfig.json`, `frontend/app/vite.config.ts` (`ocPackages`), and `frontend/app/build-workers.mjs` (`pkgAliases`). They agree today; they will drift. Consider a single shared module the vite/rollup/build scripts import (tsconfig has to stay separate, but the two JS copies could be deduped).

### 🟢 Module-load-time platform eval — verify Tauri injection order
`keyboard.svelte.ts` and `version.svelte.ts` freeze `isAndroidTauriApp()`/`isIosTauriApp()` into module-level consts at import time. This is correct only if `window.__TAURI_INTERNALS__` is guaranteed present before app scripts run (Tauri normally injects it first, so likely fine) — worth a sanity check on device.

### Notes
- Port unification 5003 → 5001 is consistent across `tauri.conf.json`, `lib.rs` `DEFAULT_DEV_URL`, and `vite.config.ts`; no stale `5003` left.
- `isNativeApp()` now includes iOS — intended, and all iOS-nascent paths still gate Android-specific work behind `isNativeAndroid()`.

frontend/openchat-shared/src/utils/tauri.ts
ivan-jukic: Fixed — `isIosTauriApp()` now also treats a Mac-class UA that reports touch as iOS, so iPadOS 13+ WKWebView (which defaults to a desktop UA) is detected:
```ts
/iphone|ipad|ipod/i.test(ua) || (/macintosh/i.test(ua) && navigator.maxTouchPoints > 1)
```

frontend/openchat-shared/src/utils/tauri.ts
ivan-jukic: Dropped — it had no callers anywhere. `utils` re-exports via `export *`, so removing the function removes it from the public API cleanly.

frontend/app/build-workers.mjs
ivan-jukic: Deduped — extracted the alias map into `app/oc-package-aliases.mjs` (single source), now imported by both `vite.config.ts` and `build-workers.mjs`. The two tsconfig `paths` copies still have to be hand-synced (TypeScript can't import the module) — called out in the shared file's header.

ivan-jukic:
Addressed the remaining review findings:

- **🔴 dompurify / vite pins** — both are back at the master versions (`dompurify ^3.4.11`, `vite ^6.4.3`) and `package-lock.json` resolves to `dompurify 3.4.11` / `vite 6.4.3`. No downgrade remains, so no sanitizer regression to worry about.
- **🟡 dead `OC_APP_TYPE`** — removed the `import.meta.env.OC_APP_TYPE` define in `rollup.config.mjs` and the `OC_APP_TYPE` key from the `OC_CONFIG` injection in `rollup-plugin-android-bundle.mjs`. `OC_CONFIG` itself stays: it still carries `OC_MOBILE_LAYOUT` / `OC_APP_STORE` / `OC_OTA_UPDATES`, which *are* read at runtime (`main.ts`, `themes.ts`, `features.ts`, `version.svelte.ts`, …), so the OTA bundle override path is intact.
- **🟢 module-load platform eval** — verified on device: `window.__TAURI_INTERNALS__` is present before app scripts run, so the module-level `isAndroidTauriApp()` / `isIosTauriApp()` consts evaluate correctly at import time (confirmed while diagnosing an unrelated Android issue).
- **iPad detection** and **dead `runtimeAppType`** — addressed inline.

`typecheck` + `lint` green.

ivan-jukic (commented):
## Follow-up review — new commits since `f97cc95`

Reviewed the 5 new authored commits (the branch was rebased onto master; the `#9074/#9075/#9077/#9079` commits are master's, out of scope).

### ✅ Prior review addressed — all three resolved
- **iPad detection** (`7cba83c7d`): `isIosTauriApp()` now also matches a Mac-class UA with `navigator.maxTouchPoints > 1`, catching iPadOS 13+ WKWebView. Correct, and a real desktop Mac (`maxTouchPoints === 0`) won't false-positive. 👍
- **Dead code**: `runtimeAppType()` dropped; `OC_APP_TYPE` removed from both the rollup `define` and the Android `OC_CONFIG` injection. 👍
- **Alias dup**: extracted to `app/oc-package-aliases.mjs`, imported by `vite.config.ts` + `build-workers.mjs`, with the tsconfig-must-stay-in-sync caveat documented in the file header. 👍

### ✅ Other new work — looks correct
- **Inset dedupe** (`314753e88`): Kotlin signature keys off the raw px insets + booleans (the deterministic source values, not the derived dp floats) — good choice; first emit fires (null≠signature); the JS-side `updateAndroidInterfaceSizes` equality-gate and the `value !== height` keyboard guard are sound and layer correctly with the native dedupe. No behavior regression.
- **markdownConversion straggler** (`51e92d846`): confirmed there are now **zero** remaining bare `openchat-*` imports in `.ts`/`.svelte` across the tree.
- **GTM dev/prod split** (`aec969845`): dev index.html keeps `gtag` as a no-op `dataLayer` push so page-view calls don't throw — good.

Two findings on the new commits, both inline below. Neither is merge-blocking on its own; the CSP one is worth a quick verify before release.

frontend/app/rollup.config.mjs
```diff
@@ -259,27 +285,41 @@ export default {
                 ];
                 const csp = generateCspForScripts(inlineScripts);
 
+                // Google Tag Manager + gtag is production-only: gated on the build
+                // env, never included otherwise. gtag stays defined as a no-op
+                // dataLayer push in non-production builds so page-view calls
+                // elsewhere in the app don't throw.
+                const analytics = production
```
ivan-jukic: **Verify the production analytics block is allowed by the CSP.** `generateCspForScripts(inlineScripts)` emits `script-src` with **sha256 hashes only for the two entries in `inlineScripts`** (`OC_WEBSITE_VERSION`, `parcelRequire`) and **no `'unsafe-inline'`**. This `${analytics}` block is a separate inline `<script>` that is *not* in `inlineScripts`, so its hash isn't in the CSP — with a hash-source present, browsers ignore `'unsafe-inline'`, so an un-hashed inline script is blocked, and the whole gtag/GTM bootstrap (including the dynamic `createElement` loaders inside it) never runs.

This matches the *pre-commit* behavior (the old template's inline gtag/GTM scripts were also un-hashed), so it may be that prod analytics is actually governed by a canister/header CSP rather than this meta tag. Please confirm which CSP wins in production; if it's this meta tag, add the analytics script body to `inlineScripts` so it gets hashed.

frontend/src-tauri/src/lib.rs
```diff
@@ -12,8 +12,26 @@ pub fn run() {
...
+    // OOM-killed by scudo. Confirmed with heapprofd: ~89% of the retained native
+    // memory during the crash flowed through tauri_plugin_devtools. Never built
+    // into release, and off by default in dev so a plain reload no longer
+    // crashes.
+    //
+    // `option_env!` is read at compile time (not runtime): an Android app can't
+    // pick up a shell env var at launch, so gate on the build environment and
+    // rebuild with OC_DEVTOOLS=1 when you actually need the inspector.
+    #[cfg(debug_assertions)]
+    if option_env!("OC_DEVTOOLS").is_some() {
```
ivan-jukic: Minor footgun: `option_env!("OC_DEVTOOLS").is_some()` is true for **any** value the var is set to — `OC_DEVTOOLS=0`, `OC_DEVTOOLS=false`, even `OC_DEVTOOLS=` (empty) all enable devtools, which is the opposite of what someone typing `=0` to disable would expect. Consider gating on a truthy value, e.g. `matches!(option_env!("OC_DEVTOOLS"), Some("1" | "true"))`.

frontend/app/rollup.config.mjs
ivan-jukic: Confirmed and fixed. Checked `oc.app`'s response headers — the only CSP the canister sends is `frame-ancestors` (no `script-src`), so the meta-tag CSP from `generateCspForScripts` is what governs script execution in prod, and it uses sha256 hash-sources with no `'unsafe-inline'`. So you're right — the un-hashed inline analytics block would have been blocked (as would the old pre-commit one).

Routed the gtag/GTM bootstrap body through `inlineScripts` so it's hashed into `script-src` instead of emitting a separate un-hashed `<script>`. Verified `script-src` now carries 3 sha256 hashes (version, parcelRequire, analytics body).

frontend/src-tauri/src/lib.rs
ivan-jukic: Fixed — now `matches!(option_env!("OC_DEVTOOLS"), Some("1" | "true"))`, so only an explicit truthy value enables devtools; `OC_DEVTOOLS=0` / `false` / empty no longer turns it on.

julianjelfs (approved):

Issue #9072, opened 2026-07-08
PRD: Revive unique personhood verification with an in-house IC-native face-uniqueness system


julianjelfs:
## Problem Statement

OpenChat's "verified unique person" concept was suspended in #9060 because the DecideAI verifiable-credential integration stopped working and generated constant support noise. Today every user is treated as verified: unique-person access gates silently pass, prize messages cannot be restricted to verified humans, and the verified badge is hidden. Group owners have lost a real tool against bots and multi-accounting, prize senders are exposed to sybil farming, and honest users have no way to distinguish themselves from bots.

We want the concept back — but without depending on any third-party issuer, so that the feature cannot be broken again by an external company changing direction, and so that verification is credibly "OpenChat/IC-native" end to end.

## Solution

An in-house, IC-native proof of unique personhood based on face-embedding uniqueness:

- A user opens a "Verify humanity" flow, consents to biometric processing, and completes a short camera challenge (a random sequence of head poses within a deadline).
- The captured frames are uploaded to a new OpenChat canister which computes a face embedding **on-chain** (the client is fully untrusted — it cannot forge a vector), verifies the pose challenge on-chain, and compares the embedding against all previously enrolled embeddings.
- If no similar face is already enrolled, the user's embedding is stored and a `UniquePersonProof` (new provider: OpenChat) is recorded through the existing, dormant propagation plumbing — restoring the verified badge, unique-person access gates, and human-only prize eligibility.
- Raw frames are never persisted; only the embedding is kept, and it is deleted when the account is deleted.

The whole mechanism — model weights included — is governed by the SNS DAO. No external issuer, no off-chain service.

## User Stories

1. As an OpenChat user, I want to verify that I am a unique human from within the app, so that I can access human-only groups and prizes.
2. As an OpenChat user, I want the verification to be a short guided camera flow with live feedback, so that I can complete it in under a minute without reading instructions.
3. As an OpenChat user, I want an explicit consent screen before my camera activates, so that I understand what biometric data is processed, where it is stored, and how to delete it.
4. As a privacy-conscious user, I want my raw photos/video frames discarded immediately after processing, so that no images of me persist anywhere.
5. As a privacy-conscious user, I want my stored face embedding deleted when I delete my OpenChat account, so that I can exercise my right to erasure.
6. As a verified user, I want a verified badge shown against my profile and messages, so that others know I am a unique human.
[...]

PR #9081, opened 2026-07-10
feat: in-house IC-native personhood verification (#9072)

julianjelfs:
Revives unique personhood verification (#9072) as an **in-house, on-chain, SNS-controlled** system, replacing the sunsetted third-party DecideAI/AWS integration.

## What this is (and isn't)

A new `personhood_verifier` canister runs the whole face-uniqueness pipeline on-chain under DAO governance. It is a **probabilistic sybil-resistance layer at parity with what DecideAI provided** — the same face-embedding primitive — but fully on the IC and SNS-controlled, with no third-party dependency and no biometric data leaving the chain.

It is **not** a cryptographic 1:N uniqueness guarantee at million scale. Face embeddings carry ~40–60 bits of effective separating entropy, so `P(collision) = 1-(1-FMR)^N` compounds — no single threshold is both strongly sybil-resistant and low-false-reject past low tens of thousands of users. This is an information-theoretic ceiling, not an implementation gap, and it applies equally to the DecideAI system this replaces. Positioned and tuned accordingly.

## How it works

- JPEG decode → UltraFace RFB-320 detection → insightface 2d106det landmarks → 5-point ArcFace alignment → w600k **r50** ArcFace embedding → i8-quantized cosine uniqueness scan.
- Liveness: canister-issued random head-pose challenge (Center/Left/Right/Up/Down), verified on-chain from uploaded frames.
- Privacy: raw frames are heap-only and never persisted (structurally excluded from upgrade serialization); only the i8-quantized embedding is stored, and it is deleted on account deletion — or at any time via the profile's "Remove verification" button (`remove_unique_person_proof`), which revokes the proof and erases the embedding; the user can re-verify whenever they like.
- One heavy inference per timer execution to stay inside deterministic time slicing.

## On-chain feasibility (measured)

`tract-onnx` on wasm32 with SIMD. r50 embed = **24.76B** instructions; the full Center-frame step (decode + detect + landmarks + embed) = **27.19B**, under the 40B DTS per-message ceiling (~32% headroom). Deterministic across runs. Confirmed end-to-end on a local deployment.

## Thresholds

r50-calibrated on LFW and **SNS-governable at runtime** (`set_uniqueness_thresholds`, governance-guarded, invariant `clear ≤ retry ≤ duplicate`, surfaced in `/metrics`) — the right value tracks the enrolled population, which grows. Launch defaults lean to low false-reject: `T_dup 0.55 / clear 0.45 / retry 0.50` (~1% innocent rejection at ~100k enrolled). Raise by proposal as N grows.

## Governance / lifecycle

- Models are chunk-uploaded (inert) and activated only by a **hash-pinned** `commit_model` SNS proposal.
- Legacy DecideAI proofs are wiped; a re-verification window + removal fan-out (LUI → user) handle the transition.
- `UniquePerson` group/community gates are enforced against the new proofs.

## Scope

Phases 0–3 (feasibility gate, backend skeleton, real on-chain pipeline, DecideAI cutover + UI) are complete. **Phase 4 (production rollout)** is the ordered SNS-proposal sequence, now fully documented in [`PRODUCTION_ROLLOUT.md`](https://github.com/open-chat-labs/open-chat/blob/personhood_verification/backend/canisters/personhood_verifier/PRODUCTION_ROLLOUT.md) (supersedes the earlier [rollout comment on #9072](https://github.com/open-chat-labs/open-chat/issues/9072#issuecomment-4935011239)). This PR also carries everything the rollout needs:

- `upload_model_chunks_whitelist` init arg — dev principals upload (inert) model chunks in production, mirroring `openchat_installer`'s wasm-chunk whitelist; activation stays proposal-only via the hash-pinned `commit_model` (~180×1MB chunk calls for the r50 model made proposal-per-chunk a non-starter).
- `commit_model`, `set_uniqueness_thresholds`, `set_personhood_verifier_canister_id` and `wipe_legacy_unique_person_proofs` are now `#[proposal]` endpoints, exposing the `_validate` queries SNS generic-function registration requires.
- Proposal scripts under `scripts/proposals/`: one-shot registration of the SNS functions (`commit_model` 11000, `set_uniqueness_thresholds` 11001 — verified unused against the live registry) plus an execution script per function. Per review, the verifier canister id is hardcoded (`wji62-oiaaa-aaaaf-bsc7a-cai`, created ahead of time) and the DecideAI wipe rides a designated upgrade's `post_upgrade`, so no wiring/wipe functions are needed.
- `verification_model_uploader --skip-commit` — the production upload mode; prints the exact hash-pinned commit commands for the proposals.
- **Two-phase UI rollout** via a build-time flag (`OC_UNIQUE_PERSON_REQUIREMENTS_ENABLED`): phase A ships the ability to verify (flow, badges, gate evaluator) while creating unique-person gates and prize restrictions stays unavailable; phase B — once real-world verification data looks healthy — is a website rebuild with the flag set, no code change. Verification strings are translated into all 14 locales.

## Tests

`personhood_verification` 7/7 green. `gated_group` and `delete_user` suites cover gate enforcement and embedding deletion.

## Notes for reviewers

- **New runtime dependency:** `@mediapipe/tasks-vision` — client-side face tracking for the capture UX only (self-hosted assets, degraded manual-capture fallback if it fails to load). Not in the trust path; the canister re-derives everything on-chain.
- **CSP widened:** `script-src` now includes `'wasm-unsafe-eval'` so the MediaPipe wasm can instantiate. This is a deliberately widened surface — flagged here explicitly.
- Per-stage `instruction_counter` logging in `process_verifications` is `test_mode`-gated (inert in production) — kept as a measurement aid for the r50 cost work; say if you'd prefer it stripped.
- `backend/personhood_bench/models/test_face.jpg` is a committed determinism fixture (a public-figure portrait); model weights and wasms are gitignored.

## Review response (addressed in follow-up commits)

- **Critical — stub-in-prod:** `start_verification` returns `NotReady` and `process_one_step` parks rather than ever running the stub outside `test_mode`.
- **Critical — oversized-frame DoS:** decode capped at 2048px (zune bails at the SOF header) + `upload_frame` rejects oversized dims via a new SOF parser before they enter the queue.
- **Risks:** `commit_model` builds engines before the version-bump/notify and rolls back on failure; durable retry for unacknowledged proof notifications (persisted, retried by `prune_sessions`); `cosine_similarity` asserts equal length; `MockVerifierClient` throws on a prod misconfiguration; debug overlay gated on `import.meta.env.DEV`; `saturating_sub` on byte accounting.
- **Deferred (tracked):** per-canister `test_mode` in the integration harness → prod-mode/oversized-frame tests; the remaining behavioural frontend items (getUserMedia timeout, poll clock, `NotReady` UX); session-cap split; i18n nits.

Refs #9072


backend/canisters/openchat_installer/impl/src/lib.rs
```diff
@@ -98,6 +99,8 @@ struct Data {
     registry_canister_id: CanisterId,
     translations_canister_id: CanisterId,
     website_canister_id: CanisterId,
+    #[serde(default = "anonymous_principal")]
+    personhood_verifier_canister_id: CanisterId,
```
megrogan: We should create a new canister and then hardcode the canister id here

backend/canisters/personhood_verifier/api/src/lifecycle/init.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(CandidType, Serialize, Deserialize, Debug)]
+pub struct Args {
+    // Principals permitted to commit model weights and tune thresholds (the
+    // SNS governance canister in production, the deployer locally)
+    pub governance_principals: Vec<Principal>,
+    // Principals additionally permitted to upload (inert) model chunks - the
+    // dev team in production, mirroring openchat_installer's
+    // upload_wasm_chunks_whitelist. Chunks only activate via a hash-pinned
+    // commit_model proposal, so this grants no control over the live models.
+    #[serde(default)]
```
megrogan: Don't give it a serde default - let's set the correct principals when we install

backend/canisters/user_index/api/src/updates/wipe_legacy_unique_person_proofs.rs
```diff
@@ -0,0 +1,15 @@
...
+use ts_export::ts_export;
+use types::UnitResult;
+
+// Deliberately triggers the one-off removal of all legacy DecideAI unique
+// person proofs (the cutover wipe). Governance-controlled (SNS proposal) so
+// the rollout decides exactly when badges disappear, rather than it firing
+// on upgrade.
+#[ts_export(user_index, wipe_legacy_unique_person_proofs)]
+#[derive(CandidType, Serialize, Deserialize, Clone, Debug, HumanReadable)]
+pub struct Args {}
```
megrogan: Sems a bit OTT - I would have thought doing on user_index upgrade would be fine

backend/canisters/user_index/impl/src/lib.rs
```diff
@@ -370,6 +403,17 @@ struct Data {
     pub escrow_canister_id: CanisterId,
     pub translations_canister_id: CanisterId,
     pub registry_canister_id: CanisterId,
+    #[serde(default = "anonymous_principal")]
+    pub personhood_verifier_canister_id: CanisterId,
```
megrogan: As per a previous comment - let's create a canister ahead of time and hardcode the canister id here. And no need to have a new endpoint to set it.

megrogan:
Should the personhood_spike be committed? Hasn't that done its job?

julianjelfs:
> Should the personhood_spike be committed? Hasn't that done its job?

I asked the same - apparently it's still doing a job, but I will ask it to rename to something less suspicious.

backend/canisters/personhood_verifier/impl/src/model/embeddings.rs
```diff
@@ -0,0 +1,98 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use types::UserId;
+
+pub const EMBEDDING_DIM: usize = 512;
+
+// i8-quantized L2-normalized embeddings keyed by (model version, user).
+// Heap-backed for the skeleton phase; the design moves this to a
+// StableBTreeMap before real scale (100k users ~ 51MB serialized).
+#[derive(Serialize, Deserialize, Default)]
```
megrogan: Why not start with a StableBTreeMap?

backend/canisters/openchat_installer/impl/src/lib.rs
julianjelfs: Done — the canister now exists (`wji62-oiaaa-aaaaf-bsc7a-cai`) and is hardcoded as `PERSONHOOD_VERIFIER_CANISTER_ID` in the `constants` crate (baf77fe3c). The init-arg plumbing through the installer and user_index is gone; local deployments and the integration tests create the canister with that exact id (specified-id creation), so the same constant is correct in every environment.

backend/canisters/user_index/impl/src/lib.rs
julianjelfs: Done in baf77fe3c — the endpoint is removed along with its SNS function registration and script; the user_index authenticates the verifier against the hardcoded constant, so there is nothing to set.

backend/canisters/personhood_verifier/api/src/lifecycle/init.rs
julianjelfs: Done — default removed in 09f5f3596; installs must state the whitelist explicitly.

backend/canisters/user_index/api/src/updates/wipe_legacy_unique_person_proofs.rs
julianjelfs: Agreed — endpoint removed in 09f5f3596. The wipe is now a `post_upgrade` one-liner to ship in a designated user_index release once the verification UI has baked in production (it mustn't ride the first upgrade carrying this code, because users need a working way to re-verify before their badge disappears). The rollout runbook documents this, and voters see it in that release's changelog.

backend/canisters/personhood_verifier/impl/src/model/embeddings.rs
julianjelfs: It now starts with one (09f5f3596): a StableBTreeMap is the source of truth — nothing (de)serializes through upgrades however large the enrolled population gets — with a heap cache mirroring it for the scan path, rebuilt in `post_upgrade`. The cache exists because every verification brute-force scans all enrolled embeddings, and a contiguous heap scan benched ~247M instructions per 100k users (see personhood_bench), where deserializing every entry out of stable memory per probe would cost roughly an order of magnitude more.

frontend/app/src/components/home/access/AccessGateIconsForChat.svelte
```diff
@@ -9,14 +9,10 @@
...
-    // Drop gates that reduce to no gate once suspended (unique person) gates are stripped, so we
-    // don't render empty icons or dangling separators.
-    let merged = $derived(
-        mergeAccessGates(...gates).filter((g) => stripSuspendedGate(g).kind !== "no_gate"),
-    );
+    let merged = $derived(mergeAccessGates(...gates));
 </script>
 
-{#if merged.length > 0}
+{#if gates.length > 0}
```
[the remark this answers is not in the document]
julianjelfs: Fixed in 09f5f3596 — the guard now checks `merged.length`.

julianjelfs:
I think we will probably end up not doing this since it's a bit controversial and really cannot be made to work well enough to satisfy the stated requirement.

PR #9124, opened 2026-07-27
Stop retrying c2c calls to deleted canisters, and recognise errors by code not message

hpeebles:
## The bug

A message linking to a **deleted** group or community made chats take ~13 seconds to load. Resolving the link preview queries the linked group's canister, which rejects with `DestinationInvalid` ("Canister ... not found"). That was already meant to short-circuit the retry loop, but the check was:

```ts
error.message.includes("DestinationInvalid")
```

and the SDK's message only ever contains the **numeric** code. Confirmed against the real canister from the report:

```
status: rejected
reject_code: 3            <- ReplicaRejectCode.DestinationInvalid
reject_message: Canister 4vbct-tqaaa-aaaar-bljfa-cai not found
error_code: IC0301
```

which renders as `"...Reject code: 3\n  Reject text: Canister ... not found..."` — so the substring never matched. The rejection fell through to a generic `HttpError(500)` and got the full retry treatment: **8 requests over ~12.7s** with exponential backoff. `rehydrateEventResponse` awaits preview resolution, so the whole chat blocks for that time.

Now the reject code is read from the error. The query fails on the first response, the preview is skipped, and the message renders as a plain link.

This also **un-breaks two behaviours** that silently depended on `DestinationInvalidError` being produced: detecting a group which has **moved into a community** (`group_moved`), and `channel_summary` returning `canister_not_found`.

## The same bug elsewhere

A sweep for other decisions driven by error text turned up these.

**Frontend** — where a structured field exists, use it:

- `toCanisterResponseError` now carries the rejection's IC error code (eg. `IC0301`) onto the mapped error, and the dead-ledger check reads that field instead of searching the message. An unrelated failure that merely *quotes* a dead-ledger code no longer silences error reporting.
- An invalid delegation (which forces a logout) is matched against the response **body** rather than the composed message, which also contains a JSON dump of the response headers. There is no error code for this case, so body text is the best available signal.
- Both `App.svelte` copies recognised session errors via duplicated name literals. These errors lose their prototype crossing the worker boundary (serialised with `JSON.stringify`), so `instanceof` genuinely cannot work — but the names now live beside the classes that set them, behind a shared `requiresLogout`.

**Backend** — note the situation inverts here: canisters only receive the coarse `RejectCode`, so `IC0xxx` codes found in a reject message are there at the replica's discretion. Where a real structured alternative exists, it is now used:

- **proposals_bot** treated any `CanisterError` whose text mentioned `"MessageId"` as a *successfully pushed* proposal — an 9-character substring converting a failure into `mark_proposal_pushed(...)`. It now matches `OCErrorCode::MessageIdAlreadyExists`, which is what the callee actually returns. Worth noting the old check could never have matched the real case anyway: a duplicate id arrives as `Ok(Response::Error(..))`, not a rejection.
- The **group import** "give up" check now also requires the reject code, so unrelated failures can't trigger it on text alone. The message match itself has to stay — a contract violation reaches us as a plain `CanisterError`.
- Deleted the unused duplicate of `is_out_of_cycles_error`, whose `starts_with("IC0207:")` would not have matched messages built by `convert_cdk_error` anyway.

## Tests

New `error.spec.ts` for the agent plus extensions to the shared one. The key regression test was verified to actually fail against the old code (`expected HttpError ... to be an instance of DestinationInvalidError`), not just pass against the new.

Verified: `vitest` 415 passing, `typecheck` 0 errors, `typecheck:agent`, eslint, `cargo clippy --workspace --tests`, `cargo fmt --check`, and the backend unit tests all clean.

## Not included

- **No negative caching** — each chat load still costs one failed query for a dead canister. Caching "this canister is gone" needs care, because a group genuinely can move into a community, which is what the now-working `group_moved` path handles.
- `responseTooLarge` still parses its message. The size numbers exist only in the text, and it drives page-size halving, so narrowing it has real downside for no structural gain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8493,6 +8493,7 @@ dependencies = [
  "msgpack",
  "nns_governance_canister",
  "nns_governance_canister_c2c_client",
+ "oc_error_codes",
  "proposals_bot_canister",
  "rand 0.8.5",
  "registry_canister",
--- a/backend/canisters/community/CHANGELOG.md
+++ b/backend/canisters/community/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Only abandon a group import due to a contract violation if the reject code matches too ([#9124](https://github.com/open-chat-labs/open-chat/pull/9124))
 - Prevent members from demoting others more senior than themselves ([#9115](https://github.com/open-chat-labs/open-chat/pull/9115))
 - Lock against a user having two gate payments in progress concurrently ([#9080](https://github.com/open-chat-labs/open-chat/pull/9080))
 
--- a/backend/canisters/community/impl/src/jobs/import_groups.rs
+++ b/backend/canisters/community/impl/src/jobs/import_groups.rs
@@ -12,13 +12,15 @@ use chat_events::ChatEvents;
 use constants::OPENCHAT_BOT_USER_ID;
 use group_canister::c2c_export_group::{Args, Response};
 use group_chat_core::{GroupChatCore, GroupMembers};
+use ic_cdk::call::RejectCode;
 use ic_cdk_timers::TimerId;
 use std::cell::Cell;
 use std::collections::HashMap;
 use std::time::Duration;
 use tracing::{info, trace};
 use types::{
-    Caller, ChannelId, ChannelLatestMessageIndex, Chat, ChatId, CommunityUsersBlocked, Empty, MultiUserChat, UserId, UserType,
+    C2CError, Caller, ChannelId, ChannelLatestMessageIndex, Chat, ChatId, CommunityUsersBlocked, Empty, MultiUserChat, UserId,
+    UserType,
 };
 
 const PAGE_SIZE: u32 = 19 * 102 * 1024; // Roughly 1.9MB (1.9 * 1024 * 1024)
@@ -95,7 +97,7 @@ async fn import_group(group: GroupToImport) {
                 }
                 Err(error) => {
                     mutate_state(|state| {
-                        if error.message().contains("violated contract") {
+                        if is_unrecoverable(&error) {
                             state.data.groups_being_imported.take(&group_id);
                         } else {
                             state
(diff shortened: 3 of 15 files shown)


julianjelfs:
Nice fix — reading `rejectCode` off the error rather than the message is clearly the right call, and it supersedes the `IC0301` string match I had on my `dead-ledger-retry-short-circuit` branch, which I'll drop.

One gap worth noting before this closes the issue out, though: **the retry stampede still happens for frozen and uninstalled canisters.** Only `DestinationInvalid` short-circuits `executeQuery`, and per [`ic-error-types`](https://github.com/dfinity/ic/blob/master/packages/ic-error-types/src/lib.rs) the three codes we care about don't map to the same reject code:

| IC code | Name | RejectCode |
|---|---|---|
| IC0207 | `CanisterOutOfCycles` (frozen) | `SysTransient` (2) |
| IC0301 | `CanisterNotFound` | `DestinationInvalid` (3) |
| IC0537 | `CanisterWasmModuleNotFound` | `CanisterError` (4) |

So a frozen or wasm-less canister still falls through to the generic `HttpError(500)` and gets the full treatment — 8 requests over ~12.7s, exactly the stall this PR fixes for deleted canisters, just with a different trigger. `canisterAgent/base.ts` isn't touched here, so its bail-out list is unchanged.

This bites the dead-ledger case in particular. The `isDeadLedgerError` change above stops the *Rollbar noise* for a frozen ledger, but the `refreshAccountBalance` call underneath it still burns ~12.7s of retries first, and that's exactly the ~30 day frozen window the comment describes.

The fix is small and sits neatly on top of what you've already built here — `rejectErrorCode` on `HttpError` is the hook. Roughly:

```ts
// Frozen (IC0207) or no wasm module (IC0537). The canister may recover eventually via
// top-up / reinstall, so callers shouldn't treat it as gone, but retrying within a single
// request cannot succeed. Keyed on the IC error code rather than the reject code, since
// SysTransient and CanisterError are both retryable in general.
const UNAVAILABLE_ERROR_CODES = ["IC0207", "IC0537"];
if (UNAVAILABLE_ERROR_CODES.includes(rejectErrorCode(error) ?? "")) {
    return new CanisterUnavailableError(error);
}
```

plus a `CanisterUnavailableError extends HttpError` (503) and one more line in the `base.ts` bail-out list.

Happy either way on who picks it up — I have it working on my branch already so I can rebase it onto this once this lands, or say the word and it's yours to fold in here.

julianjelfs (approved):

PR #9173, opened 2026-08-20
Name Microsoft/PhotoDNA in terms and privacy for media hash matching

julianjelfs:
The GDPR disclosure delta owed for the media hash-matching go-live (open-chat-labs/internal-governance#8):

- Terms B6: names PhotoDNA/Microsoft using the licence's pre-approved sentence, states that only a non-reversible perceptual hash is shared and the media itself never leaves OpenChat infrastructure, and that nothing is retained for clean content. Fixes the previous "specialist child-safety organisations" imprecision (the hash list is Microsoft-maintained). Mobile terms copy synced verbatim.
- Privacy §6 (safety processors): same disclosure, plus a pointer to the Microsoft privacy statement for the matching service's own records.
- Privacy §7 (retention): explicit Article 17(3)(b) erasure limits for quarantined / authority-reported content, and the clean-content non-retention statement.
- `CURRENT_TERMS_VERSION` 1 → 2, so the acceptance modal fires once on next release.

Needs a website release to take effect — the DPIA and sequencing minute are in the internal-governance PR.


julianjelfs:
Added: the terms-updated modal summary now describes this change (PhotoDNA/Microsoft, fingerprint-only matching, non-retention, deletion limits) instead of the previous update's text — the modal re-fires on the version bump, so stale copy would have described the wrong change. Stale translations of the summary dropped in favour of English fallback until retranslated.

hpeebles (approved):

hpeebles (approved):

PR #9174, opened 2026-08-20
Bump backend terms version so the acceptance notice fires

julianjelfs:
Completes the terms-version bump started in #9173. The modal gate (`Home.svelte`) compares the user's accepted version against `current_terms_version` from the user_index `current_user` response — the frontend `CURRENT_TERMS_VERSION` is only a fallback — and `accept_terms` clamps submissions to the backend constant. So without this, website 2.0.2038 shows no acceptance notice (as observed on web-test) and any acceptance would record version 1.

Needs a user_index release; website 2.0.2038 is correct as-is and the modal fires once the upgraded user_index reports version 2.

### DIFF
--- a/backend/canisters/user_index/CHANGELOG.md
+++ b/backend/canisters/user_index/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


julianjelfs:
Added a lockstep test: `terms_version_matches_frontend` parses `frontend/openchat-shared/src/constants.ts` and asserts equality with the backend constant, so the half-bump that caused this PR fails CI in either direction. Verified it fails against the un-bumped frontend and passes now.

hpeebles (approved):

Issue #9179, opened 2026-08-22
Frontend performance & memory improvements (umbrella)


julianjelfs:
Umbrella for frontend (Svelte app, openchat-client, openchat-agent, web worker, service worker) performance **and memory** improvements identified by a static audit (2026-08-22), with every load-bearing claim adversarially re-verified against the code.

**Ground rules for every child issue**
- Behaviour must not change. Add/extend a small characterisation test for what you touch *before* changing it. Shared baselines live on branch `perf-baseline-tests` (`openchat-client/src/utils/chatMerge.baseline.spec.ts`, `openchat-agent/src/utils/chat.baseline.spec.ts`) — merge that first, then branch children off `master`.
- Security must not be compromised. Issues labelled `security` touch sanitisation, payload validation, the service worker or crypto deps and need an extra reviewer.
- No new dependencies. Minimal diffs. One child issue ≈ one PR (split further if it gets big).
- CI: `cd frontend && npm run typecheck && npm run typecheck:agent && npm run lint && npm test`.

**Parallelism.** Children are grouped so most can be worked concurrently. Known file overlaps (serialise or coordinate): the *store-engine*, *allChatsStore* and *unread-counts* issues all touch `openchat-client/src/state/app/stores.ts`; *chat-rows* and *mobile-modal-stack* both touch mobile `ChatList.svelte`; *bundle-lazy-low-risk* and *composer* both touch `RichTextEditor.svelte` (lowlight hoist — do it once).

**Priority (highest value first).** Bundle split of App trees; store no-op publishes + allChatsStore rebuild; unread sweep; events pipeline; worker typebox compile; worker identicon regen; Markdown re-parse; chat-rows `$now`; SafeMap key codec; mobile modal stack; gesture listeners (`willChange`); leaks. Everything else is cheap and independent.

## Children

**Client state (serialise: #9180 → #9181 → #9182; #9183/#9184/#9185 parallel)**
- [x] #9180
- [x] #9181
- [x] #9182
- [x] #9183
- [x] #9184
- [x] #9185

**Worker / agent (all parallel)**
- [x] #9186
- [x] #9187
- [x] #9188
- [x] #9189
- [x] #9190

**Rendering (parallel; #9192 & #9196 share mobile ChatList.svelte)**
- [x] #9191
- [x] #9192
- [x] #9193
- [x] #9194
- [x] #9195
- [x] #9196
- [x] #9204
- [x] #9205
- [x] #9203
- [x] #9206

[...]

PR #9210, opened 2026-08-22
Stop chat-list rows rebuilding on the unused 5s clock

### LINKED ISSUE #9192: perf: chat-list rows rebuild every 5s on an unused `$now`; per-keystroke row invalidation; hidden rows still reactive
Part of #9179.

**Area:** `app/src/components/home/ChatSummary.svelte` (93, 117-140, 142, 266-270, 415-422), `app/src/components_mobile/home/ChatSummary.svelte` (98, 181), `components/home/CurrentChatMessages.svelte:341`, `thread/Thread.svelte:373`, `components/home/ChatList.svelte` (63-71, 187, 238-244), `components_mobile/home/ChatList.svelte:72-82`, `components/Avatar.svelte:44-56`, `app/src/utils/effects.svelte.ts`, `app/src/stores/chatShortcuts.ts:68-112`

**Confirmed.**
- Both `ChatSummary` files: `normaliseChatSummary($now, ...)` where `_now` is never referenced — every row rebuilds its `chat` object (displayName, avatar URL, typing string) every 5s, including while the panel is `display:none` on mobile. Two-line fix.
- `editing={$selectedChatDraftMessageStore?.editingEvent === evt}` inside the row snippet; every keystroke produces a new `DraftMessage` so every rendered row re-evaluates. Hoist `editingEvent` to a component-level `$derived`.
- Desktop `ChatSummary`: duplicated unarchive effect (136 and 267) drags `chat` (hence `$now`) into the effect deps.
- `Avatar` status `trackedEffect` re-runs per avatar per 5s tick; `trackedEffect` bookkeeping runs in prod — gate on `import.meta.env.DEV`.
- `ChatList.chatsToShow` only grows; mobile never resets on scope change (`chatsToShow = TO_SHOW` in `onScopeChanged`); desktop hidden (`visible=false`) rows still mount and run effects — move the test up to `{#if}` in `ChatList`.
- `chatShortcuts.ts` sorts all chats before filtering, recomputes on every `allUsersStore` tick.

---

julianjelfs:
Closes #9192. Part of #9179.

- Desktop and mobile `ChatSummary.svelte`: `normaliseChatSummary` took `$now` as a parameter it never read, so every row rebuilt its derived `chat` object (display name, avatar URL, typing string) every 5 seconds. Dependency removed.
- Desktop `ChatSummary.svelte`: removed a duplicated unarchive effect that pulled `chat` (and hence the clock) into its dependencies; the identical check already runs in `updateUnreadCounts`.
- `CurrentChatMessages.svelte`: `editingEvent` hoisted to a component-level `$derived` so typing a draft no longer re-evaluates every rendered row's `editing` prop per keystroke.
- `effects.svelte.ts`: `trackedEffect` bookkeeping (timestamp array + filter per run) now only runs in development builds (`OC_BUILD_ENV === "development"`).
- Mobile `ChatList.svelte`: `chatsToShow` resets on scope change instead of growing for the session.
- `chatShortcuts.ts`: filter before sort.

Skipped deliberately: unmounting hidden archived rows (auto-unarchive relies on them being mounted) and the `Avatar` status poll (intentional 5s cadence).

Tests: new `chatShortcuts.spec.ts`; full suite 459 passing; svelte-check 0 errors.


hpeebles (approved):

julianjelfs:
## Quantified (analytic; app-level webtest measurement to follow when a visible tab is available)

- **5 s clock dependency:** on master every mounted chat-list row re-ran `normaliseChatSummary` (display name, avatar URL, typing string, favourite lookup, fresh object) on each `$now` tick — 12 recomputes per row per minute, for every row, for the whole session, including while the mobile chat list is hidden behind `display:none`. With 50 rows: **600 derived recomputes/min → 0**; with 300 rows after scrolling a long list: **3,600/min → 0**. The only remaining per-tick work is the avatar status effect, which is intentional.
- **Keystroke fan-out:** each character typed created a new draft object and re-evaluated `editing` on every rendered message row (≈ 30–60 in the virtual window). Typing a 40-character message: **~1,600–2,400 row re-evaluations → 0** (one component-level derived instead).
- **`trackedEffect` bookkeeping** (timestamp array + filter per run) is now skipped in production for every tracked effect on every run.
- **Mobile `chatsToShow` reset:** switching scope after scrolling a 300-row list now mounts 30 rows instead of 300.

These are structural counts rather than timings; the per-recompute cost is small (tens of µs) — the point is removing a constant background tick multiplied by row count.

frontend/app/src/components/home/ChatSummary.svelte
```diff
@@ -263,12 +261,6 @@
         return `${user}: ${latestMessageText}`;
     }
 
-    trackedEffect("unarchive-chat", () => {
```
julianjelfs, line 266: why has this effect been removed. How are chats appropriately unarchived without it.

frontend/app/src/components/home/ChatSummary.svelte
```diff
@@ -90,7 +88,7 @@
     let hovering = $state(false);
     let unreadMessages = $state<number>(0);
     let unreadMentions = $state<number>(0);
-    let chat = $derived(normaliseChatSummary($now, chatSummary, $typersByContext));
+    let chat = $derived(normaliseChatSummary(chatSummary, $typersByContext));
```
julianjelfs, line 91: $now was definitely being passed in to deliberately cause this function to run every five seconds. This looks quite intentional. Are we sure it can be removed without causing some staleness in the UI. It's possible that it is hack that was only necessary in Svelte 4 and because unnecessary in Svelte 5 through better dependency tracking but I would like to be reassured on that.

frontend/app/src/utils/effects.svelte.ts
```diff
@@ -10,6 +10,11 @@ export function trackedEffect(
         durationMs?: number;
     },
 ) {
+    if (import.meta.env.OC_BUILD_ENV !== "development") {
```
julianjelfs, line 13: This means we are blind to runaway effects in the prod environment. Is this really saving us anything? Is it worth the loss of visibility?

frontend/app/src/components/home/ChatSummary.svelte
julianjelfs: The same check already lives in `updateUnreadCounts` (line 128–140 on master):

```ts
if (chatSummary.membership.archived && unreadMessages > 0 && !chat.bot) {
    unarchiveChat();
}
```

and that runs on every `chatSummary` change (`$effect(() => updateUnreadCounts(chatSummary))`, line 117) and on every `messagesRead` publish (the subscription at line 120). `unreadMessages` is only ever assigned inside `updateUnreadCounts`, so the removed `trackedEffect` could only fire immediately after it, with identical inputs — the duplicate was always a no-op. The one extra dependency the effect had was `chat.bot`; since `them?.kind === "bot"` is `false` while the user record hasn't loaded, the effect would already have unarchived before bot-ness was known, so removing it doesn't change that edge either.

The reason for removing it rather than leaving a harmless duplicate: it read `chat`, which pulled the 5 s clock into the effect's dependencies. If you'd rather keep a belt-and-braces copy I can put it back reading `chatSummary` + `unreadMessages` only — say the word.

frontend/app/src/components/home/ChatSummary.svelte
julianjelfs: Fair challenge — I checked rather than assumed. `$now` was added to this call in 25159a092 (2022, Svelte 3, "Begin to extract an open chat client library") and `_now` has never been read in the body.

What `normaliseChatSummary` actually reads: `$chatListScopeStore`, `$favouritesStore`, `membership.notificationsMuted`, `videoCallInProgress`, `$allUsersStore.get(them)`, `client.displayName/userAvatarUrl/groupAvatarUrl` (pure functions of those), `$selectedCommunitySummaryStore`, and `client.getTypingString($_, $allUsersStore, ctx, $typersByContext)`. No wall-clock input anywhere. In Svelte 5 every store read inside the `$derived` is a tracked dependency, so any of those changing re-runs it.

The two things in the row that *look* time-driven:
- typing indicator expiry — driven by the typing store's own `setTimeout → _delete` (`stores/typing.ts:39–75`), which publishes `$typersByContext` and re-runs the derived; the tick was never what expired it.
- the date label — that's the separate `displayDate` derived (line 95) formatted at line 480 via `formatMessageDate(…, "today", "yesterday")`, and it never depended on `$now` either (so a "Today" label rolls over on the next `chatSummary` change — pre-existing, unchanged by this PR).

So I can't find a staleness path. If you want a safety net, the sensible place is the date label (the one genuinely clock-sensitive display), not the whole object: `$derived(($now, client.getDisplayDate(chatSummary)))` keeps a 5 s refresh for just that string. Happy to add that if you prefer.

frontend/app/src/utils/effects.svelte.ts
julianjelfs: Honest answer: we were already blind in prod. The only output is `console.warn` (no Rollbar hook — there's no `captureConsole` anywhere in the Rollbar config), so the warning only ever reached someone with DevTools open on the production site. The saving is small but per-run on the hottest effects (`read-by-them` per message row, the virtual list): `performance.now()` + array push + `filter` on every execution.

Three options, your call:
1. Revert the gate — keep the dev-console visibility at that per-run cost.
2. Keep the gate but add a runtime opt-in (`localStorage.effect_debug = 1`, same pattern as `vcl_debug`) so it can be switched on in prod without a rebuild.
3. Make it genuinely useful in prod: when the threshold is exceeded, report once via Rollbar (rate-limited per effect name) instead of `console.warn`, and keep the bookkeeping on — turns a blind check into telemetry.

I'd lean to 3 (or 2 if you'd rather not add Rollbar noise). Tell me which and I'll push it to this branch.

PR #9212, opened 2026-08-22
Skip store publishes that change nothing

### LINKED ISSUE #9180: perf: stop no-op store publishes (`notEq` writables, `setUpdated` empty guard)
Part of #9179.

**Area:** `openchat-client/src/utils/stores.ts`, `openchat-client/src/state/utils.ts`, `state/users/state.ts`

**Problem.** 72 `writable` + 19 `derived` stores use `notEq = () => false`, so every `.set`/`.update` publishes even when nothing changed. Concretely `userStore.setUpdated()` (`state/users/state.ts:61`) calls `allUsersStore.update` every 60s user poll even with an empty list (`addMany` guards empty; `setUpdated` doesn't). Derived stores also become permanently live after first `.value` read (`stores.ts:238`, NOOP subscriber never released), so every publish recomputes all of them synchronously.

**Do.**
1. Early-return in `setUpdated` (and any similar mutators) when the input is empty / nothing mutated.
2. Audit each `notEq` mutation site: where the update is a no-op, skip the publish (return the same reference and use real equality, or a `didChange` guard). Keep `notEq` where in-place mutation is intentional and callers rely on it.
3. Optional, separate PR: `.value` on derived is eager-forever. A lazy `.value` needs new invalidation (version counters) — the existing `dirty` machinery only works while subscribed. Scope first; do not attempt blind.

**Risk:** medium — a missed mutation site drops a legitimate update. Characterise with `openchat-client/src/utils/stores.spec.ts`.

---

julianjelfs:
Closes #9180 (items 1 and 2). Part of #9179.

Most client stores use `notEq`, so every `set`/`update` publishes and synchronously recomputes every live derived store even when nothing changed. This adds cheap guards at the mutation sites where the caller can tell:

- `userStore.setUpdated`: only touches users whose `updated` differs; returns before publishing when there are none. This ran `allUsersStore.update` on every 60-second user poll even with an empty list, which in turn re-derived chat summaries and re-parsed mention-bearing Markdown.
- `userStore.userSuspended`: no publish for unknown users or when already in the requested state.
- `cryptoBalanceStore.setBalance`: no publish when the balance is unchanged (timestamp still refreshed).
- `updateUserStoreFromEvents`: `selectedChatUserIdsStore` only updated when the events introduce user ids not already in the set (runs on every events load / scroll-back).

Left alone: undo-paired local-update stores, `markMessageRead` (callers already gate on `isRead`), and stores that receive fresh server objects where equality would need a deep compare. The lazy derived `.value` idea in the issue is out of scope — it needs new invalidation machinery.

Tests: new `state/users/state.spec.ts` (no-op cases failed before the change, pass after), `appStores.spec.ts`, extended `stores.spec.ts`; full suite 467 passing; svelte-check 0 errors.

### DIFF
--- a/frontend/openchat-client/src/openchat.ts
+++ b/frontend/openchat-client/src/openchat.ts
@@ -2803,14 +2803,15 @@ export class OpenChat {
             allUserIds.add(u);
         }
         userStore.addWebhookIds([...webhooks]);
-        selectedChatUserIdsStore.update((set) => {
-            [...allUserIds].forEach((u) => {
-                if (u !== userId) {
-                    set.add(u);
-                }
+        const newChatUserIds = [...allUserIds].filter(
+            (u) => u !== userId && !selectedChatUserIdsStore.value.has(u),
+        );
+        if (newChatUserIds.length > 0) {
+            selectedChatUserIdsStore.update((set) => {
+                newChatUserIds.forEach((u) => set.add(u));
+                return set;
             });
-            return set;
-        });
+        }
         await this.getMissingUsers(allUserIds);
     }
 
--- a/frontend/openchat-client/src/state/app/appStores.spec.ts
+++ b/frontend/openchat-client/src/state/app/appStores.spec.ts
@@ -37,6 +37,7 @@ import {
 import { addToWritableMap } from "../utils";
 import {
     allChatsStore,
+    cryptoBalanceStore,
     allServerChatsStore,
     chatListScopeStore,
     chatSummariesStore,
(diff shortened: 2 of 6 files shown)


hpeebles (approved):

julianjelfs:

`allUsersStore` seeded with 2,000 users; one derived subscribed that does a fixed 0.2 ms of work per recompute (stand-in for the real fan-out, which is many derived stores); 60 simulated polls; median of 7 runs.

**`userStore.setUpdated`** — downstream derived recomputes per 60 polls:

| scenario | before | after |
|---|---|---|
| empty id list (what the 60 s poll sends when nothing changed) | 60 recomputes, ~0.21 ms/call | **0 recomputes**, 0.000 ms |
| 600 ids already at the timestamp | 60 recomputes, 0.24–0.46 ms/call | **0 recomputes**, 0.025 ms |
| 600 ids genuinely needing update | 60 recomputes, 0.23–0.33 ms/call | 60 recomputes, 0.25 ms (no regression; pre-filter ≈0.02 ms) |

**`selectedChatUserIdsStore` on events load** (300 events, all senders already known): 60 → **0** publishes; 0.2–0.34 ms → 0.007 ms per call.

The ms columns are dominated by the synthetic derived; the real result is the count column — every no-op publish on master fanned out to all live derived stores hanging off `allUsersStore`, once a minute, for the whole session. That tick now costs ~0 when nothing changed.

PR #9213, opened 2026-08-22
Share one derived store across all Translatable nodes

### LINKED ISSUE #9204: perf: `<Translatable>` allocates a 3-store derived + subscription per instance (~2000 sites)
Part of #9179.

**Area:** `app/src/actions/translatable.ts:75-104`, `app/src/components/Translatable.svelte:14`, `app/src/i18n/i18n.ts:177`

**Confirmed.** Each mounted `<Translatable>` builds `derived([locale, dictionary, editmode])` and subscribes, to decide whether to show a translator pencil. `editmode` is toggled only from `UserProfile.svelte:495`. 2051 call sites, churned by virtual-list recycling. No leak (unsubscribed on destroy) — pure setup cost and 2000+ subscribers on `dictionary`/`locale`.

**Do.** One module-level derived over `[locale, dictionary, editmode]`; per node, subscribe to `editmode` alone and do the key lookup lazily when it flips on.

---

julianjelfs:
Closes #9204. Part of #9179.

`use:translatable` is on ~2000 sites and is churned by virtual-list recycling. Each instance built its own `derived([locale, dictionary, editmode])` plus three subscriptions, purely to decide whether to show the translator pencil — a mode almost nobody is in.

Now a single module-level derived yields `{locale, dictionary}` only while translation edit mode is on for a non-English locale, and `undefined` otherwise; Svelte dedupes `undefined → undefined`, so in the normal case locale/dictionary churn notifies nobody and no node does a dictionary lookup. A per-node boolean preserves the old dedupe so a locale change in edit mode does not insert a duplicate pencil.

Tests: new `translatable.spec.ts` (13 tests; 12 pass against the old implementation too — only the subscription-count assertion flips). Full suite 468 passing; svelte-check 0 errors.

### DIFF
--- a/frontend/app/src/actions/translatable.spec.ts
+++ b/frontend/app/src/actions/translatable.spec.ts
@@ -1,0 +1,190 @@
+import { addMessages, dictionary, locale } from "svelte-i18n";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+
+// The real theme module drags in component-lib's svelte components (which vitest's
+// svelte plugin can't preprocess) and needs matchMedia. The action only reads
+// `theme.accent`, so a stub is enough.
+vi.mock("../theme/themes", async () => {
+    const { readable } = await import("svelte/store");
+    return { currentTheme: readable({ accent: "#22a7f2" }) };
+});
+
+import { editingLabel, editmode, i18nKey } from "../i18n/i18n";
+import { translatable } from "./translatable";
+
+type Handle = ReturnType<typeof translatable>;
+
+const handles: Handle[] = [];
+
+function mount(key: string | undefined, parent?: HTMLElement): HTMLElement {
+    const container = parent ?? document.body.appendChild(document.createElement("div"));
+    const node = container.appendChild(document.createElement("span"));
+    const handle = translatable(node, { key: key === undefined ? undefined : i18nKey(key) });
+    handles.push(handle);
+    return node;
+}
+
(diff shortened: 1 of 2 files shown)


hpeebles (approved):

julianjelfs:

`translatable` action mounted on 2,000 jsdom spans, real `fr.json` dictionary (2,024 keys, one per node), `editmode` off; median of 9 runs, `--expose-gc` for heap.

| metric | before | after | Δ |
|---|---|---|---|
| mount 2,000 nodes | 1.85 ms | 0.34 ms | −81% |
| destroy 2,000 nodes | 0.68 ms | 0.13 ms | −80% |
| store subscriptions per mount cycle (locale / dictionary / editmode) | 2,000 / 2,000 / 2,000 | **1 / 1 / 1** | 6,000 → 3 |
| one `dictionary.set()` with 2,000 mounted, editmode off | 0.98 ms | **0.02 ms** | −98% |
| same, editmode on (translator mode) | 1.6 ms | 0.88 ms | −45% |
| heap retained by 2,000 mounted nodes | 4.78 MB | 0.82 MB | −83% |

At app scale: virtual-list recycling remounts `<Translatable>` continuously while scrolling, so the per-node setup and the three subscriptions come off every recycled row; and every `dictionary` publish (each lazily-loaded locale chunk, each `addMessages`) stops walking 2,000 no-op subscribers on the main thread because the shared store dedupes `undefined → undefined`.

PR #9215, opened 2026-08-22
Memoise identicon generation in the worker

### LINKED ISSUE #9188: perf (worker): identicon data-URIs regenerated (md5+SVG+base64) for every avatarless user every call
Part of #9179.

**Area:** `openchat-agent/src/services/openchatAgent.ts:1249-1276` (`rehydrateUserSummary`), `:449-455` (`getAllCachedUsers`), `services/userIndex/userIndex.client.ts:514-517, 559`, `utils/chat.ts:466-482`

**Problem.** `rehydrateUserSummary` short-circuits only if `blobUrl` is set, but cached `UserSummary` records never have it — `UserIndexClient.getUsers` writes to cache *before* `OpenChatAgent.getUsers` rehydrates. So every minute up to ~600 users get `md5 + new Identicon(svg) + base64`, and `getAllCachedUsers` at startup does it for the entire user cache, then structured-clones all the data URIs to the main thread.

**Do.** Memoise `buildIdenticonUrl` per userId in a bounded `Map` (identicon is a pure function of userId), and/or persist the derived `blobUrl` into the cache keyed on `blobReference.blobId` so avatar changes still invalidate. Verify behaviour with a test that the same userId yields the same URL and that a user with a real `blobReference` is not memoised.

---

julianjelfs:
Closes #9188. Part of #9179.

`rehydrateUserSummary` only skips avatar work when the cached summary already carries a `blobUrl`, but cached records never do (the user index client writes to the cache before rehydration). So every avatarless user was regenerating its identicon — md5 + SVG construction + base64 — on every 60-second user poll and for the whole user cache at startup.

`buildIdenticonUrl` is a pure function of the user id, so it is now memoised behind a bounded map (5000 entries, insertion-order eviction). Nothing persisted to the cache changes.

Tests: new `identicon.baseline.spec.ts` pins the exact data-URI produced by the old code; full suite 464 passing; `tsc` for openchat-agent clean.


hpeebles (approved):

julianjelfs:

`buildIdenticonUrl`, 7 trials, fresh module per trial (empty memo), median.

**Per-minute poll shape — 600 users × 10 passes**

| | before | after |
|---|---|---|
| pass 1 (cold) | 4.4 ms | 4.8 ms (+8%, memo insert) |
| passes 2–10 (each) | ~3.9 ms | **0.02 ms** (−99%) |
| 10-pass total | 39.9 ms | 5.1 ms (−87%) |

**Startup shape — 3,000 distinct ids once:** 20.3 ms → 21.3 ms (no change; cold is cold).

**Retained heap after 3,000 ids:** ~0 → **+3.3 MB** (~1.1 KB per data URI; ~5.5 MB at the 5,000 cap).

Honest reading: the CPU win is real but small in absolute terms (~4 ms/min of worker time); the more material effect is allocation churn — ~600 × 1.1 KB of throwaway data-URI strings per poll (~40 MB/hour) now zero, and in a shared-process run the discarded strings measurably slowed neighbouring work (directional, not isolated). Cost is bounded retention in the worker. If that trade is unwelcome, `MAX_IDENTICONS` can drop to ~1,500 (covers the per-poll working set, ~1.6 MB) at the price of re-generating on cold startup only.

PR #9216, opened 2026-08-22
Cap and dedupe mentions; tidy summary merge hot paths

### LINKED ISSUE #9190: perf/memory (worker): mentions accumulate forever; `mergeThreads` includes-in-filter; previews extracted twice
Part of #9179.

**Area:** `openchat-agent/src/utils/chat.ts` (293-299, 391-399), `utils/community.ts` (234-240, 268-275), `services/openchatAgent.ts` (1089-1106, 1180-1202), `services/common/chatEvents.ts:582-599`

**Confirmed.**
- `membership.mentions: [...g.membership.mentions, ...c.membership.mentions]` — no cap, no dedupe; the canister caps each *response* at 50 but the frontend array grows for the life of the cache (30 days), is re-serialised into the chat blob every updates cycle and filtered on every unread recompute.
- `mergeThreads`: `unfollowedThreads.includes(...)` inside `filter` — hoist to a `Set`.
- `extractMessagePreviews` runs twice per text message (`findMissingMessagePreviewsByChat` and `rehydrateEvent`).
- Window-expansion fallback awaits both directions sequentially despite the "concurrently" comment.

**Do.** Dedupe mentions by `messageId` and cap to the most recent N (match server cap, e.g. 50 — confirm product is OK with very old unread mentions dropping); `Set` for unfollowed; compute previews once per event; `Promise.all` the two directions.

Baseline: `openchat-agent/src/utils/chat.baseline.spec.ts` pins current mention behaviour (`[7, 1, 1]` case) — update that expectation deliberately in the same PR.

---

julianjelfs:
Closes #9190. Part of #9179.

- `membership.mentions` was concatenated on every updates cycle with no cap or dedupe. The canister caps each response at 50 (`MAX_RETURNED_MENTIONS`), but the frontend array grew for the life of the cache, was re-serialised into the chat blob every 5 s and filtered on every unread recompute. New `mergeMentions` dedupes by `messageId` and keeps the 50 newest. **Behaviour note:** a user with more than 50 unread mentions in one chat now sees only the 50 newest in the cached array — matches what a fresh load would return.
- `mergeThreads` (group + channel): `Set` lookup for unfollowed threads instead of `includes` inside `filter`.
- `extractMessagePreviews` ran twice per text message (`findMissingMessagePreviewsByChat` and again in `rehydrateEvent`); computed once and passed through.
- `chunkedChatEventsWindowFromBackend`: both directions fetched concurrently as the comment already claimed; failure short-circuit order preserved.

Tests: `chat.baseline.spec.ts` expectation deliberately changed from `[7, 1, 1]` to `[7, 1]`, plus dedupe and cap tests; new `community.baseline.spec.ts`; full suite 461 passing; `tsc` for openchat-agent clean.


hpeebles (approved):

julianjelfs:

`mergeGroupChatUpdates` over 200 group chats, 40 threads each, updates carrying 40 thread updates + 20 unfollowed, mentions arrays pre-grown to 2,000 per chat (long-lived cache) with 50 new per cycle; 100 cycles; median of 5 runs.

| | before | after | Δ |
|---|---|---|---|
| ms per update cycle (200 chats) | 15.7 (and climbing each cycle) | 3.3 (flat) | **−79%** |
| final `mentions.length` per chat | 7,000 and growing | 50 | −99% |
| `JSON.stringify` size of merged summaries (proxy for the chat blob written to IndexedDB every 5 s) | 83 MB | 1.7 MB | **−98%** |
| `mergeThreads` alone (40×20) | 1.3 ms | 1.5 ms | within noise |

The `includes → Set` change is not measurable at realistic sizes — a complexity tidy, not a claimed win. The 2,000-mention input is an upper bound; the point is that on master the array (and the IDB blob it rides in every 5 s) had no ceiling for the 30-day cache lifetime, and now it is bounded at 50 regardless of session age.

PR #9217, opened 2026-08-22
Stop the cache primer on agent replacement; fix stranded worker requests

### LINKED ISSUE #9186: perf/memory (worker): CachePrimer grows forever, never stops; stranded worker requests
Part of #9179.

**Area:** `openchat-agent/src/utils/cachePrimer.ts` (118-125, 140-155, 373-383), `services/localUserIndex/localUserIndex.client.ts:145-203`, `utils/chatsDb.ts:1204-1221` (`readAll`), `openchat-worker/src/worker.ts:331-334`, `openchat-client/src/workerAgent.ts:124-152`

**Confirmed.**
- `proposalChatIds.push(chat.id)` with no dedupe; entries never removed (`removeFromPending` only touches `#pending`). `processProposalTallies` re-arms itself every minute in `finally`, no stop — a replaced/logged-out `OpenChatAgent` (worker.ts 286/301/320) stays alive via the timer closure, with its HttpAgent, clients and IDB connection.
- `localUserIndex.client.ts`: cache reads and writes in the batch loop are awaited serially (`Promise.all` them).
- `readAll`: one `await cursor.continue()` per record, `transaction.done` never awaited — use `getAll()`/`getAllKeys()`.
- `#usersLoaded` never cleared.
- `worker.ts`: request arriving while `agent` is undefined is logged and dropped — no `worker_error` posted, so `#inflightRequests` entry and caller promise hang forever (logout / pre-`setAuthIdentity` window).

**Do.** Dedupe proposal chats (Set keyed by chat id string); keep timer handle + add a `stop()` called when the agent is replaced; parallelise batch IDB reads/writes; `getAll` in `readAll`; post a `worker_error` for the no-agent branch (callers must tolerate rejection) and optionally TTL-sweep inflight requests.

---

julianjelfs:
Closes #9186. Part of #9179.

- `CachePrimer` tracked governance-proposal chats in a plain array with no dedupe, so ids were appended every time a proposal chat had updates, and the per-minute tally job re-armed itself forever — a replaced or logged-out `OpenChatAgent` stayed alive through that timer (with its HttpAgent, clients and IndexedDB connection). Now keyed by chat id, removed chats are dropped, and `CachePrimer.stop()` / `OpenChatAgent.dispose()` are called from the worker before constructing a new agent and on logout.
- `localUserIndex.client.ts`: independent cache reads and writes in the batch loop run concurrently instead of serially.
- `chatsDb.readAll`: `getAllKeys` + `getAll` in one transaction instead of one `await` per cursor step; `transaction.done` awaited.
- `worker.ts`: a request arriving while there is no agent (logout / before `setAuthIdentity` resolves) was logged and dropped, leaving the caller's promise and the `#inflightRequests` entry hanging forever. It now posts a `worker_error` so the promise rejects. **Behaviour note:** callers in that window now get a rejection instead of a hang; no dependence on "never settles" was found.

Tests: new `cachePrimer.spec.ts` (fails against the old code), `workerAgent.spec.ts`; full suite 459 passing; `tsc` for openchat-agent clean; svelte-check 0 errors.

### DIFF
--- a/frontend/openchat-agent/src/services/localUserIndex/localUserIndex.client.ts
+++ b/frontend/openchat-agent/src/services/localUserIndex/localUserIndex.client.ts
@@ -142,14 +142,15 @@ export class LocalUserIndexClient extends MultiCanisterMsgpackAgent {
         const partialCachedResults = [] as EventWrapper<ChatEvent>[][];
         const requestsToBackend = [] as ChatEventsArgs[];
 
+        const cacheResults = await Promise.all(
+            requests.map((request) =>
+                this.getEventsFromCache(request.context, request.args, !cachePrimer),
+            ),
+        );
+
         for (let i = 0; i < requests.length; i++) {
             const request = requests[i];
-
-            const [cached, missing, dirty, totalMiss] = await this.getEventsFromCache(
-                request.context,
-                request.args,
-                !cachePrimer,
-            );
+            const [cached, missing, dirty, totalMiss] = cacheResults[i];
 
             if (missing.size + dirty.size > MAX_MISSING || totalMiss) {
                 requestsToBackend.push(request);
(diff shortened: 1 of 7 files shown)


hpeebles (approved):

julianjelfs:
## Quantified (master vs branch; counts exact, latencies simulated where stated)

**(a) Proposal-tally polling** — 720 update cycles (1 h at 5 s) with 20 governance chats updating each cycle, then one 1-minute tick:

| | before | after |
|---|---|---|
| chat ids sent to `updateProposalTallies` per tick | 14,400 (grows by 20 every cycle, forever) | 20 (constant) |
| LUI `update_proposal_tallies` calls per minute | **1,440** | **2** |

This is effectively a production bug: a user in governance chats was issuing ~1,440 canister calls a minute after an hour online, doubling each hour.

**(b) Agent replacement** — 5 simulated `setAuthIdentity` cycles: live per-minute timers (each retaining an orphaned `OpenChatAgent` + HttpAgent + clients + IDB handle) **5 → 1**.

**(c) LUI batch loop** — 40 requests, all partial cache hits, 2 ms simulated per IDB op: 198 ms → 5.3 ms (80 serial round-trips → 2 parallel waves). Simulated latency; the round-trip shape is exact.

**(d) `chatsDb.readAll`** — 500 cached chats, 0.5 ms simulated per cursor step: 630 ms → 1.3 ms (501 awaits → 2 bulk reads). Runs once per login before the cache primer can start; with ~500 cached chats this is roughly half a second off startup.

PR #9218, opened 2026-08-22
Batch IndexedDB reads into one transaction per call

### LINKED ISSUE #9187: perf (worker): one IndexedDB transaction per user / per event index
Part of #9179.

**Area:** `openchat-agent/src/utils/userCache.ts:36-43`, `utils/chatsDb.ts` (544-565 `getCachedEventsByIndex`, 865-880 `loadMessagesByMessageIndex`)

**Problem.** `getCachedUsers` does `Promise.all(userIds.map(u => db.get("users", u)))` — `idb`'s `db.get` opens a transaction per call. Called every minute from `#updateUsers` with hundreds of ids (500 stale + 100 recent + direct-chat counterparties) and from the cache primer's `populateUserCache`. `setCachedUsers` already does it right (one explicit tx). Same per-key pattern in chatsDb for events by index and messages by message index.

**Do.** Open one `readonly` transaction, take the store/index once, issue all gets against it. Pure mechanical change; add a small test with `fake-indexeddb` only if already available — otherwise rely on existing coverage and manual check.

---

julianjelfs:
Closes #9187. Part of #9179.

`getCachedUsers` issued `db.get` per user id — one implicit IndexedDB transaction each — and is called every minute from the user poll with several hundred ids, plus from the cache primer. `getCachedEventsByIndex` and `loadMessagesByMessageIndex` did the same per event/message index. Each now opens a single `readonly` transaction, issues all gets against it synchronously, and awaits `tx.done`. Results are collated positionally, so output order is now explicitly input order (previously the de facto order).

Tests: new `userCache.spec.ts` and `chatsDb.spec.ts` using a hand-rolled fake that counts transactions for both the implicit and explicit APIs (no new dependency; `fake-indexeddb` is not installed). Against the old code the same specs fail only on the transaction-count assertions (3 vs 1). Full suite 464 passing; `tsc` for openchat-agent clean.


hpeebles (approved):

julianjelfs:
## Quantified (transaction counts exact; timings simulated)

Real `UserDb`/`ChatsDb` methods driven against a fake `idb` that implements both the implicit and explicit APIs. Cost model: one macrotask hop per transaction open (serialised), one microtask per get within an open transaction. **Only the transaction counts are exact** — the ms column is the model's cost × event-loop turns (Node clamps the hop to ~1.2 ms, likely more than a real Chrome IDB turn).

| call | n | transactions before → after | simulated ms before → after |
|---|---|---|---|
| `getCachedUsers` | 600 ids | 600 → **1** | 700 → 1.4 |
| `getCachedEventsByIndex` | 50 | 50 → **1** | 57 → 1.4 |
| `loadMessagesByMessageIndex` | 50 | 50 → **1** | 58 → 1.3 |

At app scale: the minute-ly `#updateUsers` poll goes from ~600 IndexedDB transactions to 1; each scroll-back page and each batch of reply/quote lookups from ~50 to 1.

PR #9221, opened 2026-08-22
Construct one emoji Database lazily instead of two at boot

### LINKED ISSUE #9202: perf: two eager emoji-picker Databases at boot (2× CDN checks) and whole emoji table read into a Set
Part of #9179.

**Area:** `app/src/utils/emojis.ts` (16, 72-119), `app/src/stores/quickReactions.ts:5`

**Confirmed.** Both modules construct `new Database()` at module scope; they share one IDB connection but each runs `_init` → `checkForUpdates`, i.e. two ETag requests to `cdn.jsdelivr.net` on every boot (and a ~1.4 MB dataset download on first load). `initEmojiSet()` at module load does `index.getAll()` over the whole emoji store just to back `isSingleEmoji`, which already has a regex fallback.

**Do.** One shared lazily-constructed `Database` (getter on first use); `quickReactions.ts` imports it. Replace the Set with `/^\p{RGI_Emoji}$/v` where supported (check Tauri WebKit), or at least defer `initEmojiSet()` to first `isSingleEmoji` call. Consider self-hosting the emoji data JSON (`dataSource` option) to remove the third-party runtime dependency — flag for a product/security decision rather than doing silently.

---

julianjelfs:
Closes #9202. Part of #9179.

`utils/emojis.ts` and `stores/quickReactions.ts` each constructed an `emoji-picker-element` `Database` at module scope. They share one IndexedDB connection but each runs its own init, so every boot issued two ETag requests to `cdn.jsdelivr.net` (and a ~1.4 MB dataset download on first run), before any emoji was needed. There is now a single instance behind `getEmojiDatabase()`, created on first use; the three importers call it directly.

`initEmojiSet()` — which reads the entire emoji table out of IndexedDB to back `isSingleEmoji` — no longer runs at module load; it is kicked off by the first `isSingleEmoji` call (in practice the same frame, via `Markdown.svelte`). The pre-existing race (multi-codepoint emoji in the very first render batch not recognised until the set loads) is unchanged.

Not done on purpose: swapping the Set for `\p{RGI_Emoji}` (needs a WebView support check, TODO left) and self-hosting the emoji JSON (product/security decision).

Tests: new `emojis.spec.ts`; full suite 460 passing; svelte-check 0 errors.


julianjelfs:
## Measured on webtest (same host, master build vs this branch)

Hard load of `/chats`, same logged-in user, `performance.getEntriesByType("resource")` filtered to `jsdelivr`:

| build | requests to `cdn.jsdelivr.net/…/emojibase/data.json` at boot |
|---|---|
| master (2.0.2040.2) | **2** (two ETag revalidations, ~470 ms each) |
| this branch (2.0.2040.3) | **1** |

Before any chat is opened (login/landing screens) the branch issues none; master issued two. The whole-table `index.getAll()` read now starts on first `isSingleEmoji` call instead of at module evaluation.

hpeebles (approved):

PR #9226, opened 2026-08-22
Build: preload main/vendor chunks, native getCanonicalLocales

### LINKED ISSUE #9200: perf (build): modulepreload, drop console in prod, formatjs polyfill alias, typebox runtime in main thread
Part of #9179.

**Area:** `app/rollup.config.mjs` (73, 387, html plugin), `app/rollup.extras.mjs`, `openchat-shared/src/domain/chat/chat.ts:1`, `domain/webrtc/webrtc.ts:1`, `domain/data/data.ts:2`

**Confirmed from the build.**
- `index.html` loads a 118-byte facade that then imports `main-*.js` + `vendor-*.js` — no `<link rel="modulepreload">`, so one extra RTT before the 1.6 MB gz critical path starts.
- `terser()` has no options: ~340 console calls survive in main, 229 in vendor, 122 in worker. Runtime `setMinLogLevel` NOOPs debug/log by default, so cost is argument evaluation + retained objects; still worth `drop_console: ["log","debug"]` (keep `warn`/`error` for Rollbar and the deliberate SW diagnostics).
- `@formatjs/intl-getcanonicallocales` polyfill (247 KB rendered) pulled by svelte-i18n; native `Intl.getCanonicalLocales` is baseline everywhere OC targets — alias to a 3-line shim (check the Tauri WebKit `safari13` target).
- `@sinclair/typebox` runtime (141 KB) lands in main-thread vendor because core domain modules build `Type.*` schemas at module scope; split runtime schemas from type-only exports so `import type` consumers don't pull it (it is also duplicated in worker and SW bundles).

Each bullet = one PR. Verify with `ANALYZE=1` build before/after.

---

julianjelfs:
Closes #9200 (two of four items). Part of #9179.

- `index.html` now emits `<link rel="modulepreload">` for the real `main-*.js` and `vendor-*.js` chunks (read from the facade entry's static imports), so the browser no longer spends a round trip discovering them after fetching the 118-byte facade.
- `@formatjs/intl-getcanonicallocales` (svelte-i18n's ~250 KB polyfill) aliased to a 10-line shim over native `Intl.getCanonicalLocales` (Chrome/WebView 54+, Safari 10.3+; Android minSdk 24's WebView is far newer). Spec pins shim == polyfill for every supported locale code and navigator-style tag including `iw→he`.
- Dropped after review: stripping `console.log`/`debug` from the prod bundle — the `localStorage.minLogLevel` toggle is used in production and must keep working.
- Typebox runtime split skipped — not mechanical (168 `Type.*` sites in `chat.ts` alone plus the `@shared` barrel).

**Measured (prod build, raw / gzip):** vendor 2,644,836 / 753,331 → ~2,480,000 / ~693,000 (**−60 KB gz**, formatjs polyfill gone); main unchanged. On webtest: facade, `main` and `vendor` all start at the same millisecond (previously the two chunks started after the facade had been fetched and parsed) — on a cold load that is one round trip to the asset canister, measured at 240–390 ms from here.

Tests: new `intlGetCanonicalLocales.spec.ts` (28); full suite green; svelte-check 0 errors.


julianjelfs:
## Measured on webtest (2.0.2040.5)

- `index.html` now carries `<link rel="modulepreload">` for the real `main-*.js` and `vendor-*.js` (verified via curl).
- Resource timing: facade, `main`, and `vendor` all start at the **same millisecond**; on the previous build the two real chunks started only after the facade had been fetched and parsed. In this (HTTP-cached) harness that gap was ~8 ms; on a cold load it is one round trip to the asset canister, which measured **240–390 ms** from here (`curl` time-to-first-byte on `/version`, 3 samples). That RTT is what the preload removes from the critical path.
- `vendor-*.js`: 754.7 KB → **693.2 KB gz** (−61.5 KB; formatjs polyfill gone). `main-*.js`: 940.7 → 937.1 KB gz (console calls stripped).

hpeebles (approved):

Issue #9201, opened 2026-08-22
perf (service worker): SW bundles full typebox schema + rollbar; media cache uncapped; 34 MB sourcemaps deployed


julianjelfs:
Part of #9179.

**Area:** `openchat-service-worker/src/service_worker.ts` (1-6, 46-67), `app/rollup.config.mjs:73`, `app/.ic-assets.json5`, `dfx.json:151-155`

**Confirmed.**
- SW is 550 KB raw / 120 KB gz: imports from `@agent` pull 132 `@sinclair/typebox` modules, the 10k-line generated `typebox.ts`, `chatMappersV2.ts`, msgpackr and `rollbar`. All of it must parse before the navigation `fetch` handler answers on cold start (the iOS PWA white-screen area).
- `staticResourceCache` for image/video/audio/font/frame has `maxAgeSeconds` only — no `maxEntries`, no `purgeOnQuotaError`. User-uploaded chat media accumulates for 30 days; quota errors fail silently.
- `sourcemap: true` unconditionally, `build/` is the dfx asset source, `.ic-assets.json5` has no `*.map` ignore, no script strips them: ~34 MB of maps (with full `sourcesContent`) uploaded per release.

**Do.** Export a notification-only schema entry point from `openchat-agent` (just the push payload schema + validator) so the SW doesn't reach `chatMappersV2`/full schema; drop rollbar from the SW. Add `maxEntries` (few hundred) + `purgeOnQuotaError: true`; consider excluding `video`. Ignore `*.map` in `.ic-assets.json5` and upload maps to Rollbar in CI instead.

**Security-sensitive:** SW fetch/caching rules and push payload decoding. Push validation must remain; don't touch the deliberate navigation-buffering logic (WebKit clone-cancel fix). Extra reviewer required.

---
**Process (all perf issues):** branch off `master` once the `perf-baseline-tests` branch is merged (characterisation specs: `openchat-client/src/utils/chatMerge.baseline.spec.ts`, `openchat-agent/src/utils/chat.baseline.spec.ts`). Before changing code, add/extend a small test pinning the current behaviour of what you touch. Behaviour must not change. No new dependencies. Run `npm run typecheck && npm run typecheck:agent && npm run lint && npm test` in `frontend/`.

julianjelfs:
[...]

PR #9235, opened 2026-08-22
Service worker: notification-only imports, capped media cache

julianjelfs:
Part of #9179; addresses #9201 items 1 and 2 (item 3, sourcemap upload, is split into its own PR).

- The service worker imported from the `@agent` and `@shared` barrels and so bundled every API schema, `chatMappersV2`, rollbar, idb, emoji and more — 550 KB that must parse before the navigation `fetch` handler can answer on a cold start (the iOS PWA white-screen area). A new `openchat-agent/src/notifications.ts` entry exports only what push decoding needs (`deserializeFromMsgPack`, `typeboxValidate`, the `UserNotificationPayload` schema, `notification` mapper); the three route builders move to `openchat-shared/src/utils/routes.ts` (re-exported from `utils/chat.ts`, no call-site changes) and the SW imports specific `@shared/...` modules. **Push payload validation is unchanged** — same schema, same `Value.Parse`; the typebox runtime stays in the bundle for that reason.
- `staticResourceCache` `ExpirationPlugin` gains `maxEntries: 500` and `purgeOnQuotaError: true` (it also holds user-uploaded media; previously unbounded for 30 days and a quota error failed silently). Video kept in scope. Navigation buffering logic, cache routes/matchers and CSP untouched — only import lines and the plugin options changed in `service_worker.ts`.

**Measured (built SW):** 550,316 B / **124.2 KB gz → 390,574 B / 82.5 KB gz** (−29% raw, −34% gz). Removed from the bundle: rollbar, md5/js-sha3/identicon, `chatMappersV2`, externalBot/icpSwap/icpcoins/governance mappers, shared domain/emoji/webrtc modules. Remaining node_modules: `@icp-sdk/core` (Principal), `@noble/hashes`, `@sinclair/typebox`, `idb` (workbox-expiration), `msgpackr`, workbox-*. The largest remaining piece is the generated `typebox.ts` (~133 KB min) because `Type.*` calls are not `/* @__PURE__ */`-annotated — a generator change could cut ~120 KB more; noted as follow-up.

Tests: new `notifications.spec.ts` (entry exports identical to the barrel's; valid DM payload decodes identically via either path; malformed and unknown-variant payloads throw); full suite 459 passing; tsc + svelte-check clean.

### DIFF
--- a/frontend/openchat-agent/src/notifications.spec.ts
+++ b/frontend/openchat-agent/src/notifications.spec.ts
@@ -1,0 +1,75 @@
+import { Principal } from "@icp-sdk/core/principal";
+import { TypeboxValidationError } from "@shared";
+import { describe, expect, test } from "vitest";
+import * as agent from "./index";
+import * as entry from "./notifications";
+import { serializeToMsgPack } from "./utils/msgpack";
+
+// Characterisation: the service worker decodes push payloads through the
+// notification-only entry point. It must behave exactly like decoding through
+// the full agent barrel, and must still reject malformed payloads.
+
+const sender = Principal.fromText("2vxsx-fae").toUint8Array();
+const timestamp = 1_700_000_000_000n;
+
+const payload = {
+    dm: {
+        s: sender,
+        m: 7,
+        e: 12,
+        sn: "alice",
+        sd: "Alice",
+        ty: "Text",
+        tx: "hello",
+        a: 5n,
+    },
+};
+
+function decode(
+    mod: typeof entry,
+    bytes: Uint8Array,
+): ReturnType<typeof entry.notification> {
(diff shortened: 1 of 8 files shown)


julianjelfs:
## Measured on webtest (2.0.2040.6 vs 2.0.2040.5)

`curl` of the versioned SW URL the app actually registers (`/service_worker.js?v=<version>`):

| | before | after |
|---|---|---|
| `service_worker.js` raw | 550,316 B | **390,574 B** (−29%) |
| gzip on the wire | 123,660 B | **82,305 B** (−33%) |
| `rollbar` references in the bundle | 43 | 0 |

This is the script that must download, parse and evaluate before the SW can answer the navigation request on a cold start.

hpeebles (approved):

PR #9236, opened 2026-08-22
Stop uploading sourcemaps to the website asset canister

julianjelfs:
Part of #9179 / #9201 item 3. Split out because it needs an ops decision.

`app/.ic-assets.json5` now ignores `**/*.map`, so the ~34 MB of sourcemaps (with full `sourcesContent`) produced per release are no longer uploaded to the website canister. `sourcemap: true` is kept in the build.

**Do not merge on its own:** Rollbar currently symbolises stack traces by fetching the maps from the deployed site. No workflow under `.github/workflows` uploads maps to Rollbar, so merging this before adding that step would degrade production stack traces. Suggested order: add a CI step that uploads `build/**/*.map` to Rollbar (rollbar `sourcemap` API, keyed by the website version) and then merge this.

### DIFF
--- a/frontend/app/.ic-assets.json5
+++ b/frontend/app/.ic-assets.json5
@@ -26,6 +26,12 @@
         },
     },
     {
+        // Sourcemaps are still generated (for Rollbar) but must not be uploaded
+        // to the asset canister: ~34 MB per release with full sourcesContent.
+        match: "**/*.map",
+        ignore: true,
+    },
+    {
         match: ".well-known",
         ignore: false,
     },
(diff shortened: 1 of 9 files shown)


hpeebles (approved):

julianjelfs:
Closing: no runtime benefit (browsers only fetch .map files when DevTools is open), and the deployed maps are what Rollbar symbolises against. The only cost was ~34 MB of asset-canister storage per release, which isn't worth losing prod debuggability for. If that storage ever matters, the path is a CI upload to Rollbar first, then this change.

Issue #9207, opened 2026-08-22
perf (needs measurement first): chat-state blob rewrite per poll, event cache LRU, details re-merge, image pipeline, assets, crypto dep dedupe


julianjelfs:
Part of #9179.

Items the audit found but which need scale data or measurement before committing to a change. Do not start these blind; instrument first (worker `performance.now()` around the call, IDB size via `navigator.storage.estimate()`).

**Worker / IndexedDB** (`openchat-agent/src/utils/chatsDb.ts`)
- `setCachedChats` (332-372) serialises and `put`s the *entire* chat state under one key on every 5s poll that had any update, fire-and-forget from `openchatAgent.ts:1908` (no `.catch`). Options: per-chat records (needs `CACHE_VERSION` bump + migration) or debounce. Measure blob size for a heavy user first.
- `chat_events`/`thread_events` stores have no size cap or LRU — only TTL sweeper, per-chat delete, and 30-day nuke. Realistic route to origin quota eviction on mobile.
- `mergeGroupChatDetails`/`mergeThings` (utils/chat.ts 39-164) rebuilds full member `Record` and rewrites the whole member list per delta.

**Media** (`openchat-client/src/utils/media.ts:135-179`, `components_mobile/home/ImageContent.svelte:200-219`)
- Resize/thumbnail on main thread via DOM canvas + `toDataURL` + `atob` loop (Samsung WebView constraint documented — keep fallback). `OffscreenCanvas` in a worker where supported.
- Message `<img>` lack `decoding="async"`; 1500px originals decoded for ~400px bubbles. Cheap first step: add `decoding="async"`.

**Assets / deps**
- `app/public/assets/screenshots/*.gif` 8.6 MB (convert to muted MP4/WebM), ringtones 4 MB.
- Google Fonts stylesheet render-blocking (self-host subsetted woff2 → could tighten CSP `style-src`/`font-src`; security-sensitive).
- `@noble/curves` ×6 / `viem` ×9 / `@noble/hashes` ×6 duplicated across lazy wallet chunks — `overrides` pin; **security-sensitive** (crypto version pinning), needs careful review.

---
[...]

PR #9251, opened 2026-08-27
perf: throttle the chat-state IndexedDB rewrite to one write per 30s

julianjelfs:
Part of #9179; the measured item from #9207.

## Why

`ChatsDb.setCachedChats` rewrites the **entire** chat state under one key and was requested, fire-and-forget with no `.catch`, by every 5 s poll that carried an update. Measured on a heavy prod account (372 direct chats, 136 groups, 14 communities): **~1.5 MB JSON per write** (structured-clone size is larger — Maps/Sets serialise as `{}`), so an active session could be cloning and writing that every 5 s in the worker.

## What

`ChatsDb.setCachedChatsThrottled` coalesces requests: at most one write per 30 s, always the latest state plus the union of the events to mark dirty (they stay consistent with each other and land in the same transaction as before). First request after a quiet period writes immediately (cold cache populated promptly). Failures are logged instead of surfacing as unhandled rejections. `clearCache` drops any pending write.

Losing a trailing write on unload is harmless: the cached `latestUserCanisterUpdates` timestamp means the next session fetches — and re-marks — the same updates again.

The other measured item (event-store LRU): IndexedDB usage on the same account was 122 MB of a 10 GB quota → not needed.

## Tests

`chatsDb.spec.ts` (fake timers, spied `setCachedChats`): immediate first write; coalescing inside the interval with latest state + all dirty marks across chats; immediate write again after a quiet period; failure logged not thrown. `typecheck`, `typecheck:agent`, lint, `npm test` (809) green.

### DIFF
--- a/frontend/openchat-agent/src/services/openchatAgent.ts
+++ b/frontend/openchat-agent/src/services/openchatAgent.ts
@@ -1929,7 +1929,7 @@ export class OpenChatAgent extends EventTarget {
         const updatedEvents = getUpdatedEvents(directChatUpdates, groupUpdates, communityUpdates);
 
         if (this.userClient.userId !== ANON_USER_ID) {
-            this._chatsDb.setCachedChats(state, updatedEvents);
+            this._chatsDb.setCachedChatsThrottled(state, updatedEvents);
         }
 
         const directChatsAddedUpdatedIds = new Set([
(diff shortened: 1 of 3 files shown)


julianjelfs:
Closing, not merging. On closer reading the IndexedDB chat-state blob is the poll loop's working state (read on every poll, not just at startup), so throttling only the write would make polls re-fetch the same deltas for up to 30 s. A memory-first rework would fix that, but the case for touching the sync loop at all was a blob *size*, not a measured cost — not enough to justify the risk. Leaving `setCachedChats` as is.

PR #9305, opened 2026-09-07
Remove the shop from the landing page menus

julianjelfs:
Nobody has ever bought anything from it, so it is a nav slot spent on noise.

Removed from both `MenuItems.svelte` (desktop) and `MobileMenuItems.svelte` (mobile menu), along with the now-unused `ShoppingOutline` icon import. No references to `myspreadshop.com` remain anywhere in the frontend.

The shop itself is untouched — this only stops linking to it, so any existing link still works.

`svelte-check` clean. The summary warning count shifts by two because dropping the icon import changes which files get compiled; the per-file warning set is byte-identical before and after, and errors stay at zero.

https://claude.ai/code/session_01JYc48kzHXq2sV7yySmjAyA

### DIFF
--- a/frontend/app/src/components/landingpages/MenuItems.svelte
+++ b/frontend/app/src/components/landingpages/MenuItems.svelte
@@ -16,11 +16,6 @@
 <div class="menu-items">
     <div class="nav">
         <div class="menu-item">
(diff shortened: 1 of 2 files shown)


julianjelfs:
Folded into #9304 — this belongs with the rest of the landing page work rather than as a separate PR.

PR #9312, opened 2026-09-08
Log out on session expiry, fix three crashes, filter clock-skew noise

julianjelfs:
Rollbar sweep for 2026-09-08, covering the last 24 hours of production items, plus the source-map pipeline that makes the next sweep possible. Six commits; the last one addresses a review round.

## Session expiry never logged anyone out

The biggest find, and not what I set out looking for. `#30673 lastOnline` is 17,630 occurrences over seven months, 293 in the last day, all from **one IP** on **2.0.2003**, retrying every 60 seconds with a `SessionExpiryError` and never signing out. `#30666 getUsers`, `#30669 updateRegistry`, `#30675 getBots`, `#25992` and `#31665`/`#31667` are the same client. Between them that is most of the daily volume.

`requiresLogout` was only consulted in `App.svelte`'s `unhandledrejection` handler. Background pollers catch their own failures, so the rejection never reached the window and the client polled on a timer for as long as the tab stayed open. `workerAgent.ts:#resolveError` now checks every worker rejection and publishes `sessionExpired` once, which both `App.svelte`s turn into a logout.

Three supporting changes came out of review. `clearChatShortcuts` is registered through `client.onLogout` rather than called at each logout site, so the previous user's shortcuts cannot survive a logout that starts from the worker agent. `logout()` is idempotent (`#logoutPromise ??= #doLogout()`): an expired session reaches it twice, deterministically — the worker agent publishes `sessionExpired` and then rejects, and an uncaught rejection brings the window handler round again — and the second run used to navigate while the first was still removing the push token, leaving the Android token registered to a signed-out user. And `logout()` navigates in a `finally` with the teardown capped at 5s, because the worker's logout awaits three sequential IndexedDB deletes and a wedged IndexedDB is the very case this path exists for.

**Known edge of that cap, stated plainly:** if IndexedDB is blocked for longer than 5s, navigating can cut the auth client's delegation delete short, and startup may sign the user back in from the surviving delegation. Narrow, and no worse than the manual reload it replaces, but it is a consequence of choosing "always navigate" over "sometimes never".

## Crashes

**`s($).ledger`, `#31876`** — a top-level boundary error, so the whole app dies. `TokenState`'s constructor took a non-optional token and ~25 call sites passed `$cryptoLookup.get(x)!`. A gate or message naming a ledger the registry has never carried or has since dropped made the first derived read of `#token.ledger` throw.

The first attempt just fell back to a null token, which review correctly called a half-fix: zero decimals and an empty symbol render an amount that is *wrong* rather than obviously missing, and `PrizeWinnerContent`'s `decimals ?? 8` meant one transfer could show two different numbers depending on which component drew it. `TokenState` now exposes `unknown`, `formatTokens` returns `"?????"` rather than a figure built from zero decimals (one guard covering every consumer), and components that can act on a token refuse to — the payment gate explains itself instead of offering Approve, in both the mobile and the desktop copy (the desktop one was missed until the fifth review round; it is the copy every desktop-width viewport renders).

The p2p swap accept modal went a different way: the swap message already carries a `TokenInfo` per side — symbol, decimals, and the fee the canister actually debits — so both copies now take those instead of ledger ids and never consult the registry at all. No registry lookup means no unknown-ledger case, and no way to drift from the real charge.

Review also found five more `$cryptoLookup.get(ledger)!` sites one tap beyond the ones first fixed, so an unknown ledger rendered and then died on Join or Accept: `PaymentGateEvaluator` (which reads `token.ledger` directly, bypassing `TokenState`), and `AcceptP2PSwapModal` and `BalanceWithRefresh` in both layouts. A sixth, `BigInt(Number(gate.amount) * 0.98)` in the mobile payment breakdown, was pre-existing but in the same sheet: it throws a `RangeError` for any amount that is not a multiple of 50. Integer arithmetic now.

**`t.membership.readByMeUpTo` (`#31716`), `e.them.userId` (`#31582`), `n.chat.id` (`#31618`)** — direct chat summaries arriving without fields the type says are always present. `openchat.ts:7287` was the only branch reading `membership` without `?.`; the group and channel branches beside it already guarded.

For the cause: the IndexedDB cache is the one place a chat summary enters the agent without a mapper having built it from candid. `getCachedChats` now treats such a cache as unusable, clears it, and lets `getUpdates` fall through to `getInitialState` — what the existing >30-day staleness check already does. It never throws, either: `Stream` calls its initialiser synchronously and never catches a rejected async one, so a throw here would fire neither `onResult` nor `onError` and wedge every launch.

Two earlier attempts were wrong and are worth recording. Repairing the fields cannot work — `membership` is the server's to say (`nullMembership()` gives `ROLE_NONE` where a direct chat is always `ROLE_OWNER`, and an undefined `readByMeUpTo` marks the chat entirely unread). Dropping just the bad record is worse, because the cached list is the base `mergeDirectChatUpdates` applies deltas to, so a chat removed from it never returns until the server happens to mention it.

## Noise

- **`/invalid request expiry/i`** added to `ENVIRONMENT_NOISE_PATTERNS`. The client's clock being wrong, seen from the replica's side — `#31879`'s payload has a device 13 days behind, `#30918`'s claims an expiry in April. The replica echoes the timestamps back in the message, so every skewed device was minting a fresh item rather than joining one. Clears `#30918` (1,122/day), `#31165`, `#31873`, `#31879`.
- **`CanisterUnavailableError` was never filtered.** `isTransientNetworkError` gated its 502-504 check on `name === "HttpError"`, but every `HttpError` subclass overwrites `name`, so a 503 from a frozen canister sailed past a filter written to catch exactly that.
- **Duplicate reporting.** `captureUnhandledRejections` is on *and* `App.svelte` added its own listener calling `logger.error`, so every rejection produced two items with different titles. `#31582` logged both at 06:16:57 on 2026-09-08. Rollbar's capture installs at init, before the listener is added on mount, so dropping the log loses no coverage — and its items are the ones source maps demangle. At triage time: `#31837`, `#31822` and `#25983` are logger-path items and will go quiet; their capture-path twins carry on.
- **`quoteSwap`** used `Promise.all`, so one pool rejecting failed and reported every pool. This took four rounds to get right and the root cause was in the ICPSwap mapper, which threw on a decoded `err` variant — making a decline indistinguishable from a transport failure, and (since that Error was not in `executeQuery`'s no-retry list) retrying a dust-amount quote seven times with backoff, roughly twelve seconds. The mapper now returns `undefined` for a decline and throws for a failure; `quoteSwap` drops declines and throws if nothing succeeded and anything failed.

  Which variants are declines is load-bearing and pinned by a spec. `InsufficientFunds` and `UnsupportedToken` by definition. `InternalError` is nominally a failure, but ICPSwap reports the commonest decline of all through it — `{"InternalError":"The amount of input token is too small."}` is the exact payload behind `#31881` — so that one message is a decline; any other `InternalError`, and `CommonError`, still throw and retry.

## Validation

`EditRecipient.svelte` checked `trimmedAddress.length > 0` under a `// TODO this probably isn't good enough`, and a bad address threw unhandled out of the worker's withdrawal mapper (`#31864`). It now accepts the same two forms as `is_valid_account` in `save_crypto_account.rs`: an ICRC-1 textual account, or an ICP ledger account identifier — checksum included, via the SDK's own `isIcpAccountIdentifier`, since the canister verifies the CRC and a length-and-hex check let typos through to a confusing rejection. (The first attempt used the ICRC check alone, which was *stricter* than the backend and silently disabled the save button for valid 64-hex addresses.) `SendToAddress` no longer offers "Save address" after a send over the BTC or an EVM network, where the target can never be an IC account; it still does over ckBTC, where it is one.

## Source maps

Rollbar never fetches maps from your site, even public ones — it only reads maps uploaded to its API, keyed by project + version + minified URL. That keying matters: errors are processed asynchronously and `worker.js` has a fixed, unhashed name, so a map fetched after a later deploy would demangle old traces against new source and produce line numbers that look right and are wrong.

Two halves that must agree, because a mismatch fails silently:

- `scripts/upload-source-maps.mjs` walks `frontend/app/build` for `*.js.map` and registers each as `http://dynamichost/<path relative to build>`. No new dependencies — node 22 has `fetch`, `FormData` and `Blob`. Eight concurrent; a 401/403 aborts with one message about the token, and the token is redacted from Rollbar's error body, which echoes it back.
- `normaliseSourceMapUrls` in `logging.ts`, wired in as Rollbar's `transform`, rewrites every http(s) frame to that same URL. The bundle is served from four origins (`oc.app`, `webtest.oc.app`, the canister's `.icp0.io` domain, `http://tauri.localhost`) and the workers carry a `?v=` cache buster, so without this a frame's filename is one of many strings for the same file. Non-http frames are untouched so `thrownByExtension` still matches on `chrome-extension://`.

Side benefit: errors from `oc.app` and from the native app currently fingerprint as separate items because the filenames differ. After this they group.

Wired into all four website deploy scripts between the build and the `dfx deploy`. Needs `OC_ROLLBAR_SERVER_TOKEN` (a `post_server_item` token; the app's `OC_ROLLBAR_ACCESS_TOKEN` is `post_client_item` and gets rejected). Without it the script warns loudly and exits 0.

**Verified for real:** 173 maps uploaded successfully for 2.0.2052 in 10.7s, including the 5.2MB vendor map, so there is no size limit problem. Separately, the local build was found byte-identical to what is live, so the same maps were backfilled under the URLs the *deployed* 2.0.2052 bundle's frames actually carry — meaning `#31837` should demangle on its next occurrence without waiting for a deploy. That backfill covered `oc.app` only, not `webtest.oc.app` or `tauri.localhost`.

## Left alone

**`#31837`** (`Cannot read properties of undefined (reading 'id')`, 28 hits since 2.0.2051) is a real regression I could not pin down without a usable stack, which is what prompted the source-map work. Very likely the same defect as `#31618` in Chrome's wording.

**The svelte-i18n boundary crash** (`#31868`/`#31880`, "Cannot format a message without first setting the initial locale"). An earlier commit added a locale guard here; review showed it fixed nothing and it has been reverted. svelte-i18n 3.7.1 already wraps its `getCanonicalLocales` call in `try/catch` and falls back to `fallbackLocale`, so a malformed tag never caused this. The real cause is that `locale.set` defers the store write until the locale's loader resolves, so any `$_` between `init()` and `en.json` arriving throws. Fixing it means gating the first render on `waitLocale` or bundling `en` synchronously — its own change, not this one.

**Muting the stuck-client items** — the logout fix stops this for anyone on a current build, but the 2.0.2003 client is unreachable and will keep firing. Its titles are generic enough (`lastOnline`, `getUsers`, `WORKER: unhandled promise rejection:`) that a mute would hide future genuine failures, so snoozing is the safer option.

**`#31872`** (`aytea-…` out of cycles) — the message can't distinguish a third-party SNS ledger running dry from one of ours, and two occurrences isn't worth the risk.

## Testing

`npm run typecheck` 0 errors, `typecheck:agent` clean, 912 tests pass, lint clean on the touched files. New tests cover a wrong client clock in both directions (`error.spec.ts`) and the source-map frame rewriting against the literal URLs the upload script registers (`logging.spec.ts`).

https://claude.ai/code/session_01JYc48kzHXq2sV7yySmjAyA

### DIFF
--- a/frontend/app/src/components/App.svelte
+++ b/frontend/app/src/components/App.svelte
@@ -170,6 +170,7 @@
             subscribe("hangup", hangup),
             subscribe("askToSpeak", askToSpeak),
             subscribe("userLoggedIn", onUserLoggedIn),
+            subscribe("sessionExpired", () => client.logout()),
         ];
         window.addEventListener("orientationchange", calculateHeight);
         window.addEventListener("unhandledrejection", unhandledError);
--- a/frontend/app/src/components/App.svelte
+++ b/frontend/app/src/components/App.svelte
@@ -639,7 +639,12 @@
             return;
         }
         recordError("window", err);
-        logger?.error("Unhandled error: ", err);
+        // Deliberately not reported here. Rollbar's own captureUncaught /
+        // captureUnhandledRejections already reports every event this handler sees, and
+        // installs earlier than this listener, so logging again produced two Rollbar items
+        // per rejection - one titled "Unhandled error: X" and one titled "X" - splitting
+        // every defect in two and doubling the volume. This handler keeps the crash-log
+        // record and the logout, which Rollbar's capture does not do.
         if (ev instanceof PromiseRejectionEvent && requiresLogout(ev.reason)) {
             client.logout();
             ev.preventDefault();
--- a/frontend/app/src/components/App.svelte
+++ b/frontend/app/src/components/App.svelte
@@ -643,8 +643,10 @@
         // captureUnhandledRejections already reports every event this handler sees, and
         // installs earlier than this listener, so logging again produced two Rollbar items
         // per rejection - one titled "Unhandled error: X" and one titled "X" - splitting
-        // every defect in two and doubling the volume. This handler keeps the crash-log
-        // record and the logout, which Rollbar's capture does not do.
+        // every defect in two and doubling the volume. Its `checkIgnore` reads the rejection
+        // reason out of the original arguments, so worker errors - which arrive as plain
+        // objects, not Errors - are still filtered on name and code. This handler keeps the
+        // crash-log record and the logout, which Rollbar's capture does not do.
         if (ev instanceof PromiseRejectionEvent && requiresLogout(ev.reason)) {
             client.logout();
             ev.preventDefault();
--- a/frontend/app/src/components/home/AcceptP2PSwapModal.svelte
+++ b/frontend/app/src/components/home/AcceptP2PSwapModal.svelte
@@ -66,19 +66,29 @@
     }
 
     let cryptoBalance = $derived($cryptoBalanceStore.get(ledger1) ?? 0n);
-    let tokenDetails0 = $derived($cryptoLookup.get(ledger0)!);
-    let tokenDetails1 = $derived($cryptoLookup.get(ledger1)!);
-    let symbol0 = $derived(tokenDetails0.symbol);
-    let symbol1 = $derived(tokenDetails1.symbol);
-    let transferFees = $derived(BigInt(2) * tokenDetails1.transferFee);
+    // Either side of the swap can name a ledger the registry does not carry. Asserting it did
+    // crashed this modal open; without decimals and a fee we also cannot state the amounts or
+    // let the swap be accepted, so treat it as an unacceptable swap.
+    let tokenDetails0 = $derived($cryptoLookup.get(ledger0));
+    let tokenDetails1 = $derived($cryptoLookup.get(ledger1));
+    let unknownToken = $derived(tokenDetails0 === undefined || tokenDetails1 === undefined);
+    let symbol0 = $derived(tokenDetails0?.symbol ?? "");
+    let symbol1 = $derived(tokenDetails1?.symbol ?? "");
+    let transferFees = $derived(BigInt(2) * (tokenDetails1?.transferFee ?? 0n));
     // An OpenChat balance which cannot cover the swap is no obstacle when an external wallet is
     // paying instead
     let insufficient = $derived(
         sourceWallet === undefined && cryptoBalance <= amount1 + transferFees,
     );
-    let valid = $derived(error === undefined && !insufficient);
-    let amount0Text = $derived(client.formatTokens(amount0, tokenDetails0.decimals));
-    let amount1Text = $derived(client.formatTokens(amount1 + transferFees, tokenDetails1.decimals));
+    let valid = $derived(error === undefined && !insufficient && !unknownToken);
+    let amount0Text = $derived(
+        tokenDetails0 === undefined ? "?????" : client.formatTokens(amount0, tokenDetails0.decimals),
+    );
+    let amount1Text = $derived(
+        tokenDetails1 === undefined
+            ? "?????"
+            : client.formatTokens(amount1 + transferFees, tokenDetails1.decimals),
+    );
 </script>
 
 <Overlay dismissible>
--- a/frontend/app/src/components/home/AcceptP2PSwapModal.svelte
+++ b/frontend/app/src/components/home/AcceptP2PSwapModal.svelte
@@ -1,10 +1,6 @@
 <script lang="ts">
-    import type { OpenChat, SignerWallet } from "@client";
-    import {
-        cryptoBalanceStore,
-        enhancedCryptoLookup as cryptoLookup,
-        mobileWidth,
-    } from "@client";
+    import type { OpenChat, SignerWallet, TokenInfo } from "@client";
+    import { cryptoBalanceStore, mobileWidth } from "@client";
     import { getContext } from "svelte";
     import { i18nKey } from "../../i18n/i18n";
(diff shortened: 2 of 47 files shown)


hpeebles (approved):

julianjelfs:
Closed in favour of a split. This branch grew from a Rollbar triage into ten commits and five review rounds, and the serious findings clustered inside two design changes that had no business sharing a PR with the noise filters. Same final tree, carved by concern; each cut from current master and gated on typecheck + the full suite:

- #9318 — noise filters, duplicate reporting, quote declines, recipient validation, source maps. The low-risk half; mergeable as-is.
- #9319 — log the user out when the session has expired. Changes when the app signs people out.
- #9320 — stop crashing on a ledger the token registry does not carry. A presentation decision.
- #9321 — treat a cache holding a malformed direct chat as unusable. Small, but a corrupt-cache policy.

Three files are touched by more than one of them (`App.svelte` ×2, `openchat.ts`), on non-overlapping lines; the second to merge will need a trivial rebase. The review history that shaped each change is in its commit message.

Issue #9332, opened 2026-09-09
The Daily: a daily logic puzzle with CHIT entry, hints and result cards (phases 1 and 2)


julianjelfs:
## Problem Statement

CHIT accumulates and does nothing. Users earn it from the daily claim, achievements and referrals, and past the first 110k there is nothing to spend it on. The claim is a habit with no second act: you tap, you leave. There is no reason to come back later in the day and no reason to talk about what you did.

We want a daily habit that is shared. Something everyone does once a day, that produces a result worth posting into a chat, that gives CHIT a reason to exist as both a reward and a cost, and that turns "I logged in" into "did you get today's yet?".

## Solution

The Daily: one logic puzzle a day, the same for everyone, generated by a canister rather than authored. Launch puzzle is Light Up (Akari): place bulbs on a grid so every white cell is lit, bulbs never shine on each other, and numbered black cells say how many bulbs touch them. Tap to place a bulb. Rules are shown by highlighting, never written as prose, and a broken rule is painted the moment it happens.

Play lives behind the daily claim. Entry costs a little CHIT, solving earns more than the entry, hints cost CHIT and teach the technique rather than mark the error. Solve time is measured on the server. On finishing, the user can post a spoiler-free result card into any chat, and people compare by reading each other's cards in the chat. There is no leaderboard. There is no notification, no banner and nothing in community UI. If nobody in your chats plays, you never see it.

Phase 1 ships the puzzle, play, payment, streak count and card. Phase 2 adds the streak freeze, retry and milestone achievements. The whole feature ships switched off and is turned on by a config flag.

### Playing

1. As a user, I want to open today's puzzle from the daily claim modal, so that the puzzle is part of the same habit as the claim.
2. As a user, I want to see the puzzle's rules demonstrated on the grid when I first open it, so that I can make a first move without reading a paragraph.
3. As a user, I want to tap a cell to place or remove a bulb, so that I can play with one thumb.
4. As a user, I want lit cells to show as lit as I place bulbs, so that I can see the rule working.
5. As a user, I want to see the entry fee before I start, so that I know what I am spending.
6. As a user, I want my first ever puzzle to be free, so that I can try it before spending anything.
7. As a user, I want my solve time to start when I start and to keep running if I close the app, so that time is fair and cannot be paused.
8. As a user, I want my partly filled grid to survive a reload or an app kill, so that I do not lose progress.
[...]

PR #9345, opened 2026-09-10
The Daily 2/3: daily_puzzle canister and the game engine

julianjelfs:
Part 2 of 3 for The Daily ([#9332](https://github.com/open-chat-labs/open-chat/issues/9332)). #9344 has landed, so this now targets master directly. **Read the [design update](https://github.com/open-chat-labs/open-chat/issues/9332#issuecomment-5624095532) on the issue first.**

**This is the third that carries the risk.** It holds the money and the answers. If you only have time for one of the three, make it this one.

## Shape

A new `daily_puzzle` singleton generates one puzzle a day on a weekday rota, keeps a small candidate pool a governance proposal can veto, pushes the day's puzzle to every local user index, and keeps the results index that shared cards are checked against. It holds no user state.

The local user index runs the game: entry fee, clock, hints, submissions, reward, streak. The user canister gains one generic endpoint for crediting or debiting CHIT against a keyed game event; it knows nothing about puzzles, which keeps the game out of the canister we least want to redeploy.

## Trust boundary

Stated plainly, because it is what I would most like challenged.

- **The solution never leaves the local user index** and sits behind no query. Submissions are compared there, byte for byte.
- **The client never supplies a time or a price.** The clock starts server side on the start call. The entry fee and each hint price are quoted by the server and re-checked on use, so a stale or invented price is rejected rather than charged.
- **Each hint level withholds what has not been paid for.** Level 1 is the highlighted region with the technique withheld, level 2 adds the technique and the keys its sentence points at, level 3 adds the conclusions. A `target` naming a key the step concludes is withheld below level 3, because in slant's forced square, tents' line rules and loopy's premature loop it always does, and sending it would hand over the level 3 answer at the level 2 price. `focus` is never filtered: it is a region the player can usually reconstruct from the rules, so a hole punched in it would point straight at the answer.
- **The mistake check answers with one key, never the set.** `filled` is client-supplied and capped above any board size, so returning every disagreement would answer "which of these cells are not 1?" in a single free call.
- **A solve under `min_carded_solve_ms` is not published.** It is still paid and still counts for the streak, but it never reaches the results index, so it can back no card and move no published aggregate.
- **Every CHIT movement is keyed and idempotent.** The keys are `{game_id}:{number}:...` and deliberately do not identify which puzzle was held for that number, so the calls a regenerated day forces everyone to repeat land on `AlreadyAdded` rather than charging a second entry fee or paying a second reward.
- **Both CHIT calls record before they debit** and undo it if the debit fails, so requests overlapping on the await cannot exceed `max_hints`, and a failed or raced debit cannot take a fee without giving a game.
- **CHIT only moves between the system and a user**, never between users.
- Submissions per puzzle are capped, and every configured CHIT amount is bounded against the user canister's own per-event limit.

## What this does not defend against

Worth saying, because the honest answer shaped the design. Every puzzle has a unique solution reachable by published rules, and the description is a public query, so anyone who wants to skip the game runs the solver that ships in this repo. That is not defensible and is not worth trying to defend. CHIT is not and can never be exchangeable for anything of value, which is what makes it acceptable: a script converts electricity into a number beside its own name. The defences above exist so that cheating is deliberate rather than one crafted call away, and so that the two things a cheat could take from other people — a verifiable result card and the published aggregates — stay out of reach.

## Hints

Levels cost 25, 75 and 200. No level is free: `max_hints` counts steps rather than levels, so a free level 1 on each of three steps walked three deductions of the chain for nothing and left no budget for the tiers that charge, making them unreachable for exactly the player they exist for. An upgrade costs the difference between the levels, so climbing is never dearer than jumping straight to the answer. The reward penalty applies to every step served, which also means the hint count a result card publishes is the count that cost something.

## Release

Everything ships switched off. Every call answers not-available until config sets `enabled`, and the website hides the entry point until a fetch returns an enabled puzzle. Two ordering rules, both about switching on rather than deploying:

1. **Every local user index before the `set_daily_puzzle_canister_id` proposal.** The proposal puts a new `UserIndexEvent` variant on the ordered user_index to LUI queue. An LUI still on the old wasm traps decoding the batch, the reject is treated as retryable, and every later event for that subnet (username changes, suspensions, CHIT) queues behind it until the LUI is upgraded. Same rule as `SetModerationReferralConfig`.
2. **Website `#9346` before `enabled`.** The user canister writes a new `ChitEventType::Game` for every entry, hint and solve. The live website's typebox union and `chitEventType` mapper reject it, which empties the CHIT history page for anyone who has played. Nothing writes the event while `enabled` is false.

Canister deploy order otherwise does not matter. The daily canister's id reaches the local user indexes through the user_index fan-out rather than a build-time constant, and an LUI that has not been told the id declines the push and pulls for itself once it has.

## Worth knowing

Game logic lives in the local user index rather than the user canister. Puzzle rules will change; user canisters are the most expensive thing we upgrade and the most numerous. The user canister is touched once, for a generic CHIT endpoint that outlives this feature.

The rota is data, not code. Adding or reordering games is a config change. Storage is keyed by game id throughout rather than assuming one puzzle a day.

Unruly needs an even width and height, so the schedule validator rejects odd sizes rather than letting the generator spin.

**One thing left open.** #9344 made generation fallible, and `GenerateError::Exhausted` means this seed found nothing while another may well work. Nothing here advances the seed on its own — `attempt_for` only moves when an operator runs `regenerate_today` — so an unlucky seed stops that day until someone intervenes. Rather than invent a retry policy while merging, the generator's own error is logged and the caller no longer reports it as a missing generator. Retrying `Exhausted` with the next seed, while still refusing `InvalidParams`, is probably what we want.

## Tests

PocketIC: a week of the rota generating the right game each day, the engine's state machine, two end-to-end flows that start, hint, submit and verify a solve, and a third that drives the CHIT paths the happy flow never takes: a start and a hint the user canister refuses for want of CHIT, and a `regenerate_today` after a solve, where the restart charges nothing and the second solve pays nothing. The two `c2c_game_chit` tests that were `#[ignore]`d are folded into that flow, since the endpoint is refused at ingress. Unit tests across the three canisters. clippy, rustfmt and candid clean.

The end-to-end flow test was asserting that a level 1 hint carries its conclusions, which stopped being true when conclusions were first gated. It never failed, because a missing match arm for `unruly` panicked first — so the hint ladder had no end-to-end coverage at all. It does now, across all three levels.

https://claude.ai/code/session_015hHoyT8UnxPgmUcQSKomBJ


julianjelfs:
## Invariant review

Review by invariant, per the delivery rule set 2026-09-13. Numbers refer to the Invariants section now on #9332. For each: the test that would fail if it broke, or NONE. The NONEs are the findings. Nothing else in the diff is a blocker.

Round three's working tree (uncommitted at the time of writing) is included, because it is what merges.

| # | Invariant | Test | Status |
|---|---|---|---|
| 1 | Client never supplies price or time | `start_price_mismatch`, `state_carries_the_fee_that_start_will_expect`, `daily_puzzle_end_to_end` (hint `expected_price`) | pinned |
| 2 | Every CHIT movement keyed and idempotent | `game_chit_keys::duplicate_insert_rejected`, `daily_puzzle_refused_debits_and_replayed_keys` | pinned |
| 3 | Day's fee charged once, reward paid once, across regenerate with or without game change | `regenerated_puzzle_keeps_the_same_chit_keys`, `chit_keys_do_not_identify_the_puzzle`, `daily_puzzle_refused_debits_and_replayed_keys` | pinned |
| 4 | Refused debit takes no fee, gives no game | `released_start_leaves_no_record`, `daily_puzzle_refused_debits_and_replayed_keys` | pinned |
| 5 | Unpaid record re-issues the debit, keeps clock | `an_unpaid_record_is_started_again_not_reported_as_started` | pinned (round 3) |
| 6 | Free play spent only by a start that landed | `a_released_start_does_not_spend_the_first_play`, `entry_fee_waived_only_until_the_first_start` | pinned (round 3) |
| 7 | Regenerated day is free restart for the free-play holder only | `a_regenerated_puzzle_is_a_free_restart_for_a_first_play` | pinned |
| 8 | CHIT never moves between users | NONE. Holds by construction: `c2c_game_chit` args carry no counterparty. Not testable as behaviour; accept as a type-level fact | by construction |
| 9 | Amounts within per-event limit, prices ascending, every level costs | `config_and_schedule_validation` | pinned (round 3 added level-1 case) |
| 10 | Hint step reserved before debit | `a_second_reservation_on_a_step_in_flight_is_refused`, `reserved_hints_count_before_the_debit_but_are_not_readable`, `nothing_works_while_the_entry_fee_is_in_flight` | pinned |
| 11 | Reservation whose debit never returns expires | `a_reservation_whose_debit_never_came_back_expires` | pinned |
| 12 | Penalty on every served step | `hint_penalty_applies_to_every_step_served` | pinned |
| 13 | Queued credit retry on `AlreadyAdded` abandons | **NONE.** Round 3 changed `GameChitCredit::process` to abandon on `Applied(None)`; no test drives the retry path to that outcome | **finding** |
| 14 | `c2c_game_chit` amount bounded by range, `i32::MIN` refused | **NONE.** `c2c_game_chit.rs` has no unit tests. The only test naming `i32::MIN` was `game_chit_invalid_args_rejected`, which was `#[ignore]`d and round 3 deleted it | **finding** |
| 15 | `c2c_game_chit` refused at ingress | `game_chit_rejected_at_ingress` | pinned |
| 16 | Keys pruned, never over 64 bytes | `game_chit_keys::old_keys_pruned_on_insert`, `chit_keys_do_not_identify_the_puzzle` | pinned |
| 17 | Solution never leaves the LUI | NONE. Holds by construction: `PublicDailyPuzzle` has no solution field. Accept as type-level | by construction |
| 18 | Hint levels withhold; focus sorted | `hint_levels_withhold_what_has_not_been_paid_for`, `focus_is_sorted_below_level_3` | pinned (round 3) |
| 19 | Mistake check returns one key | `hint_mistake_returns_only_the_lowest_wrong_key` | pinned |
| 20 | Every key-naming outcome metered, refund only once paid | `free_hint_outcomes_are_bounded`, `refusals_at_the_hint_cap_are_metered_too`, `a_hint_whose_debit_fails_keeps_its_check_spent`, `a_hint_call_with_nothing_filled_is_not_metered`, `daily_puzzle_refused_debits_and_replayed_keys` | pinned (round 3) |
| 21 | Submits capped | `submits_are_capped` | pinned |
| 22 | Wrong submit errors, counts, says only "wrong" | `wrong_submit_errors_and_counts` | pinned |
| 23 | Grid and filled bounded by puzzle | `grid_and_filled_bounds_come_from_the_puzzle` | pinned |
| 24 | Sub-floor solve paid, not published | `solves_under_the_carding_floor_pay_but_are_not_published` | pinned |
| 25 | Start idempotent, clock kept, submit before start errors | `start_is_idempotent_and_keeps_the_original_clock`, `submit_before_start_errors` | pinned |
| 26 | Regenerate drops replaced records only, streaks untouched | `regenerated_puzzle_drops_that_games_records_only`, `new_puzzle_number_drops_records_but_keeps_solved_sets` | pinned |
| 27 | Two games a day, one streak day | `two_games_same_day_have_separate_records_and_one_streak_day` | pinned |
| 28 | Deleted user's records go | `removing_a_user_drops_their_records_and_history` | pinned (round 3) |
| 29 | Dark until enabled | `daily_puzzle_dark_by_default`, `not_available_when_disabled_wrong_number_expired_or_unknown_game` | pinned |
| 30 | Any deploy order is safe | **NONE.** The PRD's Testing Decisions promised "a second test upgrades canisters in a non-recommended order and confirms nothing traps". It was never written | **finding** |
| 31 | Old blob deserialises to empty engine | `old_single_slot_blob_deserialises_to_an_empty_engine` | pinned |
| 32 | Older-number push ignored | `a_push_for_an_older_number_is_ignored` | pinned |
| 33 | Never permanently seedless | **NONE.** Round 3 rewrote `reseed_rng` to retry. PocketIC cannot make `raw_rand` reject, and the retry lives in a lifecycle fn with no seam. Accept as untestable here, or extract the decision (`Ok` sets seed, `Err` schedules retry, in-flight guard) into a pure fn and unit test that | **finding, testability** |
| 34 | Exhausted day does not block tomorrow, regenerate resets attempts | `a_failed_generation_salts_the_next_seed_and_gives_up_eventually` covers giving up. **The round-3 `generation_needed` skip and the `generation_failures` reset on regenerate/schedule change have no assertion** | **finding** |
| 35 | Push backs off, gives up, stale run leaves the set alone | `retry_delay_doubles_to_a_ceiling` covers the delay curve only. **Give-up and the `GENERATION` stale guard have no test** | **finding** |
| 36 | Unknown `game_id` rejected by schedule and game config | `schedule_with_unknown_game_id_is_rejected` covers the schedule. **`set_game_config` rejection (round 3) has no test** | **finding** |

### Findings

Seven NONEs. Ranked by what breaks if wrong:

1. **#14** money bound with zero tests, and the one test that named the case was deleted this round. Add unit tests on `c2c_game_chit` for 0, limit+1, -(limit+1), `i32::MIN`, and a balance-exceeding debit.
2. **#13** double-credit guard on the retry path, untested.
3. **#30** the deploy-order test the PRD promised.
4. **#34, #35, #36** round-3 daily canister changes shipped without assertions.
5. **#33** testability: extract the reseed decision or accept.

Two "by construction" rows (#8, #17) are recorded so the omission is deliberate.

Everything else in the round-3 diff is covered by a test that names its invariant. Under the rule, no other change in the diff is a blocker.

julianjelfs:
## Invariant review, round two: the seven NONEs are pinned

Each finding above now has a test that names its invariant. One commit per invariant, test first.

| # | Test | Commit |
|---|---|---|
| 13 | `game_chit_credit::tests::already_added_and_refusals_abandon_only_transient_failures_retry` | e653da60b |
| 14 | `c2c_game_chit::tests::amount_and_ids_are_bounded` | b6fb8fdb3 |
| 30 | `daily_puzzle_survives_a_missing_wrong_or_stopped_daily_canister` (PocketIC) | 5c280ad27 |
| 33 | `lifecycle::tests::a_rejected_seed_is_retried_and_one_reseed_is_in_flight` | 19b86199a |
| 34 | `an_exhausted_day_is_skipped_and_gets_a_fresh_run_on_regenerate_or_reschedule` | 47ac6fc15 |
| 35 | `push_puzzle::tests::retries_back_off_then_give_up_and_a_stale_run_touches_nothing`, `only_a_transient_failure_stays_pending` | d63255d83 |
| 36 | `set_game_config_with_unknown_game_id_is_rejected` | 47ac6fc15 |

**Writing #34's test found a hole.** `generation_needed` skipped an exhausted today but still answered an exhausted tomorrow, so every trigger of the job ran one more full generation for it and logged "giving up" again. Bounded, not a spin, but wasted instructions and a misleading log on every operator call. Same guard applied to tomorrow, in the same commit as the test.

For #13, #33 and #35 the branch logic was pulled out of the async fn into a small named fn (the pattern `retry_delay` already used) so it has a test without a canister runtime. Behaviour unchanged; `Err(None)` and `Ok(())` were already equivalent to the queue.

Round three's own work is committed as 142272914. Under the rule, no further review round: the invariant list on #9332 has no NONE rows, so this merges on green CI.

julianjelfs:
CI note: first integration run failed one test of 427, `stable_memory_maps_survive_upgrades`, with "canister is stopped" on a metrics call after its fixed 30-tick wait for three upgrades. Nothing on the daily puzzle path. Rerun of the failed job passed with no changes. That test's fixed wait is a flake candidate under load; worth a poll-until-upgraded on its own small PR, not here.

Issue #9335, opened 2026-09-09
The Daily 3: veto view, aggregates and dark launch


julianjelfs:
## Parent PRD

#9332

## What to build

Everything needed to run the Daily in production without a human on the critical path, then the dark launch itself. Per the PRD's operations stories and Release strategy section.

- Veto view: a dev-only page, guarded by team principals, rendering tomorrow's candidate puzzles with a veto button. First candidate ships if nobody vetoes. Shows the puzzle only, not the solution.
- Per-puzzle aggregates in the daily canister from ingested result rows: solve count, median time, completion rate, hint rate. Exposed on the metrics endpoint.
- Out-of-order upgrade test in PocketIC: canisters upgraded in a non-recommended order, nothing traps, every call answers "not available" until the last piece lands.
- Dark launch on ic_test in the recommended order: user canister, daily canister with `enabled` false, local user index, website. Confirm at each step that nothing is visible and nothing errors. Flip `enabled`, play a puzzle on ic_test, flip it off and confirm the kill switch takes effect at the next push.
- Prod release follows the release train skill, tags per canister, with `enabled` false until the team says go.

This milestone touches deploys and needs a human at every step. Nothing here is automated end to end.

## Acceptance criteria

- [ ] Veto view renders tomorrow's candidates and a veto removes one; an unvetoed pool ships its first candidate at rollover
- [ ] Non-team principals get an error from the veto endpoints
- [ ] Aggregates appear on the daily canister metrics endpoint after solves
- [ ] PocketIC out-of-order upgrade test passes
- [ ] ic_test deployed in order with the feature invisible at each step, then visible only after the flip
- [ ] Kill switch verified on ic_test
- [ ] Prod canisters released with `enabled` false and the website released with the entry hidden; nothing user-visible changed

## Blocked by

- Blocked by #9334

## User stories addressed

- User stories 48, 49, 50, 52, 53

Issue #9334, opened 2026-09-09
The Daily 2: teaching hints and the shareable result card


julianjelfs:
## Parent PRD

#9332

## What to build

The rest of the phase 1 loop on top of the tracer: hints that teach, and the result card that gets shared. At the end of this milestone the full phase 1 experience runs locally: solve, share into a chat, a friend taps play from the card.

Two halves that touch disjoint files and can be built in parallel on the same branch.

Hints, per the PRD's hint flow:

- Local user index hint endpoint. Client sends its filled cells as key and value pairs and the level it wants. Compare filled cells against the solution and return a free mistake hint with those cells as focus if any differ. Otherwise serve the first hint in the chain whose conclusion key is not among the filled cells, at the requested level, debiting the level's price first. Level 1 free. Cap and prices per level from config. Record hints served, and count them in the reward.
- Client hint UI in three levels: highlight the focus cells, show the technique sentence, fill the conclusion. Technique sentence table per technique id per supported language.

Card, per the PRD's result card and sharing sections:

- `daily_result` custom content kind on the existing custom content type. Payload: game id, puzzle number, user id, time, hints, streak, pictogram of the black-cell layout. No backend change beyond the kind string.
- Client renderer for the card. When rendering a chat, batch one query per chat per puzzle to the daily canister's verify endpoint and render the returned time, hints and streak with a verified mark. A card with no matching row renders as unverified with no numbers. Staff accounts marked as staff. Solves under the floor from config are recorded but the client does not offer a card.
- Share button on the solve screen opening the existing chat picker used by the OS share target, setting a draft with the card as attachment and navigating. The claim modal shows today's result with the same share button for a second share or a later one.
- Play button on a card for a viewer who has not solved that puzzle, opening it. For a viewer who has, it shows their own result alongside. Gone once the puzzle is no longer current.

## Acceptance criteria

- [ ] Local user index engine unit tests: mistake hint wins over a strategy hint, hints serve in chain order skipping filled cells, level 1 free, later levels debited, cap enforced, hint count reduces the reward
- [ ] Every technique id emitted by the generator has a sentence in every supported language, checked by a test
- [ ] A forged `daily_result` message with a fake time renders as unverified
- [ ] A genuine card renders with the verified mark and the server's numbers, not the payload's
- [ ] Verification is one batched query per chat per puzzle, not one per card
- [ ] Share from the solve screen and from the claim modal lands the card in the chosen chat's composer
[...]

Issue #9333, opened 2026-09-09
The Daily 1: tracer, play and pay end to end, switched off


julianjelfs:
## Parent PRD

#9332

## What to build

The tracer bullet: a user claims, opens today's Light Up puzzle from the claim modal, pays the entry fee, solves it with live rule painting, the grid submits itself, the local user index verifies and credits CHIT, and the streak counts. No hints, no result card. The feature is switched off by default and every canister change is inert on its own, per the Release strategy section of the PRD.

Developed on one branch. Work in three steps so parts can be built in parallel by different people or agents without colliding:

1. Contracts first, alone. Generator output types, the `daily_puzzle` candid, the new local user index endpoints and args, the user canister credit/debit endpoint, the pushed puzzle bundle, the chit event variant. Api crates and shared types only.
2. Implementation in parallel, each part in its own directory: `light_up` generator crate; `daily_puzzle` canister; user canister endpoint; local user index game engine; client.
3. Integration, alone: PocketIC flow test, run locally end to end.

Checklist of parts, from the PRD's Implementation Decisions:

- `light_up` crate, pure Rust: `generate(seed, params)` returns description, canonical solution indexed by cell, ordered hint chain with focus cells, technique id and conclusion, and tier. Tatham port with the technique solver instrumented for the trace, backtracking uniqueness check, retry until the requested tier. ASCII CLI to print puzzles for eyeballing.
- `daily_puzzle` canister from the `multi_user` skeleton: config with `enabled`, entry fee, hint prices, reward table, rate cap, first-play-free flag; day-ahead timer generating a candidate pool; current puzzle pointer; public description query; push of puzzle, solution, hint chain and config to every local user index at rollover; pull endpoint for a local user index that missed it; results ingest guarded to local user index callers; verify query by user ids and puzzle number.
- User canister: one c2c endpoint callable by its local user index, credit or debit with amount, reason carrying a game id string and puzzle number, idempotency key; balance check and chit event in one message; repeat key refused. One chit event type variant carrying the game id string.
[...]

PR #9346, opened 2026-09-10
The Daily 3/3: boards, hints, demos and result cards

julianjelfs:
Part 3 of 3 for The Daily ([#9332](https://github.com/open-chat-labs/open-chat/issues/9332)). Stacked on #9345. **Read the [design update](https://github.com/open-chat-labs/open-chat/issues/9332#issuecomment-5624095532) on the issue first.**

The half a player sees. This is also the first point at which the feature is runnable end to end, so it is the one to check out and play.

## The thing to review closely

Each game ships a rule checker in TypeScript that mirrors its Rust `check_rules`, so a broken rule paints the moment you make it rather than on submission. **The two must agree**, and where they disagree the server decides, because it holds the solution. A client-side checker that is wrong in the player's favour is a bad experience, not an exploit.

They are deliberately not identical in one respect. The Rust version only ever sees a finished grid; the client sees one in progress. So a rule that is merely unsatisfied must not paint red until it can no longer come good. Tents originally got this wrong and reddened every tree on an empty grid, and Unruly's line counts only fire once a line is over quota, never on a shortfall.

## Shape

The shell knows nothing about cells, bulbs or grids. Every game implements one interface covering parsing, tapping, marking, checking, serialising and resuming, and the modal, the hint painting and the auto-submit are written once against it. Adding a game means a board, a pictogram, a demo and its strings.

## Teaching

The pre-start screen animates the game being played rather than showing an inert copy of today's puzzle: the goal, the easiest opening, a deliberate mistake, the finished grid.

No frame is hand drawn. Each is a set of marks fed through that game's own parser and checker, so a mistake goes red because the checker objects, and a rule change moves the demo with it. A spec holds every demo to that: the last frame must genuinely solve, some frame must genuinely break a rule, and every mark must land on a key the game knows.

This exists because the rules turned out to be unreadable in practice. Loopy's caption used "loop" as though it explained itself, and the rule three of its techniques depend on appeared nowhere in the app.

## Hints

They paint the one thing their sentence names strongly and their context faintly, and clear themselves once you have acted on them rather than sitting there describing a position that no longer exists.

## Cards

Solving offers a card that can be posted to any chat. Readers verify it against the daily canister's results index in a batched query, so a hand-crafted card shows as unverified.

## Also here

CHIT balances are taken from the server's responses rather than adjusted optimistically, which is what used to make the displayed balance drift from the canister.

## Tests

1113 frontend tests, `svelte-check` clean, prettier clean.

## Try it

Enable the config flag on a local replica and use the governance `regenerate_today` op to force each game in turn. All five have been played end to end locally, including sharing and verifying a card.

https://claude.ai/code/session_015hHoyT8UnxPgmUcQSKomBJ


julianjelfs:
Review of this PR runs by invariant, per the rule set 2026-09-13: #9332 now has a Client section, invariants 37 to 48. The review is a table mapping each to the frontend test that would fail if it broke, or NONE; the NONEs are the findings and nothing else in the diff is a blocker. #9333 and #9334 each list which of those numbers they pin, with a slot for the test name. The two the PR body already argues for are 37 (checker parity with `check_rules`) and 47 (demo frames through the real checker); the PR says both have tests, so the mapping should be able to name them.

julianjelfs:
Master merged in (4dd773fe2). Every conflict was this branch's stale copy of the backend; master's side taken and 11 files the backend review had removed dropped, so backend/ is byte-identical to master. Regenerating the agent's types against the merged backend surfaced the contract delta: `max_free_checks` on the puzzle, `free_checks` and `entry_fee` on the user state. Mapped through in c45d8e790, and the entry-fee getter now reads the server's per-user quote instead of deriving it from `first_play_free` and `has_solved_before` (#9332 invariant 40). Nothing displays `free_checks` yet; the review should decide whether that needs an invariant. svelte-check 0 errors, 1117 tests pass, prettier clean. Ready for the invariant review, 37 to 48.

julianjelfs:
## Invariant review, round one

Invariants 37 to 48 from #9332, against the tests on this branch at 51a250b60. For each: the test that would fail if it broke, or NONE. The NONEs are the findings. Nothing else in the diff is raised.

| # | Invariant | Test | Status |
|---|---|---|---|
| 37 | Client checker agrees with `check_rules`: accepts the solution, rejects every single-cell change, every game | Per-game `checkRules` specs cover hand-built cases and "solved grid has no violations". **No test runs the TS checker over generated puzzles from the Rust crates.** The PR body calls parity "the thing to review closely" and it has no cross-language test | **NONE** |
| 38 | Red only once a rule can no longer come good | tents `empty grid reports the short lines but never a tree still waiting`, `one tent between two trees is a grid in progress`, `under-counts are only reported once a line is fully marked`; bridges `under is only reported once every edge of the island is committed`; loopy `loops are not counted while some dot is still open-ended`; unruly `empty grid is incomplete, not wrong`; slant `vertex count: too few possible`. **Light Up**: `checkRules` emits `under` and `unlit` on an empty grid; the board keeps `under` the normal colour and ignores `unlit`, so it holds, but in `Board.svelte` rather than `check`, unlike every other game, and untested | partial |
| 39 | Auto-submit only when every rule passes; server rejection is an error, not a state | `#afterChange` gates on `solvedLocally`; `submit` toasts `wrong`. No test | **NONE** |
| 40 | Prices are the server's quotes | `entryFee` getter reads `userState.entryFee` as of c45d8e790. No test (mine) | **NONE** |
| 41 | Balances are the server's, never optimistic | `mappers.spec` "balances carried on the responses" (4 tests) pins the mapping. `#setChitBalanceFrom` setting the stores is untested | partial |
| 42 | Resume takes the newer grid by timestamp | `#resume` compares `localAt >= serverAt` with a fingerprint guard. No test | **NONE** |
| 43 | Hints below 3 render no conclusions and clear once acted on | Conclusions are withheld server-side (pinned there, #18). `#dropHintOnceDone` clears on the player's edit. No test | **NONE** |
| 44 | Unverified card: no numbers. Verified card: the server's numbers | **Broken.** `DailyResultContent.svelte` renders `content.solveTimeMs`, `content.hintsUsed`, `content.streak` from the message payload in both states; unverified is greyed, not blank, and verified never reads the row it fetched. #9334's acceptance criteria say the opposite twice | **NONE, code wrong** |
| 45 | One batched verification query per chat per puzzle | `#dailyResultsPending` batches per `(game, number)` over 50 ms across every card, which is at most one query per puzzle, fewer than the invariant asks. No test | NONE, low |
| 46 | Nothing visible until an enabled puzzle is fetched | `DailyChitModal` filters `p.enabled`. No test | NONE, low |
| 47 | Every demo frame through the game's own checker | `dailyPuzzleGames.spec`: frames and captions, ends solved, a mistake goes red, marks on known keys, per game | pinned |
| 48 | Play button only while current and only for a viewer who has not solved | `isCurrent && !viewerSolved` in the card. No test | NONE, low |

### Missing invariant

The merge changed the CHIT key format and the client did not follow. Keys are now `{number}:entry`, `{number}:solve` and `{game}:{number}:hint:{step}:{level}`, no fingerprint (the fingerprint segment was reversed in the backend review). `parseDailyPuzzleChitKey` still expects `{game}:{number}:...` and returns `undefined` for every entry and solve key the server now mints, so `ChitEventsForDay` labels a solve reward as "entry". Its spec pins the old formats and "keys minted before the fingerprint", which never shipped. Proposed **#49: the client parses every CHIT key shape the server mints, and its spec is generated from the same shapes the backend test `chit_keys_do_not_identify_the_puzzle` asserts.**

### Findings, ranked

1. **#44** card numbers. Fix the component, then a test: a card with no row renders no numbers; a card with a row renders the row's numbers when they differ from the payload's.
2. **#49** key parser. Fix to the shipped shapes, drop the fingerprint and pre-fingerprint branches, spec the three shapes.
3. **#37** parity. A Rust test in the generator crates that writes N puzzles and solutions per game to a checked-in JSON fixture, and one TS spec that loads it and asserts accept-solution and reject-every-single-cell-change for every game.
4. **#40** fee getter test.
5. **#39, #42, #43** one harness around `DailyPuzzleGame` with a fake client, three tests.
6. **#38** Light Up: move the under and unlit filtering into `check` like the other games, with the same shape of test tents has.
7. **#41, #45, #46, #48** cheap if the harness from 5 exists; otherwise record as accepted NONE.

This is round one. Round two checks these fixes only.

julianjelfs:
## Invariant review, round one: fixes

Every finding from the review above is addressed. One commit per invariant, test first. Pushed to 1837f8805.

| # | Test | Commit | Note |
|---|---|---|---|
| 37 | `dailyGames/parity.spec.ts` (24 generated puzzles, 6 games) fed by the canister test `client_parity_fixture_is_current`, which fails when `parity.json` drifts from the generators | 366a8d13b | all six checkers agree |
| 38 | lightUp.spec `unlit cells are a grid in progress, not a violation`, `a clue is under while it can still be met and impossible once it cannot`, `an unlit cell is impossible once nothing in its sight can take a bulb` | 1837f8805 | **found a gap**: Light Up never painted an impossible clue or an unlightable cell; now it does, and unlit cells in progress are no longer reported |
| 39 | dailyPuzzle.svelte.test `the grid submits itself only once every rule passes`, `a submit the server rejects is an error, not a game state` | 231af459c | |
| 40 | dailyPuzzle.svelte.test `the entry fee is the server's quote for this user, never derived here`, `start sends the quoted fee as the expected price` | 231af459c | |
| 42 | dailyPuzzle.svelte.test `resume takes the newer of the local and server grids by timestamp` | 231af459c | |
| 43 | dailyPuzzle.svelte.test `a hint clears itself once the player has made the move it pointed at` | 231af459c | **found a bug**: a hint cleared on any tap inside its region because "unmarkable" was read as "not yet marked"; it now asks the game whether the pointed keys take a mark |
| 44 | dailyResultCard.spec (3 tests) | af81dc33b | **code was wrong**: both cards now take their numbers from the results row through one pure function that has no access to the payload; no row, no numbers |
| 49 | dailyPuzzle.spec `parses an entry key`, `parses a solve key`, `parses a hint key`, `rejects malformed keys` | f20fc4e50 | **code was wrong**: parser now matches the shipped key shapes |

Accepted as NONE, recorded so the omission is deliberate: **#41** the store write in `#setChitBalanceFrom` (mapping is pinned by 4 mapper tests; the write is a private static on OpenChat), **#45** batching window in `#dailyResultsPending` (private; batches per puzzle, fewer queries than the invariant asks), **#46** the `p.enabled` filter in `DailyChitModal`, **#48** `isCurrent && !viewerSolved` in the card. All four are single expressions in components or private members with no seam; a harness for them would test Svelte templates, which this repo does not do.

Two of the seven findings were real bugs and one was a design gap, all found by writing the invariant's test rather than by reading the code. svelte-check 0 errors, 1155 frontend tests pass (up from 1117), prettier clean on every changed file. Round two checks these fixes only.

julianjelfs:
## Invariant review, round two

Run by a fresh reviewer with no context from round one, checking the round-one fixes only. Method per invariant: read the test, run it, then break the code with a one-line mutation and confirm the test fails, restore. Worktree clean at the end.

| # | Verdict | Mutation that the test caught |
|---|---|---|
| 37 | holds | unruly `solved()` forced true: all four unruly parity seeds fail |
| 38 | holds | `<` to `<=` in `toViolations`: "impossible while a neighbour is free" fails; `canStillBeLit` guard disabled: unlit test fails |
| 39 | holds | `if (this.solvedLocally)` to `if (true)`: submit fires after two taps, test fails |
| 40 | holds | asserted directly, no mutation |
| 42 | holds | asserted directly, no mutation |
| 43 | holds | old `!pointed.some(filled)` restored: test fails, so the fix is real |
| 44 | holds | `cardNumbers` returning zeros when unverified: test fails; both components read only `gameId`, `number`, `tier`, `layout`, `userId`, `caption` from the payload |
| 49 | holds | `parts.length === 5` to `4`: hint key test fails; spec shapes match the backend test `chit_keys_do_not_identify_the_puzzle` |

One note, acted on. The reviewer showed #37 was weaker than its headline: the fixture pinned the generators' output rather than a `check_rules` verdict, and a loosened Light Up clue rule survived the parity test because every one-tap change from a solution also breaks a second rule. Two responses: the Rust fixture test now asserts each solution against the game's `check_rules` before writing it (commit on this branch), and #37 on #9332 is reworded to what the fixture proves, with rule-for-rule agreement on grids in progress left to each checker's hand-built spec cases, where the reviewer confirmed the loosened rule was caught.

No new findings. Under the rule this PR merges on green CI.

julianjelfs:
Scope added, per Julian 2026-09-13: the admin page gains a Daily puzzle tab with the operator actions (series config, per-game config, schedule, push now, regenerate today). Reason: the backend review guards those endpoints behind a platform-operator user lookup, which is right, but operators are browser identities and nothing could call them; the local replica is currently dark for exactly that reason. Invariants 50 to 54 are on #9334. Veto view stays in #9335.

julianjelfs:
## Operator actions: built and reviewed by invariant

Commits 94acbc240, f541ecb14 (cherry-pick of #9356), 76f80a364. Admin page gains a Daily puzzle tab: series config, per-game config, weekday schedule, Push now, Regenerate today.

| # | Test | Status |
|---|---|---|
| 50 | inherits the admin page's `platformOperatorStore` gate; server side `daily_puzzle_canister_serves_todays_puzzle_and_guards_operator_calls` | accepted at tab level |
| 51 | `series form holds the canister's values as strings, nothing write-only`, `the schedule query maps seven days of params`, PocketIC assertion in #9356 | pinned (needed a new `schedule` query, #9356) |
| 52 | `series config round-trips unchanged`, `game config round-trips unchanged`, `schedule round-trips unchanged, all seven days` | pinned |
| 53 | `a refusal comes through as the error, not a value`; re-read after success is component logic | partial, accepted |
| 54 | `as scheduled plus one entry per game the canister reports`, `no game configs leaves only as scheduled` | pinned |

This branch now carries the #9356 commit so the agent types could be regenerated; once #9356 merges to master the same patch is already here and the merge is clean. A website built against a canister without the query shows the default schedule and says so.

svelte-check 0 errors, 1166+ frontend tests, prettier clean on changed files.

julianjelfs:
## Operator tab cut to two levers

Julian, 2026-09-13: none of that config needs to be configurable; if it needs to change it's a code change. The tab is now the enabled flag and Regenerate today, at 0e09384d9 (rebased onto the master merge that brought in #9356). Series config forms, per-game config and the schedule editor are gone, along with their calls through client, worker and agent, and the export attributes on `set_game_config` and `set_schedule`. Saving the flag sends the rest of the config back exactly as read.

The `schedule` query merged as #9356 is now unused; it goes with `set_game_config` and `set_schedule` in the backend follow-up issue, which also moves those values into code.

#9334 block rewritten: invariants 50, 52, 53, 54 with tests; 51 withdrawn with the forms. svelte-check 0 errors, 1160 frontend tests, prettier clean on changed files.

julianjelfs:
Found playing Bridges on the local replica: the free mistake check's red square landed on an unrelated cell. The server names an edge key and the board drew it as a cell index; islands and edges share one number line. Fixed at 8432eb51c through a pure resolver with a test, recorded as #9334 invariant 55. Separately flagged to Julian as a design point: the client's rule painting and the server's mistake mark use the same red and one caption, which reads as three faults when the server named one.

julianjelfs:
Freshness, at 97f49bc4c: the client now polls `daily_puzzle_fetch` on the local user index (one minute active, five idle), schedules a refetch at the current puzzle's expiry, refetches on visibility regain, and an operator's own enable or regenerate refetches at once. A modal whose puzzle is replaced under it rebuilds the board and shows "This puzzle was replaced. Here is the new one." #9334 invariants 56 to 59; 57 and 58 pinned in the domain spec, 56 and 59 are one-line wiring and recorded as accepted. Load: one query per active client per minute against the local user index, the same shape as the bots poll against the user index.

julianjelfs:
Found playing Slant on the replica: the second hint tap (level 2 of the served step) always failed. The client quoted the full level 2 price; the engine prices an upgrade at the difference and refuses the full price. Fixed at ef498a91f: the client quotes the difference, and on a PriceMismatch retries once with the price the server quotes back. #9334 invariant 60, three harness tests.

julianjelfs:
Found playing Slant: a level 2 hint highlighted all four cells round a 3, two of them already drawn, and never cleared because clearing ran only on a tap inside its cells. At 5b9eedd14 the highlight shows only what is still to do (unmarkable context plus empty cells), marked cells drop out, and once the cells the sentence asked for are marked it clears on the next edit. #9334 invariant 61, three harness tests on today's Slant chain.

PR #9359, opened 2026-09-14
The Daily: move config into code, cut the canister to set_enabled

### LINKED ISSUE #9357: Daily puzzle: move config into code, remove the unused canister setters
## Parent PRD

#9332

## What to build

Decided 2026-09-13 while reviewing the operator tab in #9346: prices, rewards, caps, per-game hint prices and the weekday rota do not need to be configurable at run time. If they need to change it is a code change with a review. The daily canister keeps two operator levers, `set_config` (for the enabled flag; the rest of `DailyPuzzleConfig` becomes constants that `set_config` no longer accepts, or is reduced to `set_enabled`) and `regenerate_today`.

Remove from the daily canister: `set_game_config`, `set_schedule`, the `schedule` query (#9356, now unused), and `game_configs` if the regenerate picker can be fed from the schedule constants instead. Move the default schedule and game configs into code as the single source. Update the operator tab in the app if `set_config` becomes `set_enabled`. Keep `enabled` and the puzzle payload travelling to the local user indexes exactly as now.

The PRD's "Config and pricing" section and the design update's "the rota is data, not code" are superseded by this issue.

## Acceptance criteria

- [x] `set_game_config`, `set_schedule` and `schedule` are gone from the candid and the impl, with their integration test coverage removed or redirected
- [x] Prices, rewards, caps and the rota are constants with one definition each; the config validation tests still pin the launch numbers (solve reward below the daily claim, entry below the streak-zero reward, entry plus three full hints above the maximum reward)
- [x] `enabled` still flips from the admin tab and pushes to every local user index
- [x] `regenerate_today` still works with and without a forced game

## Invariants

1. The only runtime-settable series value is `enabled`; every other series and per-game value has exactly one definition, in code.
2. A canister release that changes a price, reward, cap or the rota reaches every local user index on the next push without any operator call.
3. `enabled` false still answers not-available on every game call (#9332 invariant 29 holds unchanged).
4. `regenerate_today` still charges no second entry fee and pays no second reward (#9332 invariant 3 holds unchanged).

## Blocked by

- Blocked by #9346 (the tab's current shape is what this simplifies)

## User stories addressed

- Operations stories in the parent PRD

julianjelfs:
Closes #9357.

Prices, rewards, caps, per-game hint prices and the weekday rota are no longer run-time config. The daily canister keeps two operator levers: `set_enabled` (was `set_config`) and `regenerate_today`. `set_game_config`, `set_schedule`, `schedule` and `game_configs` are gone from the candid, the impl, the integration test client and the app.

## What changed

**Backend**
- `DailyPuzzleConfig` and `GameConfig` keep their shape (they travel with every push to the local user indexes). Their `Default` impls in the types crate are now the one definition of the launch numbers; the daily canister composes the enabled flag onto them in `Data::config()`.
- The rota is `model::schedule::schedule()`, the one definition. The separate test-mode schedule is gone: test mode runs the real rota, so the PocketIC week walk now generates every real slot in the canister, including the 10x10 tricky boards. That subsumed the instruction-count test, which was a `set_schedule` consumer.
- `Data` stores `enabled: bool` instead of the config, game configs and schedule. A serde shim reads the flag out of the `config` map a pre-#9357 build stored, so an upgrade does not switch the game off. The canister has not been to prod yet; the shim is for the ic_test / local replicas and can go after the first prod release.
- `Data::ensure_puzzles()` restamps every held puzzle and candidate with this build's constants before it promotes. Every path that ends in a push runs through it (upgrade, rollover, generation), so there is no separate call an upgrade could forget; `set_enabled` restamps too.
- The setter validation now lives under `#[cfg(test)]` and runs over the constants, plus the three launch relations from the issue.

**Frontend**
- `dailyPuzzleSetEnabled(enabled)` replaces `dailyPuzzleSetConfig(config)`; `dailyPuzzleGameConfigs` is gone. The regenerate picker is fed from the app's own game registry, since a game the app cannot render is not one worth forcing.
- `config` stays as a read-only query for the tab's "currently enabled" line.

## Invariants

| # | Invariant | Test |
|---|-----------|------|
| 1 | The only runtime-settable series value is `enabled`; every other value has one definition in code | `daily_puzzle_canister_impl::tests::launch_numbers_are_constants_that_pass_the_config_checks`, `daily_puzzle_canister_impl::tests::puzzles_carry_the_constant_config_and_the_enabled_flag`, integration `daily_puzzle_canister_serves_todays_puzzle_and_guards_operator_calls` |
| 2 | A release that changes a number reaches every local user index on the next push without an operator call | `daily_puzzle_canister_impl::tests::puzzles_carry_the_constant_config_and_the_enabled_flag` and `legacy_state_keeps_its_enabled_flag_and_drops_the_rest`: both hold a puzzle stamped with an older build's numbers and take it through `ensure_puzzles`, the step `init_state` runs before its push. Both fail if the restamp leaves `ensure_puzzles` |
| 3 | `enabled` false still answers not-available on every game call | integration `daily_puzzle_dark_by_default`; kill-switch section of `daily_puzzle_end_to_end`, which now relies on `set_enabled`'s own push (the `push_now` calls that masked it are gone). Verified: a wasm with the push removed from `set_enabled` fails that test |
| 4 | `regenerate_today` charges no second entry fee and pays no second reward | integration `daily_puzzle_refused_debits_and_replayed_keys` (rewritten onto the constants: the free first play, then a paid second day, then the regeneration replay) |

Acceptance criterion 2's three relations are asserted in `launch_numbers_are_constants_that_pass_the_config_checks`: max solve reward below the top daily claim, entry fee below the streak-zero reward, entry plus three full hints above the max reward.

Dropped with their subject: `set_schedule_does_not_move_todays_game`, `set_schedule_leaves_an_active_regeneration_alone`, `set_game_config_with_unknown_game_id_is_rejected` (#9332 invariant 36, whose setters no longer exist), `game_configs_default_for_every_generator`, and the operator-tab `withEnabled` spec (#9334 invariant 52, now enforced by the API shape: `set_enabled` accepts only the flag).

## Review

Round one (fresh-context, mutation-checked) found two untested invariants: the upgrade restamp was a separate `init_state` call nothing pinned, and every `set_enabled` in the flow tests was followed by `push_now`, so the flag's own push was never exercised. Both fixed as above, test first. Round two checked those fixes only.

## Release

Daily canister only; the local user index is untouched and the pushed puzzle payload is unchanged. Website after the canister, since the tab calls `set_enabled`.


julianjelfs:
**Review by invariants**

Round one (fresh-context, every pinned unit test mutation-checked): two findings, both an invariant without a test that would fail.

1. Invariant 2, "on the next push". The upgrade restamp was a separate `stamp_config()` call in `init_state`; deleting it left all 23 unit tests green. Fixed in d7720d0b9: the restamp now sits at the top of `ensure_puzzles`, which every push path (upgrade, rollover, generation) already runs through. `puzzles_carry_the_constant_config_and_the_enabled_flag` and `legacy_state_keeps_its_enabled_flag_and_drops_the_rest` both hold a puzzle carrying an older build's numbers and take it through `ensure_puzzles`; removing the restamp fails both.
2. Invariant 3. Every `set_enabled` in the flow tests was followed by `push_now`, so the push `set_enabled` owes was never exercised. The kill-switch and re-enable steps of `daily_puzzle_end_to_end` now rely on `set_enabled`'s own push. Verified with a mutant wasm (push removed from `set_enabled`): the test fails at the `puzzles.is_empty()` assert; the clean build passes.

Round two checked those two fixes only: no findings, one stale comment word, fixed in the same commit.

Follow-ups noted, not blocking: `MAX_DAILY_CLAIM` in the launch-numbers test mirrors the user canister by hand; the integration tests carry a second copy of the rota (loud on drift, not silent).

PR #9364, opened 2026-09-14
Daily puzzle: hint button follows the engine's state

### LINKED ISSUE #9360: Daily puzzle: the hint button goes dead while a mistake stands
## Parent PRD

#9332

## What happened

Playtest 2026-09-14 on #20710 (Light Up). With a wrong "no bulb" mark on the board, the hint button read "Hint (level 2, 50 CHIT) · 2 left". Pressing it painted one cell red with "Something is wrong in the red cells." Pressing it again, repeatedly, changed nothing on screen and offered nothing more.

What the code does: the engine checks marked cells against the solution before the priced ladder, and a mistake always wins. While the wrong mark stands every press returns the same lowest wrong key, free of CHIT, never advancing a level; the client overwrites the red set and caption with the same values, so nothing moves. Each such press still spends one of the 20 free checks (refunded only when a paid hint is served); after the twentieth the call fails with `Throttled` and a "failed hint" toast. The button label quotes a level and price derived from the last real hint step, ignoring the standing mistake, so it promises an upgrade the press cannot buy. A mistake hint carries no technique and no conclusions, so there is no higher level to reveal; the deduction trace only resumes once the mark is cleared.

The player's actual confusion was the rules text: "a number says how many bulbs touch that cell" reads as eight neighbours; the rule is four.

## What to build

Client only; the engine's mistake-first behaviour and the one-key-per-call cap stay.

- While the last served hint is a mistake and the board has not changed since, the hint button does not quote a level or a price. Either it is disabled with the caption saying what to do, or it reads as a free re-check. The caption says the next action: clear the red mark, then ask again.
- A press with the same marks as the last mistake hint does not call the server (no free check burned for an answer already on screen). The first press after the board changes calls as now.
- The free-check budget is visible when it matters: the caption or button shows checks left once fewer than five remain, and a `Throttled` refusal shows "no free checks left" rather than the generic failed-hint toast.
- Light Up rules text: number counts bulbs on its four sides, not diagonals. Wording in `dailyPuzzle.games.light_up.rules` and the demo caption if the demo shows a numbered cell.

## Acceptance criteria

- [ ] With a standing mistake and an unchanged board, the button shows no level or price and the caption names the next action
- [ ] A second press on an unchanged board after a mistake makes no server call
- [ ] Changing any mark re-enables a real press, which calls the server as today
- [ ] `Throttled` from the hint call is shown as "no free checks left"; checks left shown once under five
- [ ] Light Up rules text says the number counts bulbs on its sides, not diagonally

## Invariants

1. A hint press with `filled` identical to the `filled` of the last mistake response makes no `daily_puzzle_hint` call.
2. While the last served hint is a mistake and the board is unchanged, the hint button never displays a level or a CHIT price.
3. Any change to the marked cells after a mistake hint re-enables a server-calling press.
4. A `Throttled` error from the hint call is surfaced as the free-check message, not the generic failure toast.
5. The Light Up rules string names the four sides and excludes diagonals.

## Blocked by

None.

## User stories addressed

- Hint stories in the parent PRD (a hint tells the player what to do next; the player is never charged, in CHIT or free checks, for an answer they already have)

julianjelfs:
Closes #9360.

Client only. The engine's mistake-first behaviour and the one-key-per-call cap are unchanged.

## What changed

- The hint button decision moves into `DailyPuzzleGame.hintButton` (`mistake` / `noneLeft` / `hint {level, price, hintsLeft, checksLeft?}`), so the desktop and mobile pages render the same thing and the rules are unit tests.
- While a mistake hint stands and the marks it was answered against are unchanged: the button reads "Fix the red cells first", is disabled, and `hint()` returns without a server call. Any change to the marks clears the mistake state (as before) and re-enables a real press.
- A `Throttled` refusal from the hint call is shown as "No free checks left for today's puzzle"; the count of free checks left is appended to the button once fewer than five remain.
- After a level 3 reveal the caption is "Filled in for you." and the cells it filled are in the faint context highlight with nothing in the "mark this" set. Cleared on the next edit as before.
- Mistake caption now names the next action. Light Up rules: "how many bulbs sit beside it, across or down, never diagonally".

## Invariants (#9360)

| # | Invariant | Test (`app/src/utils/dailyPuzzle.svelte.test.ts`, describe `hint states (#9360)`) |
|---|-----------|------|
| 1 | A press with `filled` identical to the last mistake response makes no hint call | `a second press on the same marks after a mistake makes no server call` |
| 2 | While a mistake stands on an unchanged board the button shows no level or price | `while a mistake stands the button quotes no level and no price` |
| 3 | Any change to the marks after a mistake re-enables a server-calling press | `any change to the marks after a mistake re-enables a real press` |
| 4 | `Throttled` is surfaced as the free-check message, not the generic toast | `a Throttled refusal is shown as no free checks left, not the generic failure` |
| 5 | Light Up rules name the four sides and exclude diagonals | `the Light Up rules name the four sides and rule out diagonals` |
| 6 | After a level 3 reveal the caption is the reveal text and no highlighted cell is one the player is asked to mark | `after a level 3 reveal the caption is the reveal text and nothing is asked of the player` |

Plus `checks left are on the button only once fewer than five remain` for the visible-budget acceptance criterion.

Full frontend suite: 104 files, 1178 tests green. svelte-check 0 errors.


julianjelfs:
**Review by invariants**

Round one (fresh-context, mutation-checked, all six invariants): three findings.
1. Invariant 1 as worded also covers editing back to the same marks and a reload with the wrong mark still on the board; both called the server and spent a free check. Fixed in a585c199e: the mistake is a record of the marks it was answered against, the red cells and caption derive from whether the current marks still match, and the record is saved with the local grid. Both cases are now in the invariant 1 test.
2. Missing invariant: undoing a cell a level-3 reveal filled left "Filled in for you" over a board it no longer described. Fixed: any edit after a reveal retires it. Test `undoing a revealed cell retires the reveal caption and highlight`.
3. Invariant 3's visible half (red cells and caption clearing on the edit) was untested. Pinned in the invariant 3 test.

Round two (those fixes only): one finding, the plain reload with no edit between the mistake and the reload was only reached through intermediate edits, so the write at mistake time was unpinned. Fixed in d265a46e2 with `a reload straight after a mistake resumes the answer without a call`. Round two also confirmed the mutations for each fix fail the right test and that neither page writes to the now-derived `caption`/`mistakes`.

PR #9365, opened 2026-09-14
Daily puzzle: a reset button to clear the board

### LINKED ISSUE #9361: Daily puzzle: a reset button to clear the board
## Parent PRD

#9332

## What happened

Playtest 2026-09-14: after a run of wrong marks the fastest way out is to start the board again, and there is no way to do that. Every mark has to be undone one tap at a time. Reloading does not help either: the grid is saved on the device and on the local user index, and resume takes the newer copy, so the mess comes back.

## What to build

A "Reset" control on the board, client only.

- Clears every player mark and returns the board to the givens. Rule-violation and mistake highlights go with them. The current hint highlight is cleared too; the served hints themselves stay in state and still count.
- Saves the empty grid through the existing save-grid call, so a reload or another device resumes an empty board rather than the mess.
- Costs nothing and changes nothing on the server beyond the saved grid: the start time, entry fee, hints used, free checks and solve record are untouched. It is not a restart and does not re-enter the puzzle.
- Available once the puzzle is started and until it is solved; disabled when the board already holds no marks.
- A full board is a lot to lose on a mis-tap, so the control confirms before clearing (inline "Clear the board?" with a second tap, not a browser dialog).
- Timer keeps running.

## Acceptance criteria

- [ ] Reset empties the board and the saved grid; reloading shows an empty board
- [ ] After reset `started_at`, `hints`, `free_checks` and `solved` on the user state are unchanged and no CHIT moves
- [ ] Reset is disabled before start, after solve, and on an empty board
- [ ] Reset needs a confirming second tap
- [ ] Mistake and rule highlights are gone after reset

## Invariants

1. Reset sends exactly one save-grid call, with the empty grid, and no other daily puzzle call.
2. Reset never changes `startedAt`, `hints`, `freeChecks` or `solved` on the user state.
3. Resume after a reset yields the empty board on both the device copy and the server copy.
4. Reset is unavailable before start, after solve, and when `filled()` is empty.
5. A single tap never clears the board; the clear happens only on the confirming tap.

## Blocked by

None. Sits alongside #9360.

## User stories addressed

- Play stories in the parent PRD (a player can recover from a bad run without abandoning the day's puzzle)

julianjelfs:
Closes #9361. Stacked on #9364 (same class); merge that first.

Client only.

## What changed

- `DailyPuzzleGame.reset()`: first call arms, second call clears every mark and highlight, drops any standing mistake record, and saves the empty grid immediately through the existing save-grid call, so a reload on any device resumes an empty board. Any edit between the two taps disarms it (`resetArmed`).
- `canReset`: started, not solved, not busy, and at least one mark on the board.
- A small secondary "Reset" button beside "Hint" on both pages; it reads "Clear the board?" while armed.
- Nothing else moves: the user state (start time, hints, free checks, solve) is the server's and is not touched; no start, submit or hint call is made.

## Invariants (#9361)

| # | Invariant | Test (`app/src/utils/dailyPuzzle.svelte.test.ts`, describe `reset (#9361)`) |
|---|-----------|------|
| 1 | Reset sends exactly one save-grid call, with the empty grid, and no other daily puzzle call | `a reset sends exactly one save of the empty grid and touches nothing else` |
| 2 | Reset never changes `startedAt`, `hints`, `freeChecks` or `solved` on the user state | same test (user state compared before and after) |
| 3 | Resume after a reset yields the empty board on both the device copy and the server copy | `a resume after a reset yields the empty board` |
| 4 | Reset is unavailable before start, after solve, and when `filled()` is empty | `reset is unavailable before start, after solve, and on an empty board` |
| 5 | A single tap never clears the board; the clear happens only on the confirming tap | `a single tap never clears the board; the confirming tap does`, `an edit between the two taps disarms the reset` |

Plus `a reset clears the mistake, hint highlight and caption with the marks` for the highlight acceptance criterion.

Full frontend suite green (104 files, 1186 tests), svelte-check 0 errors.


julianjelfs:
**Review by invariants**

Round one (fresh-context, 13 mutations, all five invariants killed by at least one): two findings.
1. A level-3 reveal changes the board without disarming the reset, so the confirming tap could wipe a reveal just paid for. Fixed in 857de603b: the disarm moved into the change hook every edit and reveal runs through. Test `a level 3 reveal between the two taps disarms the reset`.
2. The highlight-clearing test served a mistake hint, which carries no highlight, so the `focus`/`target`/caption clears in `reset()` were unpinned. Now serves a level-2 step: `a reset clears the hint highlight and caption with the marks`.

Two follow-ups taken in the same commit: `canReset` false while a call is in flight (assert in the invariant 4 test), and the mistake record dropped on reset (`a reset drops the mistake record: the same wrong mark afterwards is a fresh question`).

Round two on those fixes is in progress.

julianjelfs:
Round two (fixes only): no findings. Each of the four mutations killed exactly its own test; no path changes the board without passing the disarm.

PR #9371, opened 2026-09-14
Daily puzzle: Bridges hints clear once the bridge is drawn

### LINKED ISSUE #9370: Daily puzzle: a Bridges hint stays after the bridge it asked for is drawn
## Parent PRD

#9332

## What happened

Playtest 2026-09-14, Bridges #20710. A hint highlighted the water cell between two islands and the island it pointed at; the player drew that bridge, the puzzle solved, and the highlight stayed on the finished board.

Cause: a hint's highlight clears when the keys it asked for are marked, and the shell asks the game whether a key "takes a mark" and whether it is in `filled()`. That works where hint keys and mark keys are the same space (Light Up, Tents, Slant, Unruly, Loopy). Bridges hint keys are cell indices; its marks are edge keys (`from_cell * 2 + dir`), and the two spaces overlap. `bridges.tap()` treats any key as an edge key, so a cell-index hint key that collides with an existing edge reads as an unmarked, markable cell and the hint never clears; one that collides with nothing reads as context, so the hint clears on the next tap anywhere, acted on or not. The header comment in `bridges.ts` warns exactly this about reading keys by value.

## What to build

Let the game say, in its own key space, whether a hint key is still to do. An optional `hintKeyStatus(model, state, key)` on `DailyGame` returning `"todo"`, `"done"` or `"context"`; the shell uses it for both the initial highlight (done keys are not shown) and the clear-on-edit, falling back to the current markable-and-filled logic for games without it. Bridges implements it: an island cell is context; a water cell is done once any edge running through it has been committed (a bridge, or an explicit "no bridge"), else to do; a cell on no edge is context.

## Acceptance criteria

- [ ] Drawing the bridge a hint pointed at clears the hint; the solved board carries no highlight
- [ ] A tap elsewhere leaves a Bridges hint in place until its bridge is drawn
- [ ] Other games unchanged

## Invariants

1. A Bridges hint that points at a water cell clears once an edge through that cell is committed, and not on an edit elsewhere.
2. A Bridges water cell whose edge is already committed is not highlighted when the hint arrives.

## Blocked by

None. Same file as #9365, so stacked on it.

## User stories addressed

- Hint stories in the parent PRD

julianjelfs:
Closes #9370.

## Cause

Bridges hint keys are cell indices; its marks are edge keys (`from_cell * 2 + dir`), and the spaces overlap. The shell judged "acted on" by asking whether the hint's key takes a mark and has one, which `bridges.tap()` answered as if the cell index were an edge key. A colliding edge made the hint stick after the bridge was drawn (the playtest case); no collision made it clear on the next tap anywhere.

## What changed

- `DailyGame.hintKeyStatus?(model, state, key)` returning `"todo" | "done" | "context"`. The shell uses it for the initial highlight (done keys are not shown) and the clear-on-edit, falling back to the markable-and-filled reading for games without it.
- Bridges implements it in cell space: an island is context; a water cell is done once any edge through it is committed (bridge or explicit no-bridge); a cell no edge crosses is context.
- Other games unchanged (no `hintKeyStatus`, same fallback as before).

## Invariants (#9370)

| # | Invariant | Test |
|---|-----------|------|
| 1 | A Bridges hint that points at a water cell clears once an edge through that cell is committed, and not on an edit elsewhere | `dailyPuzzle.svelte.test.ts`, describe `a Bridges hint clears in its own key space (#9370)`: `clears once an edge through the cell is drawn, and not on an edit elsewhere` (fixture where the hinted cell 1 collides with edge key 1) |
| 2 | A Bridges water cell whose edge is already committed is not highlighted when the hint arrives | same describe: `a water cell whose edge is already committed is not highlighted` |

Plus `bridges.spec.ts` › `hintKeyStatus` › `reads a hint's cell key in cell space, never as an edge key`.

Full frontend suite green, svelte-check 0 errors.


julianjelfs:
**Review by invariants**

Round one (fresh-context, mutation-checked): no blocking findings. Removing the game's answer in `#keyStatus` reproduces the original bug and fails the invariant 1 test (and, in a variant without the colliding tap, the stuck path itself); ignoring explicit zeros fails the Bridges unit test; the old highlight filter fails the invariant 2 test. Fallback confirmed equivalent to the old logic for games without the hook; Loopy and Tents need none.

One follow-up taken in 8000a3114: the fixtures targeted a water cell, but the server targets the island with its water cells in focus. `with an island target, the hint clears once the water cells in focus are decided` takes that path and would also catch an island reading as to-do (previously only the unit test did). No round two, since there was no fix to check.

PR #9373, opened 2026-09-14
Bridges: crossing edges get their own tap targets; no-bridge mark hides under a bridge

### LINKED ISSUE #9372: Bridges: an edge whose only water cell is crossed by another edge cannot be placed
## Parent PRD

#9332

## What happened

Playtest 2026-09-14, Bridges #20710. The solver confirms the puzzle has exactly one solution, reachable by the easy trace, but it cannot be entered in the UI. The bottom-middle 2 needs a single bridge up to the 4 above it, through one water cell. That cell is also on the horizontal edge between the two 2s beside it. The board draws one transparent hit rectangle per edge, in edge order, so in a cell two edges share the later rectangle wins every tap. Here that is the horizontal edge: every tap on that cell drew the horizontal bridge, and the vertical one has no other cell to tap. Reproduced in the browser.

Any edge whose water cells are all shared with crossing edges is unreachable; any shared cell hides one of its two edges.

## What to build

Per-cell tap targets that give each of two crossing edges its own area. In a cell shared by a horizontal and a vertical edge, split it along its diagonals: the left and right triangles tap the horizontal edge, the top and bottom triangles tap the vertical one. Unshared cells stay whole. The shapes are computed in the game module as data (`bridgesHitShapes(desc)`, one entry per cell per edge), so they are unit-testable; the Board renders them.

## Acceptance criteria

- [ ] Today's #20710 can be solved in the browser
- [ ] Tapping the top or bottom of a shared cell cycles the vertical edge; left or right cycles the horizontal one
- [ ] Unshared cells behave as now

## Invariants

1. Every edge of a description has at least one tap shape, and in a cell shared by two edges each edge has a shape of its own.
2. Tap shapes of different edges never overlap.

## Blocked by

None. Same file as #9371, so stacked on it.

## User stories addressed

- Play stories in the parent PRD (every legal move can be made)

### LINKED ISSUE #9374: Bridges: hide a no-bridge mark under a crossing bridge, and draw it as a cross
## Parent PRD

#9332

## What happened

Playtest 2026-09-14, Bridges #20710, with #9373 applied. A water cell shared by two edges showed a double horizontal bridge and, in the middle of it, the grey dot that marks the vertical edge as "no bridge". Both marks are true at once (a bridge on one edge rules out the other), but drawn together the dot reads as a mark on the bridge. The dot itself is also easy to mistake for a bullet or a hint mark rather than a decision the player made.

## What to build

- An explicit no-bridge mark on an edge is not drawn while any edge crossing it carries a bridge. The decision is still held in state (it is what `filled()` sends and what the count check uses); only the glyph goes. It reappears if the crossing bridge is removed.
- The glyph is a small cross rather than a dot.

Decided in the game module's `marks()` so it is unit-testable: an explicit zero whose crossing edge has a bridge produces no mark entry.

## Acceptance criteria

- [ ] On #20710, a bridge over the shared cell hides the crossing edge's no-bridge mark; removing the bridge shows it again
- [ ] The no-bridge mark is a cross

## Invariants

1. `marks()` yields no entry for an edge with an explicit zero while an edge crossing it has a bridge, and yields `none` for it again once the crossing bridge is gone; the state and `filled()` are unchanged either way.

## Blocked by

Stacked on #9373 (same files).

## User stories addressed

- Play stories in the parent PRD

julianjelfs:
Closes #9372. Closes #9374.

Two Bridges fixes from playtesting #20710, on one branch so they can be tested together.

## 1. Crossing edges each get a tap target (#9372)

**Cause.** One transparent hit rect per edge, drawn in edge order. In a water cell two edges share, the later rect took every tap. #20710 needs a single bridge whose only water cell is shared with a crossing edge, so it could not be placed: the solver finds exactly one solution, the UI could not enter it. Reproduced in the browser.

**Change.** `bridgesHitShapes(desc)`: one entry per water cell per edge. A cell one edge crosses is `whole`; a shared cell is split along its diagonals, `left`/`right` for the horizontal edge and `top`/`bottom` for the vertical one. `hitQuarter` in `gridSvg.ts` gives the polygon for each part; the Board renders those instead of per-edge rects.

| # | Invariant | Test |
|---|-----------|------|
| 1 | Every edge has at least one tap shape, and in a shared cell each edge has a shape of its own | `bridges.spec.ts` › `bridgesHitShapes` › `every edge has a shape, and in a shared cell each edge has its own` |
| 2 | Tap shapes of different edges never overlap | `bridges.spec.ts` › `shapes of different edges never overlap`; `gridSvg.spec.ts` › `hitQuarter` › `the four quarters tile the cell exactly` |

## 2. No-bridge mark under a crossing bridge (#9374)

`bridges.marks()` yields no mark for an explicit zero while any crossing edge has a bridge; state and `filled()` are unchanged, and the mark returns when the bridge is removed. The glyph is a small cross instead of a dot.

| # | Invariant | Test |
|---|-----------|------|
| 1 | `marks()` yields no entry for an explicit zero while a crossing edge has a bridge, yields `none` again once it is gone, and state and `filled()` are unchanged either way | `bridges.spec.ts` › `no-bridge marks under a crossing bridge` › `marks() drops the mark while the crossing edge has a bridge, and keeps the state` |

## Review

Both parts had a fresh-context, mutation-checked round one with no blocking findings (records in the comments here and on the closed #9375). Full frontend suite green, svelte-check 0 errors.


julianjelfs:
**Review by invariants**

Round one (fresh-context, mutation-checked): no blocking findings. Reverting to one whole shape per edge (the original bug), giving both edges the same quarters, dropping shapes for crossing edges, and adding a whole shape alongside quarters each fail the right test. Geometry, layering (every overlay is `pointer-events: none`) and a partially-crossed two-cell edge all checked.

Follow-ups taken in 020aa4766: the quarter geometry moved into `hitQuarter` in `gridSvg.ts` with a spec that the four quarters tile the cell and every interior point falls in exactly one, so invariant 2 is pinned on the shapes as drawn, not only the part names; the non-overlap test now demands four distinct quarters on a shared cell; the partially-crossed layout joins its fixtures. No round two, since there was no fix to check.

PR #9375, opened 2026-09-14
Bridges: no-bridge mark hides under a crossing bridge, drawn as a cross

### LINKED ISSUE #9374: Bridges: hide a no-bridge mark under a crossing bridge, and draw it as a cross

julianjelfs:
Closes #9374. Stacked on #9373 (same files); merge that first.

## What changed

- `bridges.marks()`: an explicit zero on an edge produces no mark while any edge crossing it has a bridge. The state and `filled()` are unchanged, so the count check and the server still see the decision; the mark returns when the crossing bridge is removed.
- The no-bridge glyph is a small cross instead of a dot.

## Invariants (#9374)

| # | Invariant | Test |
|---|-----------|------|
| 1 | `marks()` yields no entry for an explicit zero while a crossing edge has a bridge, yields `none` again once it is gone, and state and `filled()` are unchanged either way | `openchat-shared/src/utils/dailyGames/bridges.spec.ts` › `no-bridge marks under a crossing bridge` › `marks() drops the mark while the crossing edge has a bridge, and keeps the state` |

Full frontend suite green (107 files, 1240 tests), svelte-check 0 errors.


julianjelfs:
**Review by invariants.** Round one (fresh-context, mutation-checked): no blocking findings. Emitting the mark under a bridge (old behaviour), deleting the state along with the mark, and never re-emitting after the bridge is removed each fail the named test on the right assertion. Nothing else reads `marks()` for Bridges except the Board and the demo; the hint highlight and the count check read state. No round two, since there was no fix to check.

julianjelfs:
Folded into #9373 so both fixes test together on one branch.

PR #9378, opened 2026-09-14
Chat selection: position on the window target as it lands

### LINKED ISSUE #9358: Chat selection focused message jitter
On the native android app (and perhaps mobile web), when I select a chat with many unread messages, the intended behaviour is that we should load a message window around the first unread message and the viewport should arrive focused with that first unread message in the center. 

The actual behaviour is that the window loads with the wrong messages visible and then leaps to the correct position afterwards. It is not clear whether we actually load the most recent messages and then load the correct message window or whether the loaded messages are correct and it is _only_ a scrolling issue. 

The goal is to load the correct set of messages and have no visible jumping of the timeline - the chat should simply appear with the first unread message in the center of the viewport. 

## Diagnosis

The loaded window is correct. The jump is a render-before-position gap: the window's events land in the store and render at scrollTop 0 (newest rows visible), and only after the load's remaining awaits (user lookup, stream completion, `loadedMessageWindow`) does the list scroll to the target. Measured locally: newest rows visible for ~70ms before the jump, then a 188px refine. On Android the gap is longer.

## Invariants

1. A message window load announces its target before any of its events are applied: `loadEventWindow` publishes `loadingMessageWindow` with the context and message index for chat and thread windows alike.
2. When the announced target is present in the list's items, the list positions on it in the same flush that renders those items, without waiting for `loadedMessageWindow`.
3. While an announced target is outstanding and not yet positioned, the viewport is hidden.
4. A window load that ends without positioning its target (failure, fallback to latest messages, target filtered out) reveals the viewport.
5. A target announced for a different context is dropped when the list's context changes; one announced for the incoming context survives the switch and is honoured.
6. An items change landing while a navigation is in flight re-centres the navigation's target; it does not pin a row to where it sat before the navigation's refine.

## Tests

| Invariant | Test |
|---|---|
| 1 | `openchat-client/src/loadEventWindow.spec.ts` "publishes loadingMessageWindow before the window's events are applied" / "...for a thread window before its events are applied" |
| 2 | `app/src/components_shared/ChatEventListFocus.spec.ts` "positions on the target in the flush that renders it, before loadedMessageWindow" |
| 3 | `ChatEventListFocus.spec.ts` "hides the viewport while the announced target is outstanding" |
| 4 | `ChatEventListFocus.spec.ts` "reveals the viewport when the load ends without its target" / "...falls back to previous messages" / "...target never appears in the items"; `loadEventWindow.spec.ts` "a failed window load publishes loadedMessageWindow without a message index" |
| 5 | `ChatEventListFocus.spec.ts` "a target announced for the incoming chat survives the context switch" / "a target announced for another chat is dropped when the context changes" |
| 6 | `ChatEventListFocus.spec.ts` "an items change during the navigation re-centres the target instead of undoing it" |

julianjelfs:
Closes #9358.

Selecting a chat with unread messages loaded the right window but rendered it at the bottom first. The events land in the store and paint at scrollTop 0, and only after the load's remaining awaits (user lookups, later chunks, `loadedMessageWindow`) does the list scroll to the first unread. Measured locally: ~70ms of the newest messages before the jump, then a 188px refine. On Android the gap is longer.

## Change

- `loadEventWindow` publishes `loadingMessageWindow` before any events are applied (chat and thread windows). A load that ends without a window publishes `loadedMessageWindow` with no message index.
- The shared list positions on the announced target in the same flush that renders it, and hides the viewport while the target is outstanding, so a partial cache chunk without it cannot paint either.
- `VirtualChatList`'s items-pass re-anchor defers to a navigation in flight. The `chat_start` row appended in the same flush re-pinned the viewport over the navigation's refine, landing ~330px off centre until the debounced correction 700ms later.

## Invariants

Listed on #9358 with the test that pins each one. Tests 1, 4 (client) and 6 were checked red against the unfixed code; 2, 3, 4, 5 were written before the implementation and failed on it.

## Verification

Local dev app, `vcl_debug` on, a DOM sampler recording every list mutation while selecting a group chat with 35 unread messages:

| | before | after |
|---|---|---|
| first non-empty state | newest rows at scrollTop 0, target absent (~70ms) | viewport hidden |
| positioning | jump to target 186px off, refine 30ms later | target rendered at reveal, refined in the same microtask flush |

The sampler ran in a background tab, so it records DOM states, not paints.


julianjelfs:
Verified on a local Android device: chat opens with the first unread centred, no jump.

